PR #3, opened 2022-05-08
Add curp dir

rogercloud:


kv/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
pwang7, line 1: Please add as many Clippy lints as possible.

pwang7 (changes requested):

kv/src/main.rs
pzheng1025: I will add in next PR.

pzheng1025: @rogercloud Please add for curp

pzheng1025 (approved):

PR #7, opened 2022-05-22
Support distributed Put/Range/Delete

pzheng1025:


xline/src/server/xline_server.rs
```diff
@@ -86,62 +137,6 @@ impl XlineServer {
...
-    }
-    /// Consume `ExecutionRequest` and get ownership of each field
-    pub(crate) fn unpack(self) -> (Command, Sender<ResponseOp>) {
-        let Self { command, notifier } = self;
-        (command, notifier)
-    }
-}
-
 #[tonic::async_trait]
 impl Kv for XlineRpcServer {
```
rogercloud, line 141: Can we avoid repeated code via generic?

xline/src/storage/kvstore.rs
```diff
@@ -38,16 +39,46 @@ struct KvStoreInner {
...
 impl KvStore {
     /// New `KvStore`
     pub(crate) fn new() -> Self {
-        let (request_tx, mut request_rx) = channel(CHANNEL_SIZE);
+        let (request_tx, mut request_rx) = mpsc::channel(CHANNEL_SIZE);
         let inner = Arc::new(KvStoreInner::new());
         let inner_clone = Arc::clone(&inner);
         let _handle = tokio::spawn(async move {
             let mut btree = inner.btree.lock().await;
             while let Some(req) = request_rx.recv().await {
```
rogercloud, line 79: Can we put the btree into the while loop so that we can protect btree with sync lock, which is much faster?

rogercloud (commented):
Will you replace most panics with proper Error in the future?

xline/src/storage/kvstore.rs
pzheng1025: So far this is temp solution, because this is the only position to modify btree, so I get lock first and then pass guard. Will change to get lock inside the loop.

xline/src/server/xline_server.rs
pzheng1025: I will move these common codes to a function.

pzheng1025: > Will you replace most panics with proper Error in the future?

Yes. we should only panic on unrecoverable error.

PR #8, opened 2022-05-28
Support revision and refactor storage to index + db

pzheng1025:


xline/src/storage/kvstore.rs
```diff
@@ -35,8 +38,74 @@ pub(crate) struct KvStore {
...
 #[derive(Debug)]
 struct KvStoreInner {
-    /// B-tree map
-    btree: Mutex<BTreeMap<Vec<u8>, KeyValue>>,
+    /// Key Index
+    index: Mutex<BTreeMap<Vec<u8>, Vec<KeyRevision>>>,
+    /// DB to store key value
+    db: Mutex<HashMap<Revision, KeyValue>>,
+    /// Revision
+    revision: Mutex<i64>,
```
rogercloud, line 46: Is atomic faster for `i64` type?

rogercloud, line 46: Is atomic faster for `i64` type?

xline/src/storage/kvstore.rs
```diff
@@ -35,8 +38,74 @@ pub(crate) struct KvStore {
 /// KV store inner
 #[derive(Debug)]
 struct KvStoreInner {
-    /// B-tree map
-    btree: Mutex<BTreeMap<Vec<u8>, KeyValue>>,
+    /// Key Index
+    index: Mutex<BTreeMap<Vec<u8>, Vec<KeyRevision>>>,
+    /// DB to store key value
+    db: Mutex<HashMap<Revision, KeyValue>>,
```
rogercloud, line 44: Maybe BtreeMap is better? We'll remove the early revisions for GC.

xline/src/storage/kvstore.rs
```diff
@@ -35,8 +38,74 @@ pub(crate) struct KvStore {
...
+#[derive(Debug, Copy, Clone)]
+struct KeyRevision {
+    /// Last creation revision
+    create_revision: i64,
+    /// Number of modification since last creation
+    version: i64,
+    /// Last modification revision
+    mod_revision: i64,
+    /// Sub revision in one transaction
+    sub_revision: i64,
```
rogercloud, line 59: Do different sub revisions mean different key revision? If yes, a transaction may take many revisions. Is it what we're expecting?

xline/src/storage/kvstore.rs
```diff
@@ -160,55 +225,111 @@ impl KvStoreInner {
...
+
     #[allow(clippy::pattern_type_mismatch)]
     /// Handle `RangeRequest`
-    fn handle_range_request(
-        btree: &MutexGuard<BTreeMap<Vec<u8>, KeyValue>>,
-        req: RangeRequest,
-    ) -> RangeResponse {
+    fn handle_range_request(&self, req: RangeRequest) -> RangeResponse {
+        let key_index = self.index.lock();
+        let db = self.db.lock();
```
rogercloud, line 237: If they're always locked at the same time, shall we move them into one lock?

rogercloud: For several locks acquired at the same time, we'd better document the order somewhere to make sure we keep the order everywhere.

xline/src/storage/kvstore.rs
```diff
@@ -217,51 +338,83 @@ impl KvStoreInner {
...
             ONE_KEY => {
-                if let Some(kv) = btree.remove(key) {
-                    prev_kvs.push(kv);
-                    response.deleted = 1;
+                if let Some(indexes) = key_index.get_mut(key) {
+                    if let Some(index) = Self::get_index(indexes) {
+                        let prev_kv = db
+                            .get(&index.as_revision())
+                            .cloned()
+                            .unwrap_or_else(|| panic!("Failed to get key {:?} from DB", key));
```
rogercloud, line 358: Please also log the index or revision.

xline/src/storage/kvstore.rs
pzheng1025: Operations in on txn will have the same global revision but different sub_revision. It doesn't mean different global revision.

xline/src/storage/kvstore.rs
```diff
@@ -160,55 +225,111 @@ impl KvStoreInner {
...
+                let mut values: Vec<KeyValue> = key_index
+                    .values()
+                    .filter_map(|indexes| {
+                        Self::get_index(indexes)
+                            .and_then(|index| db.get(&index.as_revision()).cloned())
+                    })
+                    .collect();
                 kvs.append(&mut values);
             }
             _ => {
```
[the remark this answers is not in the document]
pzheng1025: This match is to match vector. It is impossible to list all others.

xline/src/storage/kvstore.rs
pzheng1025: So far there is only one thread. use mutex is cheap when no contention.

xline/src/storage/kvstore.rs
pzheng1025: I will separate index and db in to different module in next patch. Lock will be acquired in their own module.

xline/src/storage/kvstore.rs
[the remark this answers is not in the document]
pzheng1025: Done

rogercloud (approved):

PR #9, opened 2022-05-31
Curp code skeleton and update rustc version

rogercloud:
- Curp first commit to handle propose
- Update rustc to 1.61 otherwise madsim won't compile
- Fix 1.61 clippies


curp/src/server.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Default for Server<C, CE> {
+    #[inline]
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[allow(missing_docs)]
```
pzheng1025, line 27: already has docs?

curp/src/message.rs
```diff
@@ -0,0 +1,51 @@
...
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::cmd::Command;
+
+/// The propose send by a client to servers
+#[derive(Serialize, Deserialize, Request)]
+#[serde(bound = "C: Serialize + DeserializeOwned")]
+#[rtype("ProposeResponse<C>")]
+pub(crate) struct Propose<C: 'static + Command> {
```
pzheng1025, line 13: Why 'static is needed?

curp/src/server.rs
rogercloud: It's macro expand code, we can't modify it in this crate.

pzheng1025 (approved):

PR #12, opened 2022-06-07
Client judge

rogercloud:


curp/src/client.rs
```diff
@@ -1 +1,139 @@
...
+}
+
+impl<C> Client<C>
+where
+    C: Command + 'static,
+{
+    /// Create a new protocol client based on the addresses
+    #[inline]
+    pub async fn new(leader: &SocketAddr, addrs: &[SocketAddr]) -> Option<Self> {
+        let ep = Endpoint::bind("127.0.0.1:0")
```
pzheng1025, line 48: Will this endpoint use random port number?

curp/src/client.rs
```diff
@@ -1 +1,139 @@
...
+
+        let mut ok_cnt: usize = 0;
+        let mut max_term = 0;
+        let mut execute_result: Option<C::ER> = None;
+
+        while ok_cnt
+            >= (ft
+                .wrapping_add(ft.wrapping_add(1).wrapping_div(2))
+                .wrapping_add(1))
+            && execute_result.is_some()
```
pzheng1025, line 85: Is this logic reversed?

pzheng1025: What is the reason of using wrapping_add instead of overflowing_add?

curp/src/server.rs
```diff
@@ -1,39 +1,129 @@
...
+    /// handle Propose
+    #[rpc]
+    async fn propose(&self, p: Propose<C>) -> ProposeResponse<C> {
+        self.inner.propose(p).await
+    }
+
+    /// Run a new rpc server
+    #[allow(dead_code)]
+    async fn run(is_leader: bool, term: u64, others: &[SocketAddr]) -> Result<(), ServerError> {
+        let rx_ep = Endpoint::bind(format!("0.0.0.0:{DEFAULT_SERVER_PORT}")).await?;
```
pzheng1025, line 36: servers use same port then we cannot test on same machine.

curp/src/cmd.rs
```diff
@@ -19,6 +19,9 @@ pub trait Command:
     /// Get keys of the command
     fn keys(&self) -> &[Self::K];
 
+    /// Get cmd id
+    fn id(&self) -> &CmdId;
```
pzheng1025, line 23: Is this something like propose id?

curp/src/client.rs
rogercloud: Yes, port 0 means pick a random port number, it's the client end.

curp/src/client.rs
rogercloud: Yes, good catch. we can make sure ft the result is less than usize limit for that `ft = peer_cnt / 2`.

curp/src/cmd.rs
rogercloud: It's propose id actually, let me rename it.

curp/src/server.rs
rogercloud: Let me add a customized port option.

pzheng1025 (approved):

PR #17, opened 2022-06-18
fix propose_indexed bug

rogercloud:
Skip keyconflict error on the first rpc as we always wait the second
wait-sync response.


pzheng1025:
Issue has been fixed. Need to fix clippy.

pzheng1025 (approved):

PR #19, opened 2022-06-19
Add lock and lease proto file.

pzheng1025:


pzheng1025:
Comment out unnecessary line in auth.proto rpc.proto kv.proto instead of  removal.

PR #24, opened 2022-06-27
Execute order

rogercloud:


curp/src/keybased_channel.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+impl<K: Eq + Hash + Clone, M> KeybasedChannelReceiver<K, M> {
+    /// Receive a message
+    /// Return (message, `msg_complete_sender`)
+    #[allow(dead_code)]
+    pub(crate) fn recv(&self) -> (M, UnboundedSender<K>) {
+        loop {
+            let lisenter = {
```
pzheng1025, line 118: possible to move the common logic to a helper function?

pzheng1025:
Is this a MPSC channel?

curp/src/keybased_channel.rs
rogercloud: I've taken it into consideration, but there's logic to return the value in the middle of a loop, line 123. A function can't work that way.

rogercloud:
> Is this a MPSC channel?

It's a MPMC channel, so it can be used as a MPSC channel.

pzheng1025 (approved):

PR #25, opened 2022-06-28
change keybased channel to multi-keys version

rogercloud:


curp/src/keybased_channel.rs
```diff
@@ -12,55 +13,90 @@ use tokio::task::JoinHandle;
...
+    /// Create a new `KeysMessageInner` structure
+    fn new(keys: Vec<K>, message: M) -> Self {
+        Self { keys, message }
+    }
+}
+
+/// Keys and Messages combined structure
+pub(crate) struct KeysMessage<K, M> {
+    /// Inner
+    inner: Arc<KeysMessageInner<K, M>>,
```
pzheng1025, line 34: Is it possible to use Rc instead of Arc?

rogercloud: It's transfered via async channels, Arc is required.

curp/src/keybased_channel.rs
```diff
@@ -118,31 +220,26 @@ enum MessageOrListener<M> {
...
-                Some(msg) => return MessageOrListener::Message(msg),
-                #[allow(clippy::match_ref_pats)] // don't move listener out
-                None => match &local_listener {
-                    &Some(_) => {
-                        break;
-                    }
-                    &None => local_listener = Some(channel.new_msg_event.listen()),
-                },
-            }
+        self.buf.extend(channel.pop_all_msg().into_iter());
```
pzheng1025, line 232: possible to use mem::swap()?

rogercloud: it's just one move fromone place to another, mem::swap is not more efficient in this case I think.

pzheng1025 (approved):

PR #32, opened 2022-07-29
Tests

themanforfree:


xline/src/main.rs
```diff
@@ -1,6 +1,6 @@
 //! Xline
 #![deny(
-    // The following are allowed by default lints according to
+    // The following are allowed by default &lints according to
```
pzheng1025, line 3: remove &

xline/src/main.rs
```diff
@@ -110,19 +110,12 @@
     clippy::multiple_crate_versions, // caused by the dependency, can't be fixed
 )]
 
-use std::net::{AddrParseError, SocketAddr};
-
 use anyhow::Result;
 use clap::Parser;
 use log::debug;
+use std::net::SocketAddr;
```
pzheng1025, line 116: std need to be put before third party libs

themanforfree: cargo fmt will sort them alphabetically,  this modification will make format CI fail

pzheng1025: You need to add a empty line after std, please check other files

PR #33, opened 2022-07-30
Support watch

pzheng1025:


xline/src/server/lease_server.rs
```diff
@@ -74,12 +74,12 @@ impl Lease for LeaseServer {
         debug!("Receive LeaseKeepAliveRequest {:?}", request);
         // TODO: Temporary solution
         let mut req_stream = request.into_inner();
-        let (tx, rx) = mpsc::channel(100);
+        let (tx, rx) = mpsc::channel(128);
```
rogercloud, line 77: how do we determine the channel buffer size?

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+            request_rx,
+            event_rx,
+            event_tx,
+            watcher_map: HashMap::new(),
+            next_id: 0,
+        }
+    }
+
+    /// Get next available `WatchId`, return -1 if id is not available
+    fn get_next_watch_id(&mut self, id: WatchId) -> WatchId {
```
rogercloud, line 68: Why not return Option<WatchId>？

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,312 @@
...
+}
+
+/// KV watcher inner data
+#[derive(Debug)]
+pub(crate) struct KvWatcherInner {
+    /// KV storage
+    storage: Arc<KvStoreInner>,
+    /// watchers map
+    /// TODO: change to more effecient data structure
+    watcher_map: Mutex<HashMap<KeyRange, HashSet<Watcher>>>,
```
rogercloud, line 152: Why do we use async version Mutex here?

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,312 @@
...
+        revision
+    }
+
+    /// Handle KV store updates
+    async fn handle_kv_updates(&self, updates: (i64, Vec<Event>)) {
+        let (revision, updates) = updates;
+        let watcher_map = self.watcher_map.lock().await;
+        for (k, v) in watcher_map.iter() {
+            let events: Vec<Event> = updates
+                .clone()
```
rogercloud, line 257: don't clone the entire vec, we can clone the content after the filter.

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,312 @@
...
+
+            if !events.is_empty() {
+                for watcher in v.iter() {
+                    if revision < watcher.start_rev() {
+                        continue;
+                    }
+                    let filtered_event = watcher.filter_events(events.clone());
+                    if !filtered_event.is_empty() {
+                        watcher
+                            .notify(watcher.watch_id(), revision, filtered_event)
```
rogercloud, line 277: Do we have to send the notification while holding the watcher_map lock?

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,312 @@
...
+                    )
+                })
+                .collect();
+
+            if !events.is_empty() {
+                for watcher in v.iter() {
+                    if revision < watcher.start_rev() {
+                        continue;
+                    }
+                    let filtered_event = watcher.filter_events(events.clone());
```
rogercloud, line 274: The same, we don't need to clone the events vec here. The filtered_event can be cloned inside the function `filter_events`

xline/src/storage/kvstore.rs
```diff
@@ -221,10 +246,35 @@ impl KvStoreInner {
...
+            .map(|kv| {
+                // Delete
+                #[allow(clippy::as_conversions)] // This cast is always valid
+                let event_type = if kv.version == 0 && kv.create_revision == 0 {
+                    EventType::Delete as i32
+                } else {
+                    EventType::Put as i32
+                };
+                Event {
+                    r#type: event_type,
```
rogercloud, line 268: Why can't we use type `EventType` for the field `r#type`?

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+        // 0 means auto-generate
+        if id == 0 {
+            loop {
+                let next = self.next_id;
+                self.next_id = self.next_id.overflow_add(1);
```
rogercloud, line 73: will we reuse the watch id? what if wrapping happens?

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+                        .kv_watcher
+                        .watch(
+                            watch_id,
+                            key_range,
+                            req.start_revision,
+                            req.filters,
+                            self.event_tx.clone(),
+                        )
+                        .await;
+                    let _prev = self.watcher_map.insert(watch_id, watcher);
```
rogercloud, line 120: the prev should be none, we'd better make sure it's none, or comment the reason to ignore the return value.

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+                            revision,
+                            ..ResponseHeader::default()
+                        }),
+                        watch_id,
+                        created: true,
+                        ..WatchResponse::default()
+                    };
+                    assert!(
+                        self.response_tx.send(Ok(response)).await.is_ok(),
+                        "Receiver closed"
```
rogercloud, line 133: Please specify the channel name that closes.

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,312 @@
...
+            .or_insert_with(|| {
+                let mut set = HashSet::new();
+                let _ = set.insert(watcher.clone());
+                set
+            });
+        (watcher, events, revision)
+    }
+
+    /// Cancel a watch from KV store
+    async fn cancel(&self, watcher: Watcher) -> i64 {
```
rogercloud, line 236: We can pass in the watcher reference instead of value

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+                    } else {
+                        assert!(
+                            self.response_tx
+                                .send(Err(tonic::Status::not_found(format!(
+                                    "Watch ID {} doesn't exist",
+                                    req.watch_id
+                                ))))
+                                .await
+                                .is_ok(),
+                            "Receiver closed"
```
rogercloud, line 178: The same as above, specify the closed channel name.

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+                                revision,
+                                ..ResponseHeader::default()
+                            }),
+                            watch_id,
+                            canceled: true,
+                            ..WatchResponse::default()
+                        };
+                        assert!(
+                            self.response_tx.send(Ok(response)).await.is_ok(),
+                            "Receiver closed"
```
rogercloud, line 167: The same as above, specify the closed channel name.

xline/src/server/watch_server.rs
```diff
@@ -0,0 +1,263 @@
...
+                        } else {
+                            // TODO handle client close connection
+                            panic!("Client close connection");
+                        }
+                    }
+                    event = watch_handle.event_rx.recv() => {
+                        if let Some(event) = event {
+                            watch_handle.handle_watch_event(event).await;
+                        } else {
+                            panic!("Watch event sender is closed");
```
rogercloud, line 253: Can we just jump out the loop and leave a log instead of panicing?

xline/src/storage/kvstore.rs
```diff
@@ -37,11 +35,13 @@ pub(crate) struct KvStore {
...
     /// Sender to send sync request
     sync_tx: mpsc::Sender<SyncRequest>,
+    /// KV watcher
+    kv_watcher: Arc<KvWatcher>,
 }
 
 /// KV store inner
 #[derive(Debug)]
-struct KvStoreInner {
+pub(crate) struct KvStoreInner {
```
rogercloud, line 44: Why is the inner structure `pub(crate)`?

pzheng1025: This is referred by Kv_watcher. Will change the name.

xline/src/storage/kvwatcher.rs
pzheng1025: We need to hold this lock across await.

xline/src/server/watch_server.rs
pzheng1025: each client will have a dedicated watch id pool and watch id is i64. Don't need to reuse.

xline/src/server/lease_server.rs
pzheng1025: This is the experience from etcd. We may need to tune this in the future. 
Let me define a const first.

xline/src/server/watch_server.rs
pzheng1025: Done

xline/src/storage/kvstore.rs
pzheng1025: In `Event`, `r#type` filed is i32, we need to cast.

xline/src/storage/kvwatcher.rs
pzheng1025: Done.

xline/src/storage/kvwatcher.rs
pzheng1025: Done. Just pass reference.

xline/src/server/watch_server.rs
pzheng1025: Done

xline/src/storage/kvwatcher.rs
pzheng1025: Done

xline/src/server/watch_server.rs
pzheng1025: So far it is a non-recoverable error as sender side is closed we cannot receive watch event any more. Will consider robustness in the future.

xline/src/storage/kvstore.rs
pzheng1025: Done

xline/src/storage/kvwatcher.rs
pzheng1025: We need to generate all data to send during holding the lock, this will cause a lot of copy. And this notify is just sendimg message through async channel.

xline/src/storage/kvstore.rs
pzheng1025: Fix this cast issue.

xline/src/storage/kvwatcher.rs
rogercloud: Can we use arc to avoid payload copy?

pzheng1025: Done

xline/src/storage/kvwatcher.rs
pzheng1025: Change to sync lock.

xline/src/storage/kvwatcher.rs
```diff
@@ -0,0 +1,326 @@
...
+            .or_insert_with(|| {
+                let mut set = HashSet::new();
+                let _ = set.insert(watcher.clone());
+                set
+            });
+        (watcher, events, revision)
+    }
+
+    /// Cancel a watch from KV store
+    async fn cancel(&self, watcher: &Watcher) -> i64 {
```
rogercloud, line 250: please make functions to sync ones.

pzheng1025: interesting. Clippy doesn't catch this.

rogercloud (approved):

PR #35, opened 2022-08-05
Update parse of ServerArgs, and change crate type from bin to lib.

themanforfree:


pzheng1025 (approved):

xline/src/main.rs
```diff
@@ -132,47 +126,33 @@ struct ServerArgs {
     #[clap(long)]
     name: String,
     /// Cluster peers. eg: node1=192.168.x.x:8080;node2=192.168.x.x:8080
-    #[clap(long)]
-    cluster_peers: String,
+    #[clap(long, multiple = true, required = true)]
```
pzheng1025, line 129: Please update help message at line 128

PR #36, opened 2022-08-05
Support range with sort, limit and count-only

themanforfree:


xline/src/storage/kvstore.rs
```diff
@@ -237,7 +237,46 @@ impl KvStoreInner {
             count: kvs.len().cast(),
             ..RangeResponse::default()
         };
-        response.kvs = kvs;
+        if !req.count_only {
```
pzheng1025, line 240: Please check how etcd handle this.

xline/src/rpc/mod.rs
```diff
@@ -71,6 +71,7 @@ pub(crate) use self::etcdserverpb::{
     compare::{CompareResult, CompareTarget, TargetUnion},
     kv_server::{Kv, KvServer},
     lease_server::{Lease, LeaseServer},
+    range_request::{SortOrder, SortTarget},
```
pzheng1025, line 74: only include range_request. And use range_request::SortOrder and range_request::SortTarget.

xline/src/storage/kvstore.rs
themanforfree: [etcd](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/txn/txn.go#L166-L205)  
here is the code for etcd to handle this. I think it's more complicated than the current implementation

xline/src/storage/kvstore.rs
```diff
@@ -237,7 +237,46 @@ impl KvStoreInner {
...
+            match (req.sort_target(), req.sort_order()) {
```
pzheng1025, line 241: use req.sort_target/req.sort_order directly?

xline/src/storage/kvstore.rs
```diff
@@ -237,7 +237,46 @@ impl KvStoreInner {
...
+                    kvs.sort_by(|a, b| b.mod_revision.cmp(&a.mod_revision));
+                }
+                (SortTarget::Value, SortOrder::Ascend | SortOrder::None) => {
+                    kvs.sort_by(|a, b| a.value.cmp(&b.value));
+                }
+                (SortTarget::Value, SortOrder::Descend) => {
+                    kvs.sort_by(|a, b| b.value.cmp(&a.value));
+                }
+            }
+            response.kvs = if req.limit > 0 && kvs.len() > req.limit.cast() {
```
pzheng1025, line 274: req.limit > 0 && kvs.len() > req.limit.cast()   => (req.limit > 0) && (kvs.len() > req.limit.cast()) to make it clearer.

xline/src/storage/kvstore.rs
```diff
@@ -237,7 +237,46 @@ impl KvStoreInner {
...
+                kvs.into_iter().take(req.limit.cast()).collect()
```
pzheng1025, line 275: kvs.resize() is better? Don't need to clone.

xline/src/storage/kvstore.rs
themanforfree: Use them directly will match the i32 type, which works but is hard for humans to read.

xline/src/storage/kvstore.rs
themanforfree: When limit is less than len, resize will call truncate, so it is better to call truncate directly

pzheng1025 (approved):

PR #38, opened 2022-08-15
Support fast path.

pzheng1025:


xline/src/server/kv_server.rs
```diff
@@ -128,23 +143,31 @@ impl Kv for KvServer {
...
-                debug!("Get revision {:?} for RangeRequest", revision);
-                Self::update_header_revision(&mut res, revision);
-                if let Response::ResponseRange(response) = res {
-                    Ok(tonic::Response::new(response))
-                } else {
-                    panic!("Receive wrong response {:?} for RangeRequest", res);
-                }
-            }
-            Err(e) => panic!("Failed to receive response from KV storage, {e}"),
+        let is_fast_path = true;
```
rogercloud, line 146: Why do we hard code is_fast_path value here?

pzheng1025: This is test purpose. In the future, this value should read from protobuf.

rogercloud (approved):

PR #40, opened 2022-08-20
fix mpsc bug that blocks following messages

rogercloud:


curp/src/channel/key_mpsc.rs
```diff
@@ -120,7 +120,7 @@ impl<K: Eq + Hash + Clone + ConflictCheck, M> KeybasedChannel<MpscKeysMessage<K,
 
         while let Some(cur) = ready_pool.pop_front() {
             self.inner.push_back(cur.clone());
```
pzheng1025, line 122: move this to end so clone is not needed?

pzheng1025 (approved):

PR #51, opened 2022-09-26
benchmark in CI

themanforfree:


rogercloud (commented):
Please add multi-key test cases

.github/workflows/ci.yml
```diff
@@ -5,8 +5,7 @@ on:
     branches: [master]
 
 env:
-  # TODO: Update to 1.61.0, 1.58-1.60 has regression
```
rogercloud, line 8: Why do we change Rust version from 1.61 to 1.63?

rogercloud (commented):
We should easily get the benchmark result from the benchmark test. Where can we find the result?

rogercloud (approved):

PR #53, opened 2022-10-04
Some refators

rogercloud:
1. Refactor the msg broadcast logic
2. Make kv tests parallel runnable

### DIFF
--- a/curp/src/sync_manager.rs
+++ b/curp/src/sync_manager.rs
@@ -1,13 +1,14 @@
 use std::{iter, sync::Arc, time::Duration};
 
 use clippy_utilities::{NumericCast, OverflowArithmetic};
-use futures::{stream::FuturesUnordered, StreamExt};
+use futures::{stream::FuturesUnordered, Future, StreamExt};
 use parking_lot::Mutex;
 use tracing::{error, warn};
 
 use crate::{
     channel::{key_mpsc::MpscKeybasedReceiver, key_spmc::SpmcKeybasedSender, RecvError},
     cmd::Command,
+    error::ProposeError,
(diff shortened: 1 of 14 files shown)


xline/src/server/xline_server.rs
```diff
@@ -28,8 +30,6 @@ use crate::storage::KvStore;
 pub struct XlineServer {
     /// Server name
     name: String,
-    /// Address of server
-    addr: SocketAddr,
```
pzheng1025, line 32: Ok to remove for now. May need to store this information in the future.

pzheng1025 (approved):

curp/src/sync_manager.rs
```diff
@@ -104,97 +108,173 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                                term,
+                                index.numeric_cast(),
+                                cmds_cloned.as_ref(),
+                            )?)
+                            .await
+                    },
+                    |r| match r.into_inner().commit_response {
+                        Some(CommitResponse::Committed(_)) => 1,
+                        Some(CommitResponse::WrongTerm(_) | CommitResponse::PrevNotReady(_)) => 0,
+                        None => unreachable!("Should contain sync response"),
```
pzheng1025, line 271: Should contain commit response

xline/src/server/xline_server.rs
rogercloud: Yes. Let's add it back later if needed.

curp/src/sync_manager.rs
rogercloud: unreachable is a panic. The whole process is terminated.

PR #57, opened 2022-10-18
Client

themanforfree:


xline/src/client/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    ) -> Self {
+        let etcd_client = EtcdClient::connect(
+            all_members
+                .iter()
+                .map(ToString::to_string)
+                .collect::<Vec<_>>(),
+            None,
+        )
+        .await
+        .unwrap_or_else(|e| panic!("failed to connect to server: {}", e));
```
rogercloud, line 58: Client should not panic if any connection fails, even returning an error is a better solution.

rogercloud: We'd better create one kind of client, either etcd or curp, and providing an argument to let user choose a client.

xline/src/client/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    /// set `use_curp_client`
+    #[inline]
+    pub fn set_use_curp_client(&mut self, use_curp_client: bool) {
+        self.use_curp_client = use_curp_client;
+    }
+
+    /// Generate a new `ProposeId`
+    fn generate_propose_id(&self) -> ProposeId {
+        ProposeId::new(format!("{}-{}", self.name, Uuid::new_v4()))
```
rogercloud, line 76: If all name is "client", why should we put it into the propose Id? Is there other kind of propose id  that doesn't come from a client?

xline/src/client/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// parse `ResponseOp` into `Response`
+    fn parse_response_op(response_op: ResponseOp) -> Response {
+        if let Some(response) = response_op.response {
+            response
+        } else {
+            panic!("Receive empty ResponseOp");
```
rogercloud, line 84: Please don't panic, if any unexpected happens. It's a client used by the others, we'd better not crash users thread.

xline/src/client/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            let req_op = RequestOp {
+                request: Some(Request::RequestPut(request.into())),
+            };
+            let propose_id = self.generate_propose_id();
+            let cmd = Command::new(key_ranges, req_op.encode_to_vec(), propose_id);
+            let cmd_res = self.curp_client.propose(cmd).await?;
+            if let Response::ResponsePut(response) = Self::parse_response_op(cmd_res.decode()) {
+                Ok(response)
+            } else {
+                Err(anyhow!("Receive wrong response type"))
```
rogercloud, line 109: We should avoid anyhow "Result" and "error" in a serious lib.

xline/src/client/mod.rs
themanforfree: we need both of them, because some requests like put can use curp client get a faster response, but requests like watch can not use curp client , they communicate directly with the server use etcd client

xline/src/client/mod.rs
themanforfree: The curp client on the server side will use the name of the server

rogercloud (approved):

PR #58, opened 2022-10-25
new benchmark tool

themanforfree:


benchmark/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+    env_logger::builder()
+        .filter_module("benchmark", log::LevelFilter::Info)
+        .init();
+
+    let args: Benchmark = Benchmark::parse();
+
+    let mut runner = CommandRunner::new(args);
+    let stats = runner.run().await?;
+
+    println!("{}", stats.summary());
```
rogercloud, line 16: Please use log as you've init it at the beginning of the function.

benchmark/src/runner.rs
```diff
@@ -0,0 +1,321 @@
...
+            "  Average:      {:.4} secs\n",
+            self.avg.as_secs_f64()
+        ));
+        s.push_str(&format!("  Requests/sec: {:.2}\n", self.qps));
+        s
+    }
+
+    /// Calculate the histogram from the latencies.
+    #[allow(clippy::indexing_slicing)]
+    pub fn histogram(&self) -> String {
```
rogercloud, line 93: Is there any histogram lib existed? If not, can we provide one?

benchmark/src/runner.rs
```diff
@@ -0,0 +1,321 @@
...
+            if let Some(err) = result.error {
+                let entry = self.errors.entry(err).or_insert(0);
+                *entry = entry.overflow_add(1);
+                continue;
+            }
+            bar.inc(1);
+            stats.latencies.push(result.elapsed);
+        }
+        stats.total = start.elapsed();
+        stats.qps = stats.latencies.len() as f64 / stats.total.as_secs_f64();
```
rogercloud, line 280: please use numeric_cast instead of as

scripts/benchmark.sh
```diff
@@ -0,0 +1,244 @@
...
+OUTPUT_DIR="${WORKDIR}/out"
+REPEAT=1
+SERVERS=("172.20.0.2" "172.20.0.3" "172.20.0.4" "172.20.0.5")
+CLUSTER_PEERS=(
+    ""
+    "${SERVERS[2]}:2379 ${SERVERS[3]}:2379"
+    "${SERVERS[1]}:2379 ${SERVERS[3]}:2379"
+    "${SERVERS[1]}:2379 ${SERVERS[2]}:2379"
+)
+XLINE_TEST_CASE=(
```
rogercloud, line 12: Can we make the shell simpler as there're many duplications.

benchmark/src/runner.rs
themanforfree: numeric_cast cannot cast usize to f64

scripts/benchmark.sh
themanforfree: Yes, but cannot quickly run the specified test_case

benchmark/src/main.rs
themanforfree: We need to output these information to stdout, log is only used to observe the running status of the program

benchmark/src/runner.rs
themanforfree: I found a lib but can't output a graph like this

rogercloud (approved):

PR #60, opened 2022-10-29
Add CI job to build and push docker image

themanforfree:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,48 @@
+name: Docker
+
+on:
+  pull_request:
+    branches: 
+      - master
+    types: [closed]
```
rogercloud, line 7: If we push docker when a PR gets closed, how do we benchmark a PR as it's not in the dockerhub. Does it mean we can only benchmark a merged commit?

benchmark/benchmark.sh
```diff
@@ -103,49 +98,102 @@ stop_all() {
...
 }
 
 set_cluster_latency() {
-    set_latency node0 ${SERVERS[1]} ${SERVERS[2]} ${CLIENT} 50ms 50ms 75ms &
-    set_latency node1 ${SERVERS[0]} ${SERVERS[2]} ${CLIENT} 50ms 50ms 75ms &
-    set_latency node2 ${SERVERS[0]} ${SERVERS[1]} ${CLIENT} 50ms 50ms 25ms &
-    set_latency client ${SERVERS[0]} ${SERVERS[1]} ${SERVERS[2]} 75ms 75ms 25ms &
+    cluster_size=$1
+    docker exec client tc qdisc add dev eth0 root handle 1: prio bands $((cluster_size + 4))
+    for ((j = 1; j < ${cluster_size}; j++)); do
```
rogercloud, line 108: in the first level we use "i" instead of "j" as a convension

benchmark/benchmark.sh
```diff
@@ -103,49 +98,102 @@ stop_all() {
...
+        if [[ ${i} -eq ${cluster_size} ]]; then
+            set_latency node${i} 172.20.0.2 50ms ${idx} &
+        else
+            set_latency node${i} 172.20.0.2 75ms ${idx} &
+        fi
+    done
     wait
 }
 
+run_container() {
```
rogercloud, line 130: I recommend to add some comment before the complex functions to tell the meaning of each arguments, as it's hard to tell in a shell script.

.github/workflows/docker.yml
themanforfree: maybe we should compile xline in every benchmark, otherwise it will generate a lot of useless tags on docker hub

scripts/benchmark.sh
```diff
@@ -0,0 +1,220 @@
...
+ETCD_TESTCASE=(
+    "node1  false ${SERVERS[1]}:2379"
+    "node2  false ${SERVERS[2]}:2379"
+    "client false ${SERVERS[3]}:2379"
+)
+KEY_SPACE_SIZE=("1" "100000")
+CLIENTS_TOTAL=("1 50" "10 300" "50 1000" "100 3000" "200 5000")
+FORMAT="%-8s\t%-12s\t%-8s\t%-s\n"
+
+# generate benchmark command by arguments
```
rogercloud, line 27: The name is usually obvious, but the arguments are not. Please add comment meaning for the function.

themanforfree: done

rogercloud: We usually put the argument comments above the method definition, as you can't always make sure to read the argument at the beginning of the function definition. Here's the example:

```rust
# function doc
# args:
#    1. xxxx
#    2. xxxx
```

rogercloud (approved):

PR #66, opened 2022-11-07
Async commit

markcty:
Use Raft algorithm as curp's background sync protocol

Changes:
- log[0] now stores a fake log with 0 cmds to simplify code logics
- server state is moved into a struct and protected by a RWLock.
- 3 background tasks:
  - background_ping: send heartbeat to followers
  - backgournd_commit: check if new logs are replicated on all servers, commit if so
  - background_apply: check if commit_index is larger than applied_index, execute the log if so


curp/src/sync_manager.rs
```diff
@@ -144,137 +137,222 @@ impl<C: Command + 'static> SyncManager<C> {
...
+    pub(crate) async fn run<CE: 'static + CommandExecutor<C>>(
+        &mut self,
+        comp_chan: SpmcKeyBasedSender<C::K, SyncCompleteMessage<C>>,
+        cmd_executor: Arc<CE>,
+        commit_notify: Arc<Notify>,
+    ) {
+        // notify when a broadcast of append_entries is needed immediately(a new log is received or committed), skipping heartbeat timeout
+        let ae_notify = Arc::new(Notify::new());
+        // TODO: gracefully stop the background tasks
+        let _background_ping_handle = tokio::spawn(Self::bg_append_entries(
```
rogercloud, line 150: This task should only run for the leader?

curp/src/sync_manager.rs
```diff
@@ -144,137 +137,222 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                        // If the majority of servers has replicated the log, commit
+                        // + 1 means including the leader itself
+                        if match_count + 1 >= majority_cnt {
+                            debug!("commit_index updated to {i}");
+                            state.commit_index = i;
+                            commit_notify.notify_one();
+                            ae_notify.notify_one();
+                            break;
+                        }
+                    }
```
rogercloud, line 300: The algorithm is not efficient. We don't have to call iter().filter() for possible index. As we're holding the state write lock, the can calculate the index by scanning the match_index once.

curp/src/sync_manager.rs
```diff
@@ -144,137 +137,222 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                } else {
+                    *state.next_index.get_mut(&connect.addr).unwrap() -= 1;
```
rogercloud, line 302: If the request fails, why do we sub next_index by 1? Do we need to keep it unchanged?

curp/src/sync_manager.rs
```diff
@@ -144,137 +137,222 @@ impl<C: Command + 'static> SyncManager<C> {
...
+        state: Arc<RwLock<State<C>>>,
+        commit_notify: Arc<Notify>,
+        ae_notify: Arc<Notify>,
+    ) {
+        #[allow(clippy::integer_arithmetic)] // tokio internal triggered
+        loop {
+            // wait for either heartbeat timeout
+            tokio::select! {
+                _ = tokio::time::sleep(Self::APPEND_ENTRIES_INTERVAL) => {},
+                _ = ae_notify.notified() => {},
```
rogercloud, line 197: We should not send heartbeat once an entry getting committed. We should reset the heartbeat timeout once we get a notification.

curp/src/sync_manager.rs
```diff
@@ -144,137 +137,222 @@ impl<C: Command + 'static> SyncManager<C> {
...
+        comp_chan: SpmcKeyBasedSender<C::K, SyncCompleteMessage<C>>,
+        commit_notify: Arc<Notify>,
+    ) {
+        loop {
+            commit_notify.notified().await;
+            if state.read().need_commit() {
+                let mut state = state.write();
+                #[allow(clippy::integer_arithmetic, clippy::indexing_slicing)]
+                // TODO: will last_applied overflow?
+                for i in (state.last_applied + 1)..=state.commit_index {
```
rogercloud, line 321: Yes, we should do GC to remove the useless entries or store them on the disk. So the last_applied won't overflow.

curp/src/server.rs
```diff
@@ -1,20 +1,25 @@
...
     sync::Arc,
+    vec,
 };
 
 use clippy_utilities::NumericCast;
 use event_listener::{Event, EventListener};
 use futures::FutureExt;
 use opentelemetry::global;
 use parking_lot::{Mutex, MutexGuard, RwLock};
+use tokio::sync::Notify;
```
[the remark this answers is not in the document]
markcty: done

curp/src/sync_manager.rs
markcty: A TODO is added.

curp/src/server.rs
```diff
@@ -637,108 +719,54 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
-        if let Ok(CommitResponse {
-            commit_response: Some(commit_response::CommitResponse::Committed(true)),
-        }) = commit_result
-        {
-            let cmds = cr.cmds::<C>().map_err(|e: bincode::Error| {
-                tonic::Status::internal(format!("encode or decode error, {}", e))
-            })?;
+        // remove inconsistencies
+        while state.last_log_index() > req.prev_log_index.numeric_cast() {
+            let _drop = state.log.pop();
```
rogercloud, line 737: Vec truncate is more efficient

curp/src/sync_manager.rs
```diff
@@ -144,137 +139,233 @@ impl<C: Command + 'static> SyncManager<C> {
...
-                            .await
-                    },
-                    |r| match r.into_inner().commit_response {
-                        Some(CommitResponse::Committed(_)) => 1,
-                        Some(CommitResponse::WrongTerm(_) | CommitResponse::PrevNotReady(_)) => 0,
-                        None => unreachable!("Should contain sync response"),
-                    },
-                    || true,
-                    max_fail,
+    /// Send heartbeat to a server
```
rogercloud, line 232: It's not only a heart beat, it may also be a real append entry request.

curp/src/sync_manager.rs
```diff
@@ -144,137 +139,233 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                            .filter(|addr| state.match_index[*addr] >= last_sent_index)
+                            .count()
+                            >= majority_cnt
+                    {
+                        debug!("commit_index updated to {last_sent_index}");
+                        state.commit_index = last_sent_index;
+                        if let Err(e) = commit_trigger.send(()) {
+                            error!("ae_trigger closed: {}", e);
+                        }
+                        if let Err(e) = ae_trigger.send(()) {
```
rogercloud, line 305: When we commit a entry, why do we also trigger a "append_entry" request?

curp/src/sync_manager.rs
```diff
@@ -144,137 +139,233 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                    state.term,
+                    0, // TODO: add leader_id
+                    next_index - 1,
+                    state.log[next_index - 1].term(),
+                    state.log[next_index..].to_vec(),
+                    state.commit_index,
                 )
-                .await
             });
+        let last_sent_index = prev_log_index + entries.len();
```
rogercloud, line 254: There may be several `send_append_entries` called concurrently, the smaller `last_sent_index` doesn't mean the request is sent earlier as there's no lock protection. Please double check the correctness.

curp/src/sync_manager.rs
```diff
@@ -144,137 +139,233 @@ impl<C: Command + 'static> SyncManager<C> {
...
+            Err(e) => warn!("Heartbeat error: {}", e),
+            Ok(resp) => {
+                let resp = resp.into_inner();
+                let mut state = state.write();
+                if resp.success {
+                    // update match_index and next_index
+                    *state.match_index.get_mut(&connect.addr).unwrap() = last_sent_index;
+                    *state.next_index.get_mut(&connect.addr).unwrap() = last_sent_index + 1;
+
+                    let majority_cnt = (state.others.len() + 1) / 2 + 1;
```
rogercloud, line 288: If the total server count is 3, the others length is 2, so the majority_cnt is 2. But what we're checking is the response count, ignoring the leader. `majority_cnt` should be `(other.len() + 1) / 2`.

curp/src/sync_manager.rs
```diff
@@ -144,137 +139,233 @@ impl<C: Command + 'static> SyncManager<C> {
...
+    async fn bg_apply<CE: 'static + CommandExecutor<C>>(
+        state: Arc<RwLock<State<C>>>,
+        cmd_executor: Arc<CE>,
+        comp_chan: SpmcKeyBasedSender<C::K, SyncCompleteMessage<C>>,
+        mut commit_notify: UnboundedReceiver<()>,
+        spec: Arc<Mutex<RawMutex, VecDeque<C>>>,
+    ) {
+        while commit_notify.recv().await.is_some() {
+            if state.read().need_commit() {
+                let mut state = state.write();
```
rogercloud, line 327: Don't have to acquire this lock twice. Try [this](https://docs.rs/lock_api/latest/lock_api/struct.RwLock.html#method.upgradable_read).

curp/proto/message.proto
```diff
@@ -35,42 +35,22 @@ message WaitSyncedResponse {
     }
 }
 
-message SyncRequest {
+message AppendEntriesRequest {
     uint64 term = 1;
-    uint64 index = 2;
-    repeated bytes cmds = 3;
+    uint64 leader_id = 2;
```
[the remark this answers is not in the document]
markcty: It is mentioned in the Raft paper and is used to redirect clients. But I don't think we need it. I will remove it.

curp/src/sync_manager.rs
markcty: It is an optimization to reduce commit latency. For example, if the leader sends heartbeats every 100ms, a log is received and replicated on the majority of the servers on the 20th ms. We want to trigger an "append_entries" request at once so that the commit index is carried to the followers immediately instead of waiting for another 80ms.

curp/src/sync_manager.rs
markcty: On line 290 I explained my way of calculating the majority.

curp/src/sync_manager.rs
markcty: The variable actually means the index of the last sent log entry **in the current request**. I will add a comment above this line of code.

rogercloud: It's used in the resp handling. If the response with "last_sent_index 1" returns firstly, then "last_sent_index 0" response arrives. Will the next index and match index be set correctly?

curp/src/sync_manager.rs
rogercloud: It's a kind of "optimization" to make commit happens earlier on followers, but I'm not sure what's the benefit of the shorter commit latency.

curp/src/sync_manager.rs
rogercloud: The plus operation is not necessary. You can delete both of them to get the same result.

curp/src/sync_manager.rs
markcty: I see. I will add a check before changing match_index and next_index

curp/src/sync_manager.rs
markcty: Removed.

curp/tests/concurrent_cmd.rs
```diff
@@ -0,0 +1,28 @@
...
+                ProposeId::new(format!("id{i}")),
+                TestCommandType::Get,
+                vec![format!("id{i}")],
+                None,
+            ))
+            .await;
+    }
+
+    // watch the log while doing sync
+    tokio::time::sleep(Duration::from_secs(1)).await;
```
rogercloud, line 27: We'd better to have something to verify instead of observe the log in the test.

rogercloud (approved):

curp/tests/concurrent_cmd.rs
markcty: Tracking it in issue.

rogercloud (approved):

PR #67, opened 2022-11-08
Some refector for command execute

themanforfree:


xline/src/client/errors.rs
```diff
@@ -13,6 +13,9 @@ pub enum ClientError {
     /// Propose error
     #[error("propose error {0}")]
     ProposeError(#[from] curp::error::ProposeError),
+    /// Bincode error
+    #[error("bincode error {0}")]
+    BincodeError(#[from] bincode::Error),
```
rogercloud, line 18: May duplicate with ParseError, please double check it.

themanforfree: I think it is not duplicate, BincodeError is an error when encode the command, but ParseError is an error when the response is not the expected response type

rogercloud: We should not expose implementation detail such as Bincode to the client.

rogercloud (approved):

PR #68, opened 2022-11-08
fix the bug that the Event uses wrong prev_kv

themanforfree:


xline/src/storage/kvstore.rs
```diff
@@ -550,12 +551,12 @@ impl KvStoreBackend {
             version: new_rev.version,
             ..KeyValue::default()
         };
-        let prev = self.db.insert(new_rev.as_revision(), kv.clone());
+        let _prev = self.db.insert(new_rev.as_revision(), kv.clone());
```
rogercloud, line 554: What if another put request takes place between line 541 and line 554? Is the previous value correct in this case?

themanforfree: this key is in spec poll when sync, another put request of the same key is conflict, it cannot execute and sync, so this case is impossible

rogercloud (approved):

PR #76, opened 2022-11-14
[Curp] Add leader election

markcty:


markcty:
I still leave the `is_leader` field when creating a server. It makes it easier in testings. However, if two leaders are assigned by the user, thing may go wrong

curp/src/server.rs
```diff
@@ -764,6 +843,44 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
+        request: tonic::Request<VoteRequest>,
+    ) -> Result<tonic::Response<VoteResponse>, tonic::Status> {
+        let req = request.into_inner();
+        debug!(
+            "vote received: term({}), last_log_index({}), last_log_term({}), id({})",
+            req.term, req.last_log_index, req.last_log_term, req.candidate_id
+        );
+
+        let state = self.state.upgradable_read();
+        if req.term < state.term || state.voted_for.is_some() {
```
rogercloud, line 860: Even if you voted_for someone, you should also accept a vote with a higher term.

curp/src/sync_manager.rs
```diff
@@ -415,4 +455,120 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                    }
+                    ServerRole::Candidate => Self::CANDIDATE_TIMEOUT,
+                    ServerRole::Leader => continue, // leader should not vote
+                }
+            };
+            let next_check = last_rpc_time.read().to_owned() + timeout;
+            tokio::time::sleep_until(next_check).await;
+            if Instant::now() - *last_rpc_time.read() < timeout {
+                continue;
+            }
```
rogercloud, line 492: It still requires periodically waking up to check, I guess a tokio::select helps here. Because we don't need to start a vote if heartbeat or append_entries arrives on time. We can ignore a state read lock here if you take my way.

curp/src/sync_manager.rs
```diff
@@ -415,4 +455,120 @@ impl<C: Command + 'static> SyncManager<C> {
...
+
+            // start election
+            #[allow(clippy::integer_arithmetic)] // TODO: handle possible overflow
+            let req = {
+                let mut state = state.write();
+                let new_term = state.term + 1;
+                state.update_to_term(new_term);
+                state.set_role(ServerRole::Candidate);
```
rogercloud, line 500: you should vote_for yourself at least.

rogercloud (commented):
If a replica becomes a leader, it should try to replicate all uncommitted logs to others. Current implementation only takes care newcomers.

curp/src/server.rs
```diff
@@ -563,7 +601,6 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
             .map_lock(|mut spec| {
                 let is_conflict = Self::is_spec_conflict(&spec, &cmd);
                 spec.push_back(cmd.clone());
-                debug!("insert cmd {:?} to spec pool", cmd.id());
```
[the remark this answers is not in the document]
markcty: Sry, by accident. I will restore it.

curp/src/sync_manager.rs
markcty: Fixed

curp/src/sync_manager.rs
```diff
@@ -415,4 +444,225 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                }
+
+                #[allow(clippy::integer_arithmetic)]
+                if resp.vote_granted {
+                    debug!("vote is granted by server {}", connect.addr);
+                    let mut state = RwLockUpgradableReadGuard::upgrade(state);
+                    state.votes_received += 1;
+
+                    // the majority has granted the vote
+                    let min_granted = (state.others.len() + 1) / 2 + 1;
```
rogercloud, line 567: if there are 3 servers, others = 2, min_grant = 2, the server can  becomes a leader only if votes_received = 3. Is it correct?

curp/src/sync_manager.rs
```diff
@@ -415,4 +444,225 @@ impl<C: Command + 'static> SyncManager<C> {
...
+                        #[allow(clippy::shadow_unrelated)] // clippy false positive
+                        let (
+                            term,
+                            prev_log_index,
+                            prev_log_term,
+                            entries,
+                            leader_commit,
+                            last_sent_index,
+                        ) = state.map_read(|state| {
+                            let next_index = state.next_index[&connect.addr];
```
rogercloud, line 609: For a new leader, it usually knows nothing about the next_index of others.

curp/proto/message.proto
```diff
@@ -48,8 +48,21 @@ message AppendEntriesResponse {
     bool success = 2;
```
rogercloud, line 48: We can add a field in the response to get the last committed entry index. It's used for the calibration.

curp/src/sync_manager.rs
markcty: A new leader will reset all server's next_index to the index of last log first.

curp/src/sync_manager.rs
markcty: Fixed

curp/proto/message.proto
markcty: Do you mean this paragraph in Raft paper?
> If desired, the protocol can be optimized to reduce the number of rejected AppendEntries RPCs. For example, when rejecting an AppendEntries request, the follower can include the term of the conﬂicting entry and the ﬁrst index it stores for that term. With this information, the leader can decrement nextIndex to bypass all of the conﬂicting entries in that term; one AppendEntries RPC will be required for each term with conﬂicting entries, rather than one RPC per entry. In practice, we doubt this optimization is necessary, since failures happen infrequently and it is unlikely that there will be many inconsistent entries.

markcty: We can't use the commitIndex of the follower because if the logs a follower have can be fewer than committed logs.

rogercloud: Yes, the followers' log may be fewer  than committed logs, we should also replicate these missing committed logs too. The field I suggest is used to adjust the calibration starting point faster.

curp/src/sync_manager.rs
rogercloud: It makes sense.

curp/src/server.rs
```diff
@@ -764,6 +847,55 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
+
+        // just grab a write lock because it's highly likely that term is updated and a vote is granted
+        let mut state = self.state.write();
+
+        // calibrate term
+        match req.term.cmp(&state.term) {
+            Ordering::Less => {
+                return Ok(tonic::Response::new(VoteResponse::new_reject(state.term)));
+            }
+            Ordering::Equal => {}
```
rogercloud, line 871: If we get a vote request with the same term from another leader while we're still a leader, what will happen?

curp/src/sync_manager.rs
```diff
@@ -203,22 +221,18 @@ impl<C: Command + 'static> SyncManager<C> {
...
-                            Arc::clone(&state),
-                            commit_trigger.clone(),
-                        )
-                    })
-                    .collect();
-                let _drop = tokio::spawn(async move {
-                    let _drop: Vec<_> = rpcs.collect().await;
-                });
+            if !state.read().is_leader() {
+                continue;
```
rogercloud, line 225: It should not happen, please log something for the later diagnose.

curp/src/server.rs
markcty: Should there be only one leader per term?

curp/src/sync_manager.rs
markcty: Added.

curp/src/server.rs
rogercloud: You're handling a vote here, and maybe that peer is trying to be a leader because it doesn't know the current one.

markcty: Then it will reject if it is a leader, because when its `voteFor` must be itself(set when it starts the election).

rogercloud (approved):

PR #82, opened 2022-11-18
[Curp] Test utils.

markcty:
Create `CurpGroup` wrapper that will generate a curp group with n nodes. Provide many utils for testing, such as `enable_noded`, `get_latest_term`, `get_leader`.


markcty:
Merged to #100

PR #86, opened 2022-11-23
[Curp] Fix concurrent bug

markcty:
The bug will happen in the following situation:
1. A client proposes a command to all servers.
2. The leader receives the command, sync to others, and committed the command.
3. When a follower has not yet received the proposal from the client, it notices that the command is committed and executes it. When the execution is over, it will need to remove the command from spec pool. However, since the follower has not yet received a proposal from the client, it will find no such command in the spec pool.

Suggested fix:
When a follower tries to remove a non-exist command from spec pool, it will register the command id in an internal buffer. When the proposal is finally received, it will directly remove the command. (The ready pool needs to gc after some time)

### DIFF
--- a/curp/src/server.rs
+++ b/curp/src/server.rs
@@ -12,7 +12,7 @@ use event_listener::{Event, EventListener};
 use futures::FutureExt;
 use opentelemetry::global;
 use parking_lot::lock_api::RwLockUpgradableReadGuard;
-use parking_lot::{Mutex, MutexGuard, RwLock};
+use parking_lot::{Mutex, RwLock};
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::time::Instant;
 use tokio::{net::TcpListener, task::JoinHandle};
@@ -20,6 +20,7 @@ use tokio_stream::wrappers::TcpListenerStream;
 use tracing::{debug, error, info, instrument, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
+use crate::gc::run_gc_tasks;
 use crate::rpc::{AppendEntriesRequest, AppendEntriesResponse, VoteRequest, VoteResponse};
 use crate::util::RwLockMap;
 use crate::{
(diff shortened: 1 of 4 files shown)


curp/src/server.rs
```diff
@@ -627,87 +669,77 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
+                ));
+            }
+            is_conflict
+        };
+
+        let resp = match (is_conflict, is_leader) {
+            // conflict and is leader
+            (true, true) => {
+                // the leader needs to sync cmd
+                match self.sync_to_others(self_term, &cmd, true) {
```
rogercloud, line 692: We can't change the logic this way, because we must make sure the first-coming request sync to others first. The order is kept for by holding the spec lock.

rogercloud:
Please add a test to coverage the error.

curp/src/server.rs
```diff
@@ -205,6 +206,61 @@ impl CmdBoardValue {
     }
 }
 
+/// The speculative pool that stores commands that might be executed speculatively
+#[derive(Debug)]
+pub(crate) struct SpeculativePool<C> {
+    /// Store
+    pub(crate) pool: VecDeque<C>,
+    /// Store the ids of commands that have completed execution and should be inserted into the speculative pool again
```
rogercloud, line 214: Store the ids of commands that have completed backend syncing, but not in the local speculative pool.

curp/src/server.rs
```diff
@@ -205,6 +206,61 @@ impl CmdBoardValue {
...
+    }
+
+    /// Check whether the command pool has conflict with the new command
+    pub(crate) fn has_conflict_with(&self, cmd: &C) -> bool {
+        self.pool.iter().any(|spec_cmd| spec_cmd.is_conflict(cmd))
+    }
+
+    /// Try to remove the command from spec pool.
+    /// There could be no such command due to several situations:
+    /// * When the proposal arrived, the command has conflict with speculative pool and was not stored in it.
```
rogercloud, line 242: "conflicted"

curp/src/server.rs
```diff
@@ -621,93 +655,79 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
+                    };
+                }
+
+                // the leader will sync the command to others while grabbing the lock so that the order of the command can be preserved
+                let sync_notify = match self.sync_to_others(term, &cmd, true) {
+                    Ok(notify) => notify,
+                    Err(err) => {
+                        return ProposeResponse::new_error(
+                            is_leader,
+                            self.state.read().term,
```
rogercloud, line 689: clone the previous term? Don;'t lock the state if not necessary.

curp/src/server.rs
```diff
@@ -621,93 +655,79 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
         &self,
```
rogercloud, line 655: For the propose, we need another TODO, check the duplicate propose as one client may send the same request many times.

curp/src/server.rs
```diff
@@ -621,93 +655,79 @@ impl<C: 'static + Command, CE: 'static + CommandExecutor<C>> Protocol<C, CE> {
...
-            })
-            .await
-            .map(tonic::Response::new)
-            .map_err(|err| tonic::Status::internal(format!("encode or decode error, {}", err)))
+
+                // finally, the command meets all the conditions for speculative execution, we will release the lock here
+                sync_notify
+            };
+
+            // speculative execution
```
rogercloud, line 708: Add an issue to optimize the sync channel blocking mechanism. For instance, we can provide an execution task to serializing the execution orders.

curp/src/sync_manager.rs
```diff
@@ -433,9 +432,9 @@ impl<C: Command + 'static> SyncManager<C> {
                             let _after_sync_result = cmd
                                 .after_sync(cmd_executor.as_ref(), i.numeric_cast())
                                 .await;
-                            if Protocol::<C, CE>::spec_remove_cmd(&spec, cmd.id()).is_none() {
-                                unreachable!("{:?} should be in the spec pool", cmd.id());
-                            }
+
+                            let cmd_id = cmd.id();
```
rogercloud, line 436: If the `cmd_id` isn't used later, remove the local variable.

rogercloud (approved):

PR #94, opened 2022-12-05
[Curp] Fix blocking caused by event listener wait() and improve kv_test

markcty:
l.wait() should only be called in non-async code. If it is called in async code, it will not hand the control flow back to tokio runtime like l.await and, therefore, block the tokio worker thread.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ scripts/*
 !scripts/Dockerfile
 !scripts/benchmark.sh
 !scripts/quick_start.sh
+
+.vscode
+.idea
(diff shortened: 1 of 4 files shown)


curp/src/channel/key_mpsc.rs
```diff
@@ -197,8 +197,8 @@ pub(crate) struct MpscKeyBasedReceiver<K, M> {
 impl<K: Eq + Hash + Clone + ConflictCheck, M> MpscKeyBasedReceiver<K, M> {
     /// Receive a message
     #[allow(dead_code)]
-    pub(crate) fn recv(&mut self) -> Result<MpscKeysMessage<K, M>, RecvError> {
-        self.inner.recv()
+    pub(crate) async fn recv(&mut self) -> Result<MpscKeysMessage<K, M>, RecvError> {
```
rogercloud, line 200: We have an async_recv function, what's the difference?

rogercloud (approved):

PR #105, opened 2022-12-14
handle client closes watch connection

themanforfree:


rogercloud (commented):
Please add tests to cover.

xline/tests/watch_test.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 10)]
+async fn test_watch_client_closes_connection() -> Result<(), Box<dyn Error>> {
+    let mut cluster = Cluster::new(3).await;
+    cluster.start().await;
+    let client = cluster.client().await;
+    let mut watch_client = client.watch_client();
+
+    let (watcher, mut stream) = watch_client.watch("foo", None).await?;
+    let handle = tokio::spawn(async move {
```
rogercloud, line 51: How to tell the correctness of tests? How do we know the server handles the client close operation correctly?

themanforfree: If we call this test before fixing it, the watch server will panic

xline/src/server/watch_server.rs
```diff
@@ -48,25 +94,30 @@ struct WatchHandle {
     watcher_map: HashMap<WatchId, Watcher>,
     /// Next available `WatchId`
     next_id: WatchId,
+    /// Stop tx
+    stop_tx: mpsc::Sender<()>,
```
Phoenix500526, line 98: If the messages transferred in the `stop` channel are meaningless, why don't we use a rendezvous channel here? Like `std::sync::mpsc`, `flume` supports 'rendezvous' channels, a bounded queue with a maximum capacity of zero will block senders until a receiver is available to take the message.

themanforfree: done

rogercloud (approved):

PR #113, opened 2022-12-26
Opt out heartbeat when contention is high

markcty:


markcty:
Will reopen with ut tests added.

Issue #115, opened 2022-12-27
[Curp] Failed to compile proto in that message.proto contains proto3 optional fields.


Phoenix500526:
When I executed the `cargo build --release` as described in the Contribution Guide document, it showed me a proto compile error as follows:
```bash
$ cargo build --release
   Compiling curp v0.1.0 (/home/Xline/curp)
error: failed to run custom build command for `curp v0.1.0 (/home/Xline/curp)`

Caused by:
  process didn't exit successfully: `/home/Xline/target/release/build/curp-755359d522eb8104/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Failed to compile proto, error is Custom { kind: Other, error: "protoc failed: message.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }', curp/build.rs:4:29
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I will issue a pr to solve this problem.

PR #116, opened 2022-12-27
fix proto compile error

Phoenix500526:
This pr aims to fix issue #115.  It added a build option `--experimental_allow_proto3_optional` to make the `cargo build --release` run successfully.

Signed-off-by: Phoeniix Zhao <Phoenix500526@163.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,8 +8,14 @@ Contribution Guide
 ```bash
 # Assume that rust compile environment installed, such as cargo, etc.
 
-# install dependency
-sudo apt install -y libprotobuf-dev protobuf-compiler
+# install dependency, requires protobuf-compiler >= 3.15
+git clone --branch v3.21.12  --recurse-submodules https://github.com/protocolbuffers/protobuf
+cd protobuf
+./autogen.sh
+./configure
+make -j
+sudo make install 
+
 
 # clone source code
 git clone https://github.com/datenlord/Xline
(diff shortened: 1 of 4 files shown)


markcty:
Thanks for the PR and the issue.
Since version 3.15, Protobuf has officially supported `optional` field without the need to enable it manually. So, maybe you can upgrade your local protobuf compiler to a newer version and it will compile correctly .

Phoenix500526:
> Thanks for the PR and the issue. Since version 3.15, Protobuf has officially supported `optional` field without the need to enable it manually. So, maybe you can upgrade your local protobuf compiler to a newer version and it will compile correctly .

Ok, I'll upgrade my protobuf compiler. But given that the default protobuf-compiler provided in ubuntu apt sources is out-of-date, I think we should modify the `Quick Start` section in the Contribution Guide to make it more crystal.

Phoenix500526:
> > Thanks for the PR and the issue. Since version 3.15, Protobuf has officially supported `optional` field without the need to enable it manually. So, maybe you can upgrade your local protobuf compiler to a newer version and it will compile correctly .
> 
> Ok, I'll upgrade my protobuf compiler. But given that the default protobuf-compiler provided in ubuntu apt sources is out-of-date, I think we should modify the `Quick Start` section in the Contribution Guide to make it more crystal.

I've just upgraded my local protobuf-compiler. And it did compile successfully. Adding this build option doesn't impact the correctness of the compiling process. In fact, it goes a long way towards being compatible with the old and the new versions of protobuf-compiler in ubuntu.

rogercloud:
@Phoenix500526  Please add an issue to update the quick start doc to upgrade the `protoc` version.

rogercloud (approved):

rogercloud (approved):

PR #119, opened 2022-12-28
Opt out heartbeat when contention is high and bg tasks ut

markcty:


curp/tests/common/curp_group.rs
```diff
@@ -76,7 +76,7 @@ impl CurpGroup {
             })
             .collect();
 
-        tokio::time::sleep(Duration::from_secs(2)).await;
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
rogercloud, line 79: How do we determine the waiting time? 3 seconds are really enough?

curp/src/server/bg_tasks.rs
```diff
@@ -44,12 +50,11 @@ pub(super) async fn run_bg_tasks<C: Command + 'static, CE: 'static + CommandExec
...
     let others = state.read().others.clone();
-    let connects = rpc::try_connect(
-        others,
-        #[cfg(test)]
-        reachable,
-    )
-    .await;
+
+    #[cfg(test)]
+    let connects = Conn::try_connect_test(others, reachable).await;
```
rogercloud, line 55: Are we going to replace "reachable" to the "injecture fun" in the later commits? If yes, please leave a comment here.

curp/src/server/bg_tasks.rs
```diff
@@ -228,20 +235,26 @@ async fn send_log_until_succeed<C: Command + 'static>(
...
 ) {
-    let role_trigger = state.read().role_trigger();
+    let (role_trigger, hb_reset_trigger) =
+        state.map_read(|state_r| (state_r.role_trigger(), state_r.heartbeat_reset_trigger()));
     #[allow(clippy::integer_arithmetic)] // tokio internal triggered
     loop {
+        loop {
+            tokio::select! {
+                _ = tokio::time::sleep(HEARTBEAT_INTERVAL) => break,
+                _ = hb_reset_trigger.listen() => {}
```
rogercloud, line 249: I think "Event" is not the perfect data structure here. Imagining that the system is quite busy, hundreds of "append_entry" are called, we need to enter listen hundreds of times.

Even using "Event", the listen() function only returns the listener, the "await" is required.

curp/src/rpc/connect.rs
```diff
@@ -0,0 +1,267 @@
...
+        #[cfg(test)]
+        if !self.reachable.load(Ordering::Relaxed) {
+            return Err(ProposeError::RpcError("unreachable".to_owned()));
+        }
+
+        let mut client = self.get().await?;
+        let mut req = tonic::Request::new(request);
+        req.set_timeout(timeout);
+        client.fetch_leader(req).await.map_err(Into::into)
+    }
```
themanforfree, line 209: add a new line

curp/src/rpc/connect.rs
```diff
@@ -0,0 +1,267 @@
...
+            Arc::new(Connect {
+                id,
+                rpc_connect: RwLock::new(conn),
+                addr,
+                #[cfg(test)]
+                reachable: Arc::new(AtomicBool::new(true)),
+            })
+        })
+        .collect()
+    }
```
themanforfree, line 236: add a new line

curp/src/rpc/connect.rs
```diff
@@ -0,0 +1,267 @@
...
+            .map(|client| {
+                *connect_write = Ok(client.clone());
+                client
+            })?;
+        *connect_write = Ok(client.clone());
+        Ok(client)
+    }
+
+    /// send "propose" request
+    async fn propose(
```
themanforfree, line 124: add `#[instrument(skip(self), name = "client propose")]` for tracing

curp/src/rpc/connect.rs
```diff
@@ -0,0 +1,267 @@
...
+        let mut client = self.get().await?;
+        let mut req = tonic::Request::new(request);
+        req.set_timeout(timeout);
+        req.metadata_mut().inject_current();
+        client.propose(req).await.map_err(Into::into)
+    }
+
+    /// send "wait synced" request
+    #[allow(dead_code)] // false positive report
+    async fn wait_synced(
```
themanforfree, line 143: remove `#[allow(dead_code)]`, we use this in slow_round now,
and add `#[instrument(skip(self), name = "client propose")]` for tracing

curp/src/server/bg_tasks.rs
markcty: > Even using "Event", the listen() function only returns the listener, the "await" is required.

`tokio::select!` will call await for us.

> I think "Event" is not the perfect data structure here. Imagining that the system is quite busy, hundreds of "append_entry" are called, we need to enter listen hundreds of times.

The listener returned by `listen()` will be consumed every time we call await to it. So we had to call it every time. I guess the alternative way is to use time to check if a heartbeat is needed. This method would require us to acquire system time when `append_entries` is called. I suspect it would be more efficient.

curp/src/server/bg_tasks.rs
markcty: Sure.

curp/tests/common/curp_group.rs
markcty: Because the time a follower needs to wait before the election starts is 1~2 second. 3 seconds are almost 100% enough because vote split rarely fails and 3 seconds is enough for nearly 2 rounds of elections.

curp/src/rpc/connect.rs
```diff
@@ -0,0 +1,267 @@
...
+    async fn propose(
+        &self,
+        request: ProposeRequest,
+        timeout: Duration,
+    ) -> Result<tonic::Response<ProposeResponse>, ProposeError>;
+
+    /// send "wait synced" request
+    async fn wait_synced(
+        &self,
+        req: WaitSyncedRequest,
```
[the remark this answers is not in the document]
markcty: Fixed.

curp/src/rpc/connect.rs
markcty: Sure.

curp/src/rpc/connect.rs
markcty: Sure.

curp/src/rpc/connect.rs
markcty: Added.

curp/src/rpc/connect.rs
markcty: Fixed.

curp/src/server/bg_tasks.rs
rogercloud: I see, the new "Listener" won't get previous notifications, so it's not a hot point.

rogercloud (dismissed):

rogercloud (approved):

themanforfree (approved):

PR #123, opened 2022-12-29
fix panic in a benchmark

Phoenix500526:
A panic happened in the first bench case when I run the `benchmark.sh`.  That's because the benchmark program sent requests to the cluster while it is not yet booted up.


rogercloud:
please squash commits as 10 commits don't make sense for such a small patch.

rogercloud (approved):

PR #125, opened 2023-01-03
Add broadcast for leader changes

markcty:


markcty:
Also need the broadcast in server. WIP

PR #128, opened 2023-01-03
refactor args, use name as server id

themanforfree:


xline/src/main.rs
```diff
@@ -162,6 +156,19 @@ struct ServerArgs {
...
+    for pair in s.split(',') {
+        if let Some((id, addr)) = pair.split_once('=') {
+            let _ignore = map.insert(id.to_owned(), addr.to_owned());
+        } else {
+            return Err("parse members error".to_owned());
+        }
+    }
+    Ok(map)
+}
+
```
rogercloud, line 171: Can we deduplicate the code with that in benchmark?

themanforfree: Add another utils crate? or let lock_utils become a more universal crate?

Phoenix500526: I prefer to let `lock_utils` become more universal. The name of `lock_utils` seems to be too concrete,  or we can rename it `utils`. The `lock_utils` is one of the `utils`.

Phoenix500526 (approved):

xline/src/server/xline_server.rs
```diff
@@ -31,23 +33,14 @@ type CurpServer = Rpc<Command>;
 #[allow(dead_code)] // Remove this after feature is completed
 #[derive(Debug)]
 pub struct XlineServer {
-    /// Server name
-    name: String,
-    /// Address of peers
-    peers: Vec<SocketAddr>,
+    /// State of current node
+    state: Arc<RwLock<State>>,
```
rogercloud, line 37: why is it protected in a rwlock? I can't find a place to write to it. All references are read-only accesses.

themanforfree: when I can get new leader_id from curp, the state will be updated, but in this PR it is read-only

rogercloud: OK, I see.

rogercloud (approved):

Phoenix500526 (approved):

PR #130, opened 2023-01-04
Broadcast on leader changing

markcty:


curp/src/server/state.rs
```diff
@@ -50,6 +50,8 @@ pub(super) struct State<C: Command + 'static> {
     pub(super) cmd_board: Arc<Mutex<CommandBoard>>,
     /// Last time a rpc is received.
     pub(super) last_rpc_time: Arc<RwLock<Instant>>,
+    /// Leader changes tx
+    leader_tx: broadcast::Sender<ServerId>,
```
themanforfree, line 54: the server didn't send the `ServerId` of new leader

curp/src/server/mod.rs
```diff
@@ -557,9 +564,7 @@ impl<C: 'static + Command> Protocol<C> {
         if req.term > state.term {
             state.update_to_term(req.term);
```
themanforfree, line 565: `update_to_term` will  change `leader_id` to `None`, other places where called this function may not immediately update the `leader_id`, e.g.: [link](https://github.com/themanforfree/Xline/blob/780f5bbc537d786b71cfe0e143448938cf49adea/curp/src/server/bg_tasks.rs#L195), so there will be a period of time we don't know who the leader is, should we notify this case to xline?

markcty: Fixed

curp/src/server/mod.rs
```diff
@@ -189,6 +189,13 @@ impl<C: Command + 'static> Rpc<C> {
             .await?;
         Ok(())
     }
+
+    /// Get a subscriber for leader changes
+    #[inline]
+    #[must_use]
+    pub fn leader_rx(&self) -> broadcast::Receiver<Option<ServerId>> {
```
rogercloud, line 196: This subscriber is never used in this change. Why?

themanforfree (approved):

curp/src/server/mod.rs
markcty: I guess @themanforfree will add the usage in his fork in the next PR.

themanforfree: I think we can add tests for this, like you did for client in this PR

markcty: Sure. Added.

themanforfree (approved):

rogercloud (approved):

PR #138, opened 2023-01-06
Refactor/range

themanforfree:
based on #128 and #130

### DIFF
--- a/xline/src/storage/kvstore.rs
+++ b/xline/src/storage/kvstore.rs
@@ -85,6 +85,11 @@ impl KvStore {
     pub(crate) fn kv_watcher(&self) -> Arc<KvWatcher> {
         Arc::clone(&self.kv_watcher)
     }
+
+    /// Get revision
+    pub(crate) fn revision(&self) -> i64 {
+        self.inner.revision()
+    }
 }
 
 impl KvStoreBackend {
--- a/xline/src/storage/kvstore.rs
+++ b/xline/src/storage/kvstore.rs
@@ -12,6 +12,7 @@ use super::{
     db::DB,
     index::{Index, IndexOperate},
     kvwatcher::KvWatcher,
+    req_ctx::RequestCtx,
 };
 use crate::{
     header_gen::HeaderGenerator,
@@ -22,7 +23,6 @@ use crate::{
         TxnRequest, TxnResponse,
     },
     server::command::{CommandResponse, KeyRange, SyncResponse},
-    storage::req_ctx::RequestCtx,
 };
 
 /// Default channel size
@@ -84,11 +84,6 @@ impl KvStore {
     /// Get KV watcher
     pub(crate) fn kv_watcher(&self) -> Arc<KvWatcher> {
         Arc::clone(&self.kv_watcher)
-    }
-
-    /// Get revision
-    pub(crate) fn revision(&self) -> i64 {
-        self.inner.revision()
     }
 }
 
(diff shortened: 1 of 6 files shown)


xline/src/storage/kvstore.rs
```diff
@@ -187,55 +208,112 @@ impl KvStoreBackend {
...
     fn handle_range_request(&self, req: &RangeRequest) -> RangeResponse {
-        let mut kvs = self.get_range(&req.key, &req.range_end, req.revision);
-        debug!("handle_range_request kvs {:?}", kvs);
+        debug!("handle_range_request kvs");
+        let limit = if (req.sort_order() != SortOrder::None)
+            || (req.max_mod_revision != 0)
+            || (req.min_mod_revision != 0)
+            || (req.max_create_revision != 0)
+            || (req.min_create_revision != 0)
+            || (req.limit == 0)
```
rogercloud, line 278: limit is just limit. Why do we get unlimited results if we've set max_create_revision?

xline/src/storage/kvstore.rs
```diff
@@ -187,55 +208,112 @@ impl KvStoreBackend {
...
+        {
+            0 // get all from storage then sort and filter
+        } else {
+            req.limit.overflow_add(1) // get one extra for more flag
```
rogercloud, line 282: What's is the "more" flag?

xline/tests/kv_test.rs
```diff
@@ -155,6 +156,13 @@ async fn test_kv_get() -> Result<(), Box<dyn Error>> {
         assert!(is_identical);
     }
 
+    // test range redirect
```
rogercloud, line 159: I recommend to put the redirect test to another function.

xline/src/storage/kvstore.rs
themanforfree: because we need filter results before limit them, if we just get limited results, The final result may be less than limit

xline/src/storage/kvstore.rs
themanforfree: There are "more" results not returned to the client because of limit

xline/src/storage/kvstore.rs
```diff
@@ -192,55 +208,112 @@ impl KvStoreBackend {
             .collect()
     }
 
+    /// Sort kvs by sort target and order
+    fn sort_kvs(kvs: &mut [KeyValue], sort_order: SortOrder, sort_target: SortTarget) {
+        match (sort_target, sort_order) {
+            (SortTarget::Key, SortOrder::None) => {}
+            (SortTarget::Key, SortOrder::Ascend) => {
```
Phoenix500526, line 215: We can cover these codes from line 215 to line 217 in `test_keys_only` by constructing requests with different `SortOrder`. It can help improve our test coverage.

xline/src/storage/kvstore.rs
```diff
@@ -522,3 +595,87 @@ impl KvStoreBackend {
...
+        let response = store.inner.handle_range_request(&request);
+        assert_eq!(response.kvs.len(), 0);
+        assert_eq!(response.count, 0);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[allow(clippy::indexing_slicing)] // Checked before use index
+    async fn test_range_filter() -> Result<(), Box<dyn Error>> {
```
Phoenix500526, line 643: Please enrich this test case to cover different range request parameters, like different `SortOrder` on different `SortTarget`.

rogercloud (dismissed):

xline/src/storage/kvstore.rs
themanforfree: added a  new test

xline/src/storage/kvstore.rs
themanforfree: added a  new test

Phoenix500526 (approved):

rogercloud (approved):

Phoenix500526 (approved):

PR #144, opened 2023-01-11
fix scripts, docs and CI

themanforfree:

fix scripts, docs and CI after refactor args
fix bugs
No

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -5,11 +5,11 @@
     ```bash
     # Run in 3 terminals. If you want more logs, add `RUST_LOG=debug` before the command.
 
-    ./xline --name node1 --cluster-peers 127.0.0.1:2380 127.0.0.1:2381 --self-ip-port 127.0.0.1:2379 --leader-ip-port 127.0.0.1:2379 
+    ./xline --name node1 --members node1=127.0.0.1:2379,node2=127.0.0.1:2380,node3=127.0.0.1:2381 --is-leader
 
-    ./xline --name node2 --cluster-peers 127.0.0.1:2379 127.0.0.1:2381 --self-ip-port 127.0.0.1:2380 --leader-ip-port 127.0.0.1:2379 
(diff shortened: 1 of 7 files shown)


USAGE.md
```diff
@@ -5,16 +5,19 @@
...
 3. Download or build `etcdctl` from [etcd](https://github.com/etcd-io/etcd) project.
 4. Use `etcdctl` to operate the cluster:
     ```bash
     etcdctl --endpoints=http://127.0.0.1:2379 put foo bar
 
     etcdctl --endpoints=http://127.0.0.1:2379 get foo
-    ```
\ No newline at end of file
+    ```
+
```
markcty, line 21: Blank lines?

markcty (changes requested):

USAGE.md
```diff
@@ -5,16 +5,19 @@
     ```bash
     # Run in 3 terminals. If you want more logs, add `RUST_LOG=debug` before the command.
 
-    ./xline --name node1 --cluster-peers 127.0.0.1:2380 127.0.0.1:2381 --self-ip-port 127.0.0.1:2379 --leader-ip-port 127.0.0.1:2379 
+    ./xline --name node1 --members node1=127.0.0.1:2379,node2=127.0.0.1:2380,node3=127.0.0.1:2381
```
Phoenix500526, line 8: Is here missing the `--is-leader` option?

USAGE.md
themanforfree: removed

USAGE.md
themanforfree: we can  run without this flag now, curp will elect a leader

markcty: I guess we still need it. Because as we discussed yesterday, curp won't start an election automatically when it starts.

Phoenix500526: Apart from the reasons that markcty has mentioned, I think it's a good idea to add this option, as doing so can provide developers who are green to Xline with more details on how to use it. Just my two cents.

themanforfree: added

Phoenix500526 (approved):

PR #146, opened 2023-01-11
Recovery

markcty:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
Curp will recovery on leader crashing or restart.

* what changes does this pull request make?
New leader will recover speculatively executed cmds, old leader will reset its state, client will re-propose. Server will not be allowed to vote first time it starts.

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #150, opened 2023-01-16
Recovery on restarting

markcty:

The server will recover on restarting

Now a server won't initiate an election on starting


markcty:
Based on #127 and #146.

PR #153, opened 2023-01-16
Features/lock

themanforfree:

Support lock service
Support lock service
No

### DIFF
--- a/xline/src/client/mod.rs
+++ b/xline/src/client/mod.rs
@@ -1,8 +1,9 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Debug};
 
 use curp::{client::Client as CurpClient, cmd::ProposeId};
 use etcd_client::{
-    AuthClient, Client as EtcdClient, LeaseKeepAliveStream, LeaseKeeper, WatchClient,
+    AuthClient, Client as EtcdClient, KvClient, LeaseClient, LeaseKeepAliveStream, LeaseKeeper,
+    LockClient, WatchClient,
 };
 use utils::config::ClientTimeout;
 use uuid::Uuid;
@@ -30,7 +31,6 @@ pub mod errors;
 pub mod kv_types;
 
 /// Xline client
-#[allow(missing_debug_implementations)] // EtcdClient doesn't implement Debug
 pub struct Client {
     /// Name of the client
     name: String,
@@ -40,6 +40,17 @@ pub struct Client {
     etcd_client: EtcdClient,
     /// Use curp client to send requests when true
     use_curp_client: bool,
+}
+
+impl Debug for Client {
+    #[inline]
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Client")
+            .field("name", &self.name)
+            .field("use_curp_client", &self.use_curp_client)
+            .field("curp_client", &self.curp_client)
+            .finish()
+    }
 }
 
 impl Client {
@@ -238,15 +249,33 @@ impl Client {
         Ok(response.into())
     }
 
+    /// Gets a kv client.
+    #[inline]
+    pub fn kv_client(&self) -> KvClient {
+        self.etcd_client.kv_client()
+    }
+
     /// Gets an auth client.
     #[inline]
-    pub fn auth_client(&mut self) -> AuthClient {
+    pub fn auth_client(&self) -> AuthClient {
         self.etcd_client.auth_client()
     }
 
-    /// Gets an watch client.
+    /// Gets a watch client.
     #[inline]
-    pub fn watch_client(&mut self) -> WatchClient {
+    pub fn watch_client(&self) -> WatchClient {
         self.etcd_client.watch_client()
+    }
+
+    /// Gets a lock client.
+    #[inline]
+    pub fn lock_client(&self) -> LockClient {
+        self.etcd_client.lock_client()
+    }
+
+    /// Gets a lease client.
+    #[inline]
+    pub fn lease_client(&self) -> LeaseClient {
+        self.etcd_client.lease_client()
     }
 }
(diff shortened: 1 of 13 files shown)


xline/src/storage/kvwatcher.rs
```diff
@@ -203,10 +202,9 @@ pub(crate) trait KvWatcherOps {
     fn cancel(&self, watcher: &Watcher) -> i64;
 }
 
-#[async_trait::async_trait]
 impl KvWatcherOps for KvWatcher {
     /// Create a watch to KV store
-    async fn watch(
+    fn watch(
```
Phoenix500526, line 207: Why should we remove `async` here?

xline/src/server/kv_server.rs
```diff
@@ -433,7 +433,7 @@ impl Kv for KvServer {
     ) -> Result<tonic::Response<TxnResponse>, tonic::Status> {
         debug!("Receive TxnRequest {:?}", request);
         Self::check_txn_request(request.get_ref())?;
-        let is_fast_path = true;
+        let is_fast_path = false; // lock need revision of txn
         let (cmd_res, sync_res) = self.propose(request, is_fast_path).await?;
```
markcty, line 438: If we use slow path here, should `sync_res` always be Some(...)? Then we should at least give a warning if it's None. I think a better way is to seperate self.propose into two functions. E.g., propose_slow will always `SyncResponse`, while propose_fast will return `Option<SyncResponse>`

xline/src/server/xline_server.rs
```diff
@@ -51,6 +52,8 @@ pub struct XlineServer {
     lease_storage: Arc<LeaseStore>,
     /// Consensus client
     client: Arc<Client<Command>>,
+    /// Xline client
+    xline_client: Arc<Mutex<crate::client::Client>>,
```
markcty, line 56: Is it necessary to put xlien_client in a Mutex? I see no mutable usage. Maybe `Arc<Client>` is enough?

xline/src/client/mod.rs
```diff
@@ -238,15 +249,33 @@ impl Client {
         Ok(response.into())
     }
 
+    /// Gets an kv client.
+    #[inline]
+    pub fn kv_client(&self) -> KvClient {
```
markcty, line 254: Should we expose these sub clients? We have already exposed more high-level interface such as put, range, delete. I guess the kv_client should be purely internal.

xline/src/server/lock_server.rs
```diff
@@ -26,54 +40,218 @@ pub(crate) struct LockServer {
...
+            })?;
+            Ok((cmd_res, Some(sync_res)))
+        }
     }
 
-    /// Generate propose id
-    fn generate_propose_id(&self) -> ProposeId {
-        ProposeId::new(format!("{}-{}", self.name, Uuid::new_v4()))
+    /// Crate txn for try acquire lock
+    fn try_acquire_txn(prefix: &str, lease_id: i64) -> TxnRequest {
```
Phoenix500526, line 128: Why do we name this function as `try_acquire_txn`? The prefix `try_` looks like it should return a `Result`. In addition, this function only constructs a `txn` object. It should not fail.

markcty (changes requested):

xline/src/client/mod.rs
markcty: So do following sub clients.

xline/tests/lock_test.rs
```diff
@@ -0,0 +1,33 @@
+mod common;
+
+use std::{error::Error, time::Duration};
+
+use common::Cluster;
+use tokio::time;
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 10)]
+async fn test_lock() -> Result<(), Box<dyn Error>> {
```
Phoenix500526, line 9: Please add a test to cover the deadlock case to ensure that the lease expired will release the lock

xline/src/client/mod.rs
```diff
@@ -238,15 +249,33 @@ impl Client {
         Ok(response.into())
     }
 
+    /// Gets an kv client.
```
Phoenix500526, line 252: Typo. It should be "Gets a kv client"

xline/src/client/mod.rs
```diff
@@ -238,15 +249,33 @@ impl Client {
...
+    }
+
     /// Gets an auth client.
     #[inline]
-    pub fn auth_client(&mut self) -> AuthClient {
+    pub fn auth_client(&self) -> AuthClient {
         self.etcd_client.auth_client()
     }
 
     /// Gets an watch client.
```
Phoenix500526, line 264: Same above

xline/src/client/mod.rs
```diff
@@ -238,15 +249,33 @@ impl Client {
...
     #[inline]
-    pub fn watch_client(&mut self) -> WatchClient {
+    pub fn watch_client(&self) -> WatchClient {
         self.etcd_client.watch_client()
     }
+
+    /// Gets an lock client.
```
Phoenix500526, line 270: Same above

xline/src/client/mod.rs
```diff
@@ -238,15 +249,33 @@ impl Client {
...
+    #[inline]
+    pub fn lock_client(&self) -> LockClient {
+        self.etcd_client.lock_client()
+    }
+
+    /// Gets an lease client.
```
Phoenix500526, line 276: Same above

xline/src/server/kv_server.rs
themanforfree: In the future we may let the client decide whether to use fast path，so I think current logic is better for this.

xline/src/server/xline_server.rs
Phoenix500526: Is it a good idea to add this dependency to `client` here? I mean, the xline client is only used in `test` and `benchmark` before implementing the `Lock` feature.

xline/src/client/mod.rs
themanforfree: This is just a temporary client implementation, it will be refactor in the future.

Phoenix500526 (approved):

markcty (approved):

PR #154, opened 2023-01-17
Recovery on crash

markcty:

Recovery on crash

Now the server won't start an election the first time it boots.


markcty:
Based on #146

markcty:
Added to #146

PR #156, opened 2023-01-18
refactor and improve xline server

markcty:
1. Watcher should be an inner struct in the storage module. It should not appear in the server module. WatchServer only needs to maintain watch_id, no need to maintain watcher.
2. Some renaming
3. Use a RwLock to store watchers so that different watchers can send concurrently
4. Unnecessary Arc<Inner>: Watcher has an inner of type Arc<WatcherInner>. When Watcher needs to be shared, we previously Clone on Watcher. Just use Arc<Watcher> is both clearer and simpler.
5. Improve filter events performance: previously events are filtered by key_range twice, now only once
6. Watch id should be assigned from 1 instead of 0 because 0 means auto-generated, which might be confusing to the client
7. Reduce code duplication
8. Use atomici64 for the revision number.
9. Fix in auth backend concurrent bugs due to the revision number

Refactor and improve KV watch

Refactors

No.

### DIFF
--- a/xline/src/storage/auth_store/backend.rs
+++ b/xline/src/storage/auth_store/backend.rs
@@ -199,11 +199,11 @@ impl AuthStoreBackend {
     fn create_permission_cache(&self) {
         let mut permission_cache = PermissionCache::new();
         for user in self.get_all_users() {
-            let user_permisiion = self.get_user_permissions(&user);
+            let user_permission = self.get_user_permissions(&user);
             let username = String::from_utf8_lossy(&user.name).to_string();
             let _ignore = permission_cache
                 .user_permissions
-                .insert(username, user_permisiion);
+                .insert(username, user_permission);
         }
         self.permission_cache
             .map_write(|mut cache| *cache = permission_cache);
@@ -211,7 +211,7 @@ impl AuthStoreBackend {
 
     /// get user permissions
     fn get_user_permissions(&self, user: &User) -> UserPermissions {
-        let mut user_permisiion = UserPermissions::new();
+        let mut user_permission = UserPermissions::new();
         for role_name in &user.roles {
             let role = match self.get_role(role_name) {
                 Ok(role) => role,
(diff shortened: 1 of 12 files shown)


xline/src/storage/authstore/backend.rs
```diff
@@ -714,143 +715,131 @@ impl AuthStoreBackend {
...
         self.put(AUTH_ENABLE_KEY.to_vec(), vec![0], revision, 0);
         self.enabled.store(false, AtomicOrdering::Release);
-        true
     }
 
     /// Sync `AuthStatusRequest` and return whether authstore is changed.
-    fn sync_auth_status_request(_req: &AuthStatusRequest) -> bool {
-        false
-    }
+    fn sync_auth_status_request(_req: &AuthStatusRequest) {}
```
Phoenix500526, line 817: What's the point of existence of this function? It has neither a return value nor a function body

xline/tests/auth_test.rs
```diff
@@ -186,12 +186,12 @@ async fn test_auth_wrong_password() -> Result<(), Box<dyn Error>> {
     .await;
     assert!(result.is_err());
 
-    let result = etcd_client::Client::connect(
-        vec![cluster.addrs()["server0"].to_string()],
-        Some(ConnectOptions::new().with_user("root", "123")),
-    )
-    .await;
-    assert!(result.is_ok());
+    // let result = etcd_client::Client::connect(
```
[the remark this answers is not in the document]
markcty: My bad. Forgot to remove the comment after debugging

xline/src/storage/authstore/backend.rs
markcty: I guess @themanforfree wants to keep it so that all the sync_... functions have the same form.

themanforfree (approved):

xline/src/storage/auth_store/backend.rs
```diff
@@ -947,20 +929,20 @@ impl AuthStoreBackend {
...
-            Err(_) => return false,
+            Err(_) => return,
         };
         let idx = match user.roles.binary_search(&req.role) {
             Ok(idx) => idx,
-            Err(_) => return false,
+            Err(_) => return,
         };
         let _ignore = user.roles.remove(idx);
+        let revision = self.revision.get();
```
themanforfree, line 945: this should be revision.next()

markcty: Fixed.

themanforfree (approved):

Phoenix500526 (approved):

PR #164, opened 2023-02-01
abstract DB as a trait

themanforfree:

Abstract a storage layer to make persistent storage easier
Abstract the storage layer as a trait similar to `HashMap`, and use `Mutex<HashMap<Vec<u8>,Vet<u8>>>` to provide a default in-memory implementation
No

### DIFF
--- a/xline/src/lib.rs
+++ b/xline/src/lib.rs
@@ -127,3 +127,5 @@ pub mod server;
 mod state;
 /// Storage module
 mod storage;
+
+pub use storage::memory::Memory;
--- a/xline/src/server/xline_server.rs
+++ b/xline/src/server/xline_server.rs
@@ -28,7 +28,7 @@ use crate::{
         LockServer as RpcLockServer, WatchServer as RpcWatchServer,
     },
     state::State,
-    storage::{AuthStore, KvStore, LeaseStore},
+    storage::{storage_api::StorageApi, AuthStore, KvStore, LeaseStore},
 };
 
 /// Default channel size
@@ -40,13 +40,16 @@ type CurpServer = Rpc<Command>;
 /// Xline server
 #[allow(dead_code)] // Remove this after feature is completed
 #[derive(Debug)]
-pub struct XlineServer {
+pub struct XlineServer<S>
+where
+    S: StorageApi,
+{
     /// State of current node
     state: Arc<State>,
     /// Kv storage
-    kv_storage: Arc<KvStore>,
+    kv_storage: Arc<KvStore<S>>,
     /// Auth storage
-    auth_storage: Arc<AuthStore>,
+    auth_storage: Arc<AuthStore<S>>,
     /// Lease storage
     lease_storage: Arc<LeaseStore>,
     /// Consensus client
@@ -59,7 +62,10 @@ pub struct XlineServer {
     id_gen: Arc<IdGenerator>,
 }
 
-impl XlineServer {
+impl<S> XlineServer<S>
+where
+    S: StorageApi + Clone,
+{
     /// New `XlineServer`
     ///
     /// # Panics
@@ -73,6 +79,7 @@ impl XlineServer {
         key_pair: Option<(EncodingKey, DecodingKey)>,
         server_timeout: ServerTimeout,
         client_timeout: ClientTimeout,
+        storage: S,
     ) -> Self {
         // TODO: temporary solution, need real cluster id and member id
         let header_gen = Arc::new(HeaderGenerator::new(0, 0));
@@ -92,11 +99,13 @@ impl XlineServer {
             lease_cmd_tx.clone(),
             del_rx,
             Arc::clone(&header_gen),
+            storage.clone(), // TODO: implement bucket storage
         ));
         let auth_storage = Arc::new(AuthStore::new(
             lease_cmd_tx,
             key_pair,
             Arc::clone(&header_gen),
+            storage,
         ));
         let client = Arc::new(Client::<Command>::new(all_members.clone(), client_timeout).await);
         Self {
@@ -190,14 +199,15 @@ impl XlineServer {
 
     /// Init `KvServer`, `LockServer`, `LeaseServer`, `WatchServer` and `CurpServer`
     /// for the Xline Server.
+    #[allow(clippy::type_complexity)] // it is easy to read
     fn init_servers(
         &self,
     ) -> (
-        KvServer,
-        LockServer,
-        Arc<LeaseServer>,
-        AuthServer,
-        WatchServer,
+        KvServer<S>,
+        LockServer<S>,
+        Arc<LeaseServer<S>>,
+        AuthServer<S>,
+        WatchServer<S>,
         CurpServer,
     ) {
         let curp_server = CurpServer::new(
--- a/xline/src/storage/auth_store/backend.rs
+++ b/xline/src/storage/auth_store/backend.rs
@@ -47,6 +47,7 @@ use crate::{
         index::{Index, IndexOperate},
         lease_store::{Lease, LeaseMessage},
         req_ctx::RequestCtx,
+        storage_api::StorageApi,
     },
 };
 
@@ -62,11 +63,14 @@ pub(crate) const ROOT_USER: &str = "root";
 pub(crate) const ROOT_ROLE: &str = "root";
 
 /// Auth store inner
-pub(crate) struct AuthStoreBackend {
+pub(crate) struct AuthStoreBackend<S>
+where
+    S: StorageApi,
+{
     /// Key Index
     index: Index,
     /// DB to store key value
-    db: DB,
+    db: DB<S>,
     /// Revision
     revision: RevisionNumber,
     /// Speculative execution pool. Mapping from propose id to request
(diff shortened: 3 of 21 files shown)


utils/src/storage.rs
```diff
@@ -0,0 +1,123 @@
+use std::{collections::HashMap, convert::Infallible};
+
+use parking_lot::Mutex;
+
+/// persistent storage abstraction
+pub trait Storage {
+    /// Error type returned by storage
+    type Error: std::error::Error + Send + Sync + 'static;
+
+    /// Ger value by key from storage
```
Phoenix500526, line 10: Typo.

utils/src/storage.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// # Errors
+    ///
+    /// if error occurs in storage, return `Err(error)`
+    fn remove(&self, key: impl AsRef<[u8]>) -> Result<Option<Vec<u8>>, Self::Error>;
+}
+
+/// default storage implementation
+#[derive(Debug, Default)]
+pub struct Memory(Mutex<HashMap<Vec<u8>, Vec<u8>>>);
```
Phoenix500526, line 59: It's better to use `RwLock` here.

xline/src/storage/revision.rs
```diff
@@ -69,3 +81,18 @@ impl KeyRevision {
...
     }
 }
+
+#[cfg(test)]
+mod test {
+    use prost::bytes::Buf;
+
+    use super::*;
+    #[test]
+    fn it_works() {
```
Phoenix500526, line 91: What does the `it` refer to in `it_works`? Please modify the test name to make it more meaningful. You can rename it as `encode_should_work` or something like that.

xline/src/storage/auth_store/backend.rs
```diff
@@ -253,30 +255,32 @@ impl AuthStoreBackend {
...
-        match self.get(&key) {
-            Some(kv) => Ok(User::decode(kv.value.as_slice()).unwrap_or_else(|e| {
-                panic!(
-                    "Failed to decode user from kv value, error: {:?}, kv: {:?}",
-                    e, kv
-                )
-            })),
-            None => Err(ExecuteError::InvalidCommand("user not found".to_owned())),
-        }
+        self.get(&key)?.map_or_else(
```
Phoenix500526, line 267: Can we keep the code style consistent with that in `get_role`?

xline/src/storage/auth_store/backend.rs
```diff
@@ -726,140 +755,124 @@ impl AuthStoreBackend {
...
         if self.is_enabled() {
-            return;
+            return Ok(());
         }
         let revision = self.revision.next();
-        self.put(AUTH_ENABLE_KEY.to_vec(), vec![1], revision, 0);
+        self.put(AUTH_ENABLE_KEY.to_vec(), vec![1], revision, 0)?;
         self.enabled.store(true, AtomicOrdering::Release);
-        self.create_permission_cache();
+        self.create_permission_cache()?;
```
Phoenix500526, line 845: `create_permission_cache` returns `Result<(), ExecuteError>`, so we can use `self.create_permission_cache()` instead of `self.create_permission_cache()?; Ok(())`

utils/src/storage.rs
```diff
@@ -0,0 +1,123 @@
+use std::{collections::HashMap, convert::Infallible};
```
Phoenix500526, line 1: `utils` is an abbreviation for utilities. Does it make sense to put storage here? Or, we should create an individual module to represent the storage layer.

xline/src/storage/auth_store/backend.rs
```diff
@@ -66,7 +66,8 @@ pub(crate) struct AuthStoreBackend {
     /// Key Index
     index: Index,
     /// DB to store key value
-    db: DB,
+    // db: DB<Mutex<HashMap<Vec<u8>, Vec<u8>>>>,
+    db: DB<Memory>,
```
Phoenix500526, line 70: I think we shouldn't hardcode `DB<Memory>` here. 
First, We can add a generic parameter for `AuthStoreBackend`, like `AuthStoreBackend<S: Storage>` and let `db` be an object in type `S`. So we can inject the given storage backend which comes from a config file or command line arguments into XlineServer.
Second, We can define a `MemStore` in the `storage` module, and implement `Storage` trait for it.

xline/src/storage/auth_store/backend.rs
```diff
@@ -107,7 +108,7 @@ impl AuthStoreBackend {
     ) -> Self {
         Self {
             index: Index::new(),
-            db: DB::new(),
+            db: DB::new(Memory::new()),
```
Phoenix500526, line 111: Refer to the previous comment.

xline/src/storage/kv_store.rs
```diff
@@ -45,7 +46,7 @@ pub(crate) struct KvStoreBackend {
     /// Key Index
     index: Index,
     /// DB to store key value
-    db: DB,
+    db: DB<Memory>,
```
Phoenix500526, line 49: Refer to the comment in `AuthStoreBackend`

utils/src/storage.rs
themanforfree: moved to xline, my origin idea was to share this storage abstraction with curp, so I put it inside utils

xline/src/storage/auth_store/backend.rs
themanforfree: fixed

xline/src/storage/auth_store/backend.rs
themanforfree: fixed

xline/src/storage/revision.rs
themanforfree: fixed

utils/src/storage.rs
themanforfree: fixed

utils/src/storage.rs
themanforfree: fixed

xline/src/storage/auth_store/backend.rs
themanforfree: fixed

xline/src/storage/auth_store/backend.rs
themanforfree: fixed

xline/src/storage/kv_store.rs
themanforfree: fixed

xline/src/storage/db.rs
```diff
@@ -1,69 +1,124 @@
...
 
     /// Mark deletion for keys
-    /// TODO support don't return `prev_kvs`
-    pub(crate) fn mark_deletions(&self, revisions: &[(Revision, Revision)]) -> Vec<KeyValue> {
-        let mut storage = self.storage.lock();
+    pub(crate) fn mark_deletions(
+        &self,
+        revisions: &[(Revision, Revision)],
+    ) -> Result<Vec<KeyValue>, ExecuteError> {
         let prev_kvs: Vec<KeyValue> = revisions
```
markcty, line 73: We grab the storage lock only once before and now twice. Will there be concurrent issues?

xline/src/storage/kvwatcher.rs
```diff
@@ -246,7 +263,9 @@ impl KvWatcherInner {
         let initial_events = if start_rev == 0 {
             vec![]
         } else {
-            self.storage.get_event_from_revision(key_range, start_rev)
+            self.storage
```
markcty, line 266: Shall we at least give a warning here? I guess it's not normal if we fail to get events from storage?

xline/src/lib.rs
```diff
@@ -127,3 +127,5 @@ pub mod server;
 mod state;
 /// Storage module
 mod storage;
+
+pub use storage::storage_abstraction::Memory;
```
markcty, line 131: Shall we make a naming convention about interface? I use `XxxInterface` in Curp(now I find it too tedious). In xline, there is `xxx_abstraction` and `LeaseTrait`. I suggest we name all interfaces to `XxxApi`.

themanforfree: good suggestion, XxxApi is a better choice

xline/src/storage/kvwatcher.rs
themanforfree: fixed

xline/src/storage/db.rs
themanforfree: Yes, moved the lock out of memory, then we can lock it only once like before

markcty (dismissed):

xline/src/server/xline_server.rs
```diff
@@ -73,6 +79,7 @@ impl XlineServer {
         key_pair: Option<(EncodingKey, DecodingKey)>,
         server_timeout: ServerTimeout,
         client_timeout: ClientTimeout,
+        mem_storage: S,
```
Phoenix500526, line 82: I don't think we should expose too many details about the implementation in a variable name. Please modify `mem_storage` as `storage`.

themanforfree: my mistake, fixed

xline/src/storage/mod.rs
```diff
@@ -1,25 +1,24 @@
...
-
 /// Storage for lease
 pub(crate) mod lease_store;
+/// Memory storage module
+pub(crate) mod memory;
+/// Request context module
+pub(crate) mod req_ctx;
+/// Revision module
+pub(crate) mod revision;
+/// persistent storage abstraction
```
Phoenix500526, line 19: Keep the comment style consistent. "Persistent storage abstraction"

Phoenix500526 (approved):
LGTM

markcty (approved):

Issue #160, opened 2023-01-30
[Feature]: Make `ExecuteError` become more generic 


themanforfree:
### Description about the feature

Our `ExecuteError` now has only two members, but in actual situations, other errors may occur in the executor, but in `Curp`, it is impossible to know the errors that will occur in the upper layer, so a solution is needed to allow `Corp` users to define their own errors type.

### Approach
1. The easiest way is to let `ExecuteError` only contain the message but not the type, which is similar to our current implementation, but the upper layer will not be able to judge the type of the error.
2. Another way is let the error as the associated type of the `CommandExecutor` trait, But because many places in `Curp` used `ExecuteError`, here will be complicated changes.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #170, opened 2023-02-06
Refactor/execute error

themanforfree:
#160

### DIFF
--- a/curp/tests/common/mod.rs
+++ b/curp/tests/common/mod.rs
@@ -1,6 +1,11 @@
 #![allow(dead_code, unused)]
 
-use std::{env, time::Duration};
+use std::{
+    io,
+    {env, time::Duration},
+};
+
+use thiserror::Error;
 
 pub mod curp_group;
 pub mod test_cmd;
--- a/xline/src/server/lock_server.rs
+++ b/xline/src/server/lock_server.rs
@@ -35,8 +35,7 @@ use crate::{
 const DEFAULT_SESSION_TTL: i64 = 60;
 
 /// Lock Server
-//#[derive(Debug)]
-#[allow(dead_code)] // Remove this after feature is completed
+#[derive(Debug)]
 pub(crate) struct LockServer<S>
 where
     S: StorageApi,
--- a/xline/src/storage/db.rs
+++ b/xline/src/storage/db.rs
@@ -1,8 +1,7 @@
-use curp::error::ExecuteError;
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 use prost::Message;
 
-use super::{revision::Revision, storage_api::StorageApi};
+use super::{storage_api::StorageApi, ExecuteError, Revision};
 use crate::rpc::KeyValue;
 
 /// Database to store revision to kv mapping
@@ -32,7 +31,7 @@ where
         let key = revision.encode_to_vec();
         let value = kv.encode_to_vec();
         let _prev_val = storage.insert(key, value).map_err(|e| {
-            ExecuteError::InvalidCommand(format!("Failed to insert revision {revision:?}: {e}"))
+            ExecuteError::DbError(format!("Failed to insert key-value, error: {e}"))
         })?;
         Ok(())
     }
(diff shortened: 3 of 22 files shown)


themanforfree:
based on #164

xline/src/server/lock_server.rs
```diff
@@ -36,7 +36,6 @@ const DEFAULT_SESSION_TTL: i64 = 60;
 
 /// Lock Server
 //#[derive(Debug)]
```
Phoenix500526, line 38: Plz remove this line.

xline/src/storage/execute_error.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub(crate) fn lease_already_exists(lease_id: i64) -> Self {
+        Self::LeaseError(format!("lease {} already exists", lease_id))
+    }
+
+    /// Auth is not enabled
+    pub(crate) fn auth_not_enabled() -> Self {
+        Self::AuthError("auth is not enabled".to_owned())
+    }
+
+    /// Auth Failed
```
Phoenix500526, line 61: Auth failed. Plz keep the comment style consistent.

xline/src/storage/db.rs
```diff
@@ -117,13 +90,10 @@ where
...
                 let _prev_val = storage.insert(key, value).map_err(|e| {
-                    ExecuteError::InvalidCommand(format!(
-                        "Failed to insert revision {:?}: {}",
-                        new_rev, e
-                    ))
-                });
-                Ok(del_kv)
-            })
-            .collect()
+                    ExecuteError::db_error(format!("Failed to insert key-value, error: {e}"))
```
Phoenix500526, line 93: It looks that there is nothing different between `ExecuteError::db_error(format!("Failed to insert key-value, error: {e}"))` and `ExecuteError::DbError(format!("Failed to insert key-value, error: {e}"))`. So why do need such a function?

themanforfree: removed

xline/src/storage/execute_error.rs
themanforfree: fixed

xline/src/server/lock_server.rs
themanforfree: fixed

Phoenix500526 (approved):

curp/src/test_utils/execute_error.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[error("meet io related error")]
+    IoError(String),
+}
+
+impl From<io::Error> for ExecuteError {
+    #[inline]
+    fn from(err: io::Error) -> Self {
+        Self::IoError(err.to_string())
+    }
+}
```
markcty, line 23: I guess there is no need to have such a complex execute error for test command. String would be fine.

curp/tests/common/mod.rs
```diff
@@ -22,3 +27,23 @@ pub async fn sleep_millis(n: u64) {
...
 }
+
+/// Error met when executing commands
+#[allow(clippy::module_name_repetitions)] // this-error generate code false-positive
+#[non_exhaustive]
+#[derive(Error, Debug, Clone)]
+pub enum ExecuteError {
+    /// Command is invalid
+    #[error("invalid command {0} ")]
+    InvalidCommand(String),
```
markcty, line 38: same above

xline/src/storage/lease_store/mod.rs
```diff
@@ -277,11 +277,7 @@ impl LeaseStore {
 
     /// Keep alive a lease
     pub(crate) fn keep_alive(&self, lease_id: i64) -> Result<i64, ExecuteError> {
-        if !self.is_leader() {
-            return Err(ExecuteError::InvalidCommand(
-                "lease keep alive must be called on leader".to_owned(),
-            ));
-        }
+        assert!(self.is_leader(), "keep_alive must be called on leader");
```
markcty, line 280: I guess it leadership change can happen before this function is called. If so, assert! will panic. So I think the previous solution is better.

markcty (approved):

Phoenix500526 (approved):

Issue #183, opened 2023-02-21
[Proposal] Persistent Layer Designed Doc


Phoenix500526:
# Persistent Layer 

## Schema
We can define two DB instances, the `curpdb` and the `kvdb`.  The different instances hold different column families.
Column families that the curpdb has are:
* log: to persistent curp log entries
* snapshot: to persistent curp snapshot 
* state: to persistent curp state, like voted_for、term or something like this.

Column families that the kvdb has are:
* kv: to store kv data for `KvServer`
* auth: to store auth data for `AuthServer`
* lease: to store lease data for `LeaseServer`

## API Definitions
```Rust
pub trait StorageApi {
    /// The byte-vector type through which the database returns read values.
    type DbVector: DbVector;
    type Error:  std::error::Error + Send + Sync + 'static;
    /// A consistent read-only snapshot of the database
    type Snapshot: Snapshot;

    /// Read a value for a key from a given column family. Return `Ok(None)` if the key does not exist.
    fn get_value_cf(&self, cf: &str, key: &[u8]) -> Result<Option<Self::DbVector>, Self::Error>;
    
    /// Read values for multi keys from a given column family. 
    fn batch_get_cf(&self, cf: &str, keys: &[impl AsRef<[u8]>]) -> Result<Vec<Option<Self::DbVector>>, Self::Error>;

    /// Insert a key value pair into a given column family. Return Ok(Some(OldValue)) if the key exists, otherwise Ok(None)
    fn insert_cf(
        &mut self,
        cf: &str,
        key: impl Into<Vec<u8>>,
        value: impl Into<Vec<u8>>,
    ) -> Result<Option<Self::DbVector>, Self::Error>;

    /// Remove a key from a given column family. Return Ok(Some(OldValue)) if the key exists, otherwise Ok(None)
    fn remove_cf(&mut self, cf: &str, key: impl AsRef<[u8]>) -> Result<Option<Self::DbVector>, Self::Error>;

    /// Remove a range of key from a given column family.
    fn remove_range_cf(&mut self, cf: &str, begin_key: impl Into<Vec<u8>>, end_key: impl Into<Vec<u8>>) -> Result<Vec<Option<Self::DbVector>>, Self::Error>;

    /// Get the snapshot object.
    fn snapshot(&self) -> Self::Snapshot;
}
```

## Error Definitions 
```Rust
#[derive(Debug, Error)]
pub enum Error {
    #[error("Io {0}")]
    Io(#[from] std::io::Error),
[...]

PR #185, opened 2023-02-28
Persistent Layer Implementation

Phoenix500526:
Given that a large pr will increase the review burden, I am preparing to split this feature into two pull requests. 
The first pr is to refactor the original storage layer to conform to the design described in issue #183. 
The second pr is to add rocksdb engine implementation.

This is the first pull request.

Refactor the original persistent layer:
1. Define the interface between the middle DB layer and the lower storage layer, including the StorageEngine trait, WriteOperation struct, and the EngineError.
2. Implement a MemoryEngine to support tests.
3. Deine the interface between the upper store layer and the middle DB layer, including the StorageApi trait
4. Modify some XXStore logic, and add some error process logic.

No


xline/src/server/xline_server.rs
```diff
@@ -91,25 +95,30 @@ where
...
+            Arc::clone(&storage),
+        )?);
         let kv_storage = Arc::new(KvStore::new(
             lease_cmd_tx.clone(),
             del_rx,
             Arc::clone(&header_gen),
-            storage.clone(), // TODO: implement bucket storage
-        ));
-        let auth_storage = Arc::new(AuthStore::new(lease_cmd_tx, key_pair, header_gen, storage));
+            Arc::clone(&storage), // TODO: implement bucket storage
```
themanforfree, line 104: remove this todo, bucket in etcd is our table

xline/src/storage/db.rs
```diff
@@ -1,141 +1,84 @@
...
 
-    /// Insert a `KeyValue`
-    pub(crate) fn insert(&self, revision: Revision, kv: &KeyValue) -> Result<(), ExecuteError> {
-        let mut storage = self.storage.write();
-        let key = revision.encode_to_vec();
-        let value = kv.encode_to_vec();
-        let _prev_val = storage.insert(key, value).map_err(|e| {
+    fn insert<K, V>(&self, table: &str, key: K, value: V) -> Result<(), ExecuteError>
+    where
+        K: AsRef<[u8]> + std::fmt::Debug,
```
themanforfree, line 42: I this K and V should be `Into<Vec<u8>>` at here, otherwise there will be unnecessary heap allocation

xline/src/storage/kv_store.rs
```diff
@@ -140,31 +142,33 @@ where
...
     pub(crate) fn new(
         kv_update_tx: mpsc::Sender<(i64, Vec<Event>)>,
         lease_cmd_tx: mpsc::Sender<LeaseMessage>,
         header_gen: Arc<HeaderGenerator>,
-        storage: S,
-    ) -> Self {
-        Self {
+        db: Arc<DB>,
+    ) -> Result<Self, ExecuteError> {
+        db.create_table("kv")?;
```
themanforfree, line 161: Our table will hardly change, I think we can use a set of global `&'static str` or an enumeration to represent all tables, and collect them into an array, and create all tables when creating Engine, like this There will be no errors here. When we use RocksDB, opening an existing DB also needs to provide slices containing all Cf names

xline/src/storage/auth_store/backend.rs
```diff
@@ -105,20 +107,22 @@ where
...
     pub(super) fn new(
         lease_cmd_tx: mpsc::Sender<LeaseMessage>,
         key_pair: Option<(EncodingKey, DecodingKey)>,
         header_gen: Arc<HeaderGenerator>,
-        storage: S,
-    ) -> Self {
-        Self {
+        db: Arc<DB>,
+    ) -> Result<Self, ExecuteError> {
+        db.create_table("auth")?;
```
themanforfree, line 121: same as kv

xline/src/storage/lease_store/mod.rs
```diff
@@ -315,25 +318,27 @@ where
...
     pub(crate) fn new(
         del_tx: mpsc::Sender<DeleteMessage>,
         state: Arc<State>,
         header_gen: Arc<HeaderGenerator>,
-        storage: S,
-    ) -> Self {
-        Self {
+        db: Arc<DB>,
+    ) -> Result<Self, ExecuteError> {
+        db.create_table("lease")?;
```
themanforfree, line 332: same as kv

xline/src/storage/db.rs
```diff
@@ -1,141 +1,84 @@
...
+    #[must_use]
+    pub fn new(engine: S) -> Self {
         Self {
-            storage: RwLock::new(storage),
+            engine: Arc::new(engine),
         }
     }
+}
+
+impl<S> StorageApi for DB<S>
```
themanforfree, line 29: The interfaces provided by `StorageApi` and `StorageEngine` are very similar, they both have the ability to read and write `Vec<u8>`, why can't we merge them

engine/src/memory_engine.rs
```diff
@@ -0,0 +1,202 @@
+use std::{cmp::Ordering, collections::HashMap, sync::Arc};
```
rogercloud, line 1: If this engine is test only, please comment in the mod comment (the head of the file).

xline/Cargo.toml
```diff
@@ -23,6 +23,7 @@ event-listener = "2.5.2"
 jsonwebtoken = "8.1.1"
 itertools = "0.10.3"
 utils = { path = "../utils", version = "0.1.0", features = ["parking_lot"] }
+engine = { path = "../engine", version = "0.1.0" }
```
rogercloud, line 26: Actually I don't think we need a version, the path is enough. So does `utils` above.

xline/src/main.rs
```diff
@@ -367,17 +368,17 @@ async fn main() -> Result<()> {
     debug!("name = {:?}", cluster_config.name());
     debug!("server_addr = {:?}", self_addr);
     debug!("cluster_peers = {:?}", cluster_config.members());
-    let mem_storage = Memory::new();
+    let mem_engine = MemoryEngine::new();
```
rogercloud, line 371: Shall we make the engine generic instead of hard coding?

xline/src/server/xline_server.rs
Phoenix500526: Fixed

xline/src/storage/db.rs
Phoenix500526: Fixed

xline/src/storage/db.rs
Phoenix500526: That's because Xline Server is not the only one with persistence requirements. So has the Curp Server. Their `StorageApi` may be different in that `StorageApi` is closer to the business logic. To isolate change, I add a middle trait `StorageEngine`. If we need to replace our storage engine in the future, we just need to implement `StorageEngine` for the new storage engine. But without that, we need to implement `StorageApi` twice, one for the Curp Server while the other for the Xline Server.

xline/Cargo.toml
Phoenix500526: Fixed

engine/src/memory_engine.rs
Phoenix500526: Fixed

xline/src/storage/kv_store.rs
Phoenix500526: Fixed

xline/src/storage/auth_store/backend.rs
Phoenix500526: Fixed

xline/src/storage/lease_store/mod.rs
Phoenix500526: Fixed

xline/src/server/xline_server.rs
```diff
@@ -65,6 +65,10 @@ where
 {
     /// New `XlineServer`
     ///
+    /// # Errors
+    ///
+    /// Return `ExecuteError::DbError` if the server cannot initialize the database
+    ///
```
themanforfree, line 71: remove these comment

markcty (approved):

xline/src/server/xline_server.rs
Phoenix500526: Fixed

xline/src/main.rs
Phoenix500526: I will fixed it in the next pr

markcty (approved):

themanforfree (approved):

PR #187, opened 2023-03-02
Feature/rocksdb engine

Phoenix500526:
This is the second pr based on #185, described in issue #183

- [X] implement RocksEngine and add tests to cover it
- [x] add configuration setting
- [ ] fail point test

### DIFF
--- a/engine/src/engine_api.rs
+++ b/engine/src/engine_api.rs
@@ -5,71 +5,52 @@ use crate::error::EngineError;
 #[derive(Debug)]
 pub enum WriteOperation<'a> {
     /// `Put` operation
-    Put(Put<'a>),
+    Put {
+        /// The table name
+        table: &'a str,
+        /// Key
+        key: Vec<u8>,
+        /// Value
+        value: Vec<u8>,
+    },
     /// `Delete` operation
-    Delete(Delete<'a>),
-    /// `DeleteRange` operation
-    DeleteRange(DeleteRange<'a>),
+    Delete {
+        /// The table name
+        table: &'a str,
+        /// The target key
+        key: &'a [u8],
+    },
+    /// Delete range operation, it will remove the database entries in the range [from, to)
+    DeleteRange {
+        /// The table name
+        table: &'a str,
+        /// The `from` key
+        from: &'a [u8],
+        /// The `to` key
+        to: &'a [u8],
+    },
 }
 
-/// Put operation
-#[derive(Debug)]
-pub struct Put<'a> {
-    /// The table name
-    pub(crate) table: &'a str,
-    /// Key
-    pub(crate) key: Vec<u8>,
-    /// Value
-    pub(crate) value: Vec<u8>,
-}
-
-impl<'a> Put<'a> {
+impl<'a> WriteOperation<'a> {
     /// Create a new `Put` operation
     #[inline]
     #[must_use]
-    pub fn new(table: &'a str, key: Vec<u8>, value: Vec<u8>) -> Put<'a> {
-        Put { table, key, value }
+    pub fn new_put(table: &'a str, key: Vec<u8>, value: Vec<u8>) -> Self {
+        Self::Put { table, key, value }
     }
-}
 
-/// Delete operation,
-#[allow(dead_code)]
-#[derive(Debug)]
-pub struct Delete<'a> {
-    /// The table name
-    pub(crate) table: &'a str,
-    /// The target key
-    pub(crate) key: &'a [u8],
-}
-
-impl<'a> Delete<'a> {
     /// Create a new `Delete` operation
     #[inline]
     #[must_use]
-    pub fn new(table: &'a str, key: &'a [u8]) -> Delete<'a> {
-        Delete { table, key }
+    pub fn new_delete(table: &'a str, key: &'a [u8]) -> Self {
+        Self::Delete { table, key }
     }
-}
 
-/// Delete range operation, it will remove the database
-/// entries in the range [from, to)
-#[allow(dead_code)]
-#[derive(Debug)]
-pub struct DeleteRange<'a> {
-    /// The table name
-    pub(crate) table: &'a str,
-    /// The `from` key
-    pub(crate) from: &'a [u8],
-    /// The `to` key
-    pub(crate) to: &'a [u8],
-}
-
-impl<'a> DeleteRange<'a> {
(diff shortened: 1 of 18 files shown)


engine/src/engine_api.rs
```diff
@@ -13,7 +13,6 @@ pub enum WriteOperation<'a> {
 }
```
markcty, line 13: Maybe it's more conventional to use nested struct, e.g. 
```suggestion
pub enum WriteOperation<'a> {
    /// `Put` operation
    Put {
        /// The table name
        pub(crate) table: &'a str,
        /// Key
        pub(crate) key: Vec<u8>,
        /// Value
        pub(crate) value: Vec<u8>,
    },
    // ...
```

And put the construction functions to `WriteOperation`, eg. `WriteOperation::new_put(...)` `WriteOperation::new_delete()`.

engine/src/error.rs
```diff
@@ -9,7 +7,7 @@ use thiserror::Error;
 pub enum EngineError {
     /// Met I/O Error during persisting data
     #[error("I/O Error")]
-    IoError(#[from] io::Error),
+    IoError(String),
```
markcty, line 10: Why changed to `String`? And `I/O Error: {0}`.

engine/src/rocksdb_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+                        .ok_or(EngineError::TableNotFound(table.to_owned()))?;
+                    batch.delete_range_cf(&cf, from, to);
+                }
+            }
+        }
+        let mut opt = WriteOptions::default();
+        opt.set_sync(sync);
+        self.inner
+            .write_opt(batch, &opt)
+            .map_err(|e| EngineError::UnderlyingError(format!("{e}")))
```
markcty, line 105: What's the difference between `UnderlyingError` and `IoError`? I notice we will return `IoError` in the above code.

markcty (changes requested):

engine/src/rocksdb_engine.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+impl RocksEngine {
+    /// New `RocksEngine`
+    ///
+    /// # Errors
+    ///
+    /// Return `EngineError` when DB open failed.
+    #[inline]
+    pub fn new(data_dir: &PathBuf, tables: &[&'static str]) -> Result<Self, EngineError> {
```
themanforfree, line 24: I think use `impl AsRef<Path>` is batter than `&PathBuf`, and maybe use `open` as this method's name is better, because we don't create new `DB` every time

xline/src/main.rs
```diff
@@ -205,6 +201,12 @@ struct ServerArgs {
     /// Curp client retry timeout
     #[clap(long, value_parser = parse_duration, default_value = "50ms")]
     client_retry_timeout: Duration,
+    /// Storage engine
+    #[clap(long)]
+    storage_engine: String,
+    /// DB directory
+    #[clap(long, default_value = "/usr/local/xline/data-dir")]
```
themanforfree, line 208: Our config file and arguments parsing are a bit confusing now. Their default values are not common. For example, there is no default value for data_dir in the configuration file but it is in the arguments parsing, and some default values ​​use default_xxx in the configuration file, but in the arguments parsing directly uses a string, they may be different

utils/src/config.rs
```diff
@@ -276,6 +279,29 @@ impl Default for ClientTimeout {
     }
 }
 
+/// Storage Configuration
+#[allow(clippy::module_name_repetitions)]
+#[non_exhaustive]
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(
+    tag = "engine",
+    content = "datadir",
```
themanforfree, line 288: use `data_dir`, same as arguments parsing

utils/src/config.rs
```diff
@@ -579,6 +612,10 @@ mod tests {
                 node2 = '127.0.0.1:2380'
                 node3 = '127.0.0.1:2381'
 
+                [storage]
+                engine = 'rocksdb'
+                datadir = '/tmp/xline/data-dir'
```
themanforfree, line 617: same above

engine/src/engine_api.rs
Phoenix500526: Fixed

utils/src/config.rs
Phoenix500526: Fixed

utils/src/config.rs
Phoenix500526: Fixed

engine/src/rocksdb_engine.rs
Phoenix500526: Fixed

engine/src/error.rs
Phoenix500526: That's because the error defined in rocksdb is based on `String`. It uses different error message prefixes to tell different errors apart.  Therefore, an `IOError` that throw by rocksdb is only an error with the `Io Error: ...` message.

engine/src/rocksdb_engine.rs
Phoenix500526: Rocksdb provides 16 different errors. The `UnderlyingError` is a storage engine-specific error wrapper.

xline/src/main.rs
Phoenix500526: Fixed

markcty:
The first commit `implement DB struct and refactor some store backend logic` contains nothing? Plz remove it.

engine/src/rocksdb_engine.rs
```diff
@@ -0,0 +1,201 @@
...
+
+use rocksdb::{Error as RocksError, Options, WriteBatchWithTransaction, WriteOptions, DB};
+
+use crate::{
+    engine_api::{StorageEngine, WriteOperation},
+    error::EngineError,
+};
+
+/// Translate a `RocksError` into a `EngineError`
+fn translate_error(err: RocksError) -> EngineError {
```
themanforfree, line 11: use `From` trait, so that we don't have to `map_err` every time, just use the `?` directly

Phoenix500526: Fixed

markcty (dismissed):

markcty (approved):

themanforfree (approved):

PR #191, opened 2023-03-08
Refactor/auth store

Phoenix500526:
refactor AuthStoreBackend implementation, including:
1. use a simple kv store model to replace the mvcc in that we don't need to store the event history of auth data.
2. extract some logic, especially the execute, after sync and permission check, from AuthStoreBackend;
3. add some test to improve test coverage.

No

### DIFF
--- a/xline/src/storage/auth_store/store.rs
+++ b/xline/src/storage/auth_store/store.rs
@@ -1,19 +1,51 @@
-use std::{collections::VecDeque, sync::Arc};
+use std::{
+    cmp::Ordering,
+    collections::{HashMap, VecDeque},
+    sync::{
+        atomic::{AtomicBool, Ordering as AtomicOrdering},
+        Arc,
+    },
+};
 
 use anyhow::Result;
+use clippy_utilities::Cast;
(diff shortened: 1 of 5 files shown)


xline/src/storage/auth_store/store.rs
```diff
@@ -39,40 +84,791 @@ where
...
-            inner: Arc::new(AuthStoreBackend::new(
-                lease_cmd_tx,
-                key_pair,
-                header_gen,
-                storage,
-            )),
+    ) -> Result<Self, ExecuteError> {
+        let backend = Arc::new(AuthStoreBackend::new(storage));
+        let revision = backend.get_revision()?;
+        let enabled = backend.get_enable()?;
```
themanforfree, line 90: just use default values, I will implement recover logic in another method

Phoenix500526: Fixed

themanforfree (approved):

markcty (approved):

themanforfree (approved):

PR #196, opened 2023-03-12
add cache action to speed up github action workflow

Phoenix500526:
Use GitHub cache action to speed up our workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
           profile: default
           toolchain: ${{ env.CI_RUST_TOOLCHAIN }}
           override: true
+      - uses: Swatinem/rust-cache@v2
       - uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -39,6 +39,21 @@ jobs:
...
+      - name: Cache cargo index
+        uses: actions/cache@v3
+        with:
+          path: ~/ci/check/.cargo/git
+          key: ${{ runner.os }}-ci-check-index-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo build
+        uses: actions/cache@v3
+        with:
+          path: ~/ci/check/target
+          key: ${{ runner.os }}-ci-check-build-target-${{ hashFiles('**/Cargo.lock') }}
```
markcty, line 56: Why not use https://github.com/Swatinem/rust-cache this action? It's more simple to use.

markcty (changes requested):

.github/workflows/ci.yml
Phoenix500526: Fixed

rogercloud (approved):

markcty (approved):

PR #199, opened 2023-03-14
Refactor/batch write ops

themanforfree:

This pr only handles the kv store，I will add other store's refactor in other pr

Atomize persistence operations
batch all write ops when sync command
no

### DIFF
--- a/curp/src/cmd.rs
+++ b/curp/src/cmd.rs
@@ -118,5 +118,5 @@ where
     async fn reset(&self);
 
     /// Index of the last log entry that has been successfully applied to the command executor
-    fn last_applied(&self) -> usize;
+    fn last_applied(&self) -> Result<LogIndex, Self::Error>;
 }
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -52,8 +52,8 @@ pub(super) enum CurpError {
     #[error("storage error, {0}")]
     Storage(#[from] StorageError),
     /// Get applied index error
-    #[error("get applied index error, {0}")]
-    GetAppliedIndex(String),
+    #[error("internal error {0}")]
+    Internal(String),
 }
 
 /// Connects
@@ -287,7 +287,7 @@ impl<C: 'static + Command> CurpNode<C> {
         let uncommitted_pool = Arc::new(Mutex::new(UncommittedPool::new()));
         let last_applied = cmd_executor
             .last_applied()
-            .map_err(|e| CurpError::GetAppliedIndex(e.to_string()))?;
+            .map_err(|e| CurpError::Internal(format!("get applied index error, {e}")))?;
 
         let storage = Arc::new(RocksDBStorage::new(&curp_cfg.data_dir)?);
 
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -51,6 +51,9 @@ pub(super) enum CurpError {
     /// Storage error
     #[error("storage error, {0}")]
     Storage(#[from] StorageError),
+    /// Get applied index error
+    #[error("get applied index error, {0}")]
+    GetAppliedIndex(String),
 }
 
 /// Connects
@@ -282,7 +285,9 @@ impl<C: 'static + Command> CurpNode<C> {
         let cmd_board = Arc::new(RwLock::new(CommandBoard::new()));
         let spec_pool = Arc::new(Mutex::new(SpeculativePool::new()));
         let uncommitted_pool = Arc::new(Mutex::new(UncommittedPool::new()));
-        let last_applied = cmd_executor.last_applied();
+        let last_applied = cmd_executor
+            .last_applied()
+            .map_err(|e| CurpError::GetAppliedIndex(e.to_string()))?;
 
         let storage = Arc::new(RocksDBStorage::new(&curp_cfg.data_dir)?);
 
@@ -332,7 +337,7 @@ impl<C: 'static + Command> CurpNode<C> {
                 log_tx,
                 voted_for,
                 entries,
-                last_applied,
+                last_applied.numeric_cast(),
             ))
         };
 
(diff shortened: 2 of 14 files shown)


engine/src/engine_api.rs
```diff
@@ -16,41 +16,62 @@ pub enum WriteOperation<'a> {
     /// `Delete` operation
     Delete {
         /// The table name
-        table: &'a str,
+        table: &'static str,
         /// The target key
-        key: &'a [u8],
+        key: Vec<u8>,
```
Phoenix500526, line 21: How come you substitute `Vec<u8>` for `&'a [u8]`?  As many data structures in std, say `HashMap`, `HashSet`, or  `Vec`, using a reference to remove a key is enough. There is no need to clone the value here.

engine/src/engine_api.rs
```diff
@@ -16,41 +16,62 @@ pub enum WriteOperation<'a> {
...
     DeleteRange {
         /// The table name
-        table: &'a str,
+        table: &'static str,
         /// The `from` key
-        from: &'a [u8],
+        from: Vec<u8>,
         /// The `to` key
-        to: &'a [u8],
+        to: Vec<u8>,
```
Phoenix500526, line 30: Same as above

xline/src/server/command.rs
```diff
@@ -216,24 +222,36 @@ where
...
-        )?;
+        )];
         match wrapper.request.backend() {
-            RequestBackend::Kv => self.kv_storage.after_sync(wrapper).await,
-            RequestBackend::Auth => self.auth_storage.after_sync(wrapper),
-            RequestBackend::Lease => self.lease_storage.after_sync(wrapper).await,
+            RequestBackend::Kv => {
+                let (res, mut ops) = self.kv_storage.after_sync(wrapper).await?;
+                wr_ops.append(&mut ops);
+                self.backend.write_batch(wr_ops, false)?;
```
Phoenix500526, line 234: If you want to insert a key-value pair into the backend, using the `insert` method is ok because the `insert` method will do the same thing. It is implemented by `write_batch`.

themanforfree: There may be more than one write operation here, we need to batch them

engine/src/engine_api.rs
themanforfree: we need collect `WriteOperation` from different places，and batch write them later，If use references, there will be many lifetime related problems

engine/src/engine_api.rs
themanforfree: Same above

curp/src/server/cmd_board.rs
```diff
@@ -28,7 +27,7 @@ pub(super) struct CommandBoard<C: Command> {
     /// Whether the cmd has been speculatively executed
     pub(super) spec_executed: HashSet<ProposeId>,
     /// Whether the cmd needs after sync
-    pub(super) needs_as: HashMap<ProposeId, LogIndex>,
+    pub(super) needs_as: HashMap<ProposeId, usize>,
```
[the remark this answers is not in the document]
themanforfree: In rust, we can't use `u64` directly as an index, and except for persistence, this index does not need to consider size, so I use `usize` directly, and cast this index to `u64` when persisting

curp/src/cmd.rs
```diff
@@ -45,7 +43,7 @@ pub trait Command:
 
     /// Execute the command after_sync callback
     #[inline]
-    async fn after_sync<E>(&self, e: &E, index: LogIndex) -> Result<Self::ASR, E::Error>
+    async fn after_sync<E>(&self, e: &E, index: usize) -> Result<Self::ASR, E::Error>
```
rogercloud, line 46: Why do we use `usize` instead of `LogIndex`?

curp/src/server/cmd_board.rs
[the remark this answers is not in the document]
markcty: Maybe it's ok to use usize across 32bit and 64bit systems as long as we use u64 as the persistency format for log index. And I think usize is too simple to require a type alias. If we use usize here and `LogIndex` there, it will also be confusing.

curp/src/server/curp_node.rs
```diff
@@ -51,6 +51,9 @@ pub(super) enum CurpError {
     /// Storage error
     #[error("storage error, {0}")]
     Storage(#[from] StorageError),
+    /// Get applied index error
+    #[error("get applied index error, {0}")]
+    GetAppliedIndex(String),
```
markcty, line 56: How about `Internal(String)` and move the `get applied index error` below? It's too specific.

themanforfree: Sure

curp/src/cmd.rs
themanforfree: discussed above

xline/src/storage/kv_store.rs
markcty, line 1: Line 97 should be deleted

xline/src/server/xline_server.rs
```diff
@@ -51,6 +51,8 @@ where
     auth_storage: Arc<AuthStore<S>>,
     /// Lease storage
     lease_storage: Arc<LeaseStore<S>>,
+    /// persistent storage
+    backend: Arc<S>,
```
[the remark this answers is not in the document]
themanforfree: Renamed

xline/src/storage/kv_store.rs
themanforfree: Deleted

engine/src/engine_api.rs
Phoenix500526: I have pulled this pr and tried to use `&'a [u8]` to replace `Vec<u8>`.  Nothing went wrong. So I think we can still use `&'a [u8]` as before.

Phoenix500526 (approved):

markcty (approved):

PR #200, opened 2023-03-14
Refactor: kv&lease store

themanforfree:

based on #199 

### DIFF
--- a/curp/src/cmd.rs
+++ b/curp/src/cmd.rs
@@ -101,7 +101,7 @@ impl ConflictCheck for u32 {
 /// Command executor which actually executes the command.
 /// It is usually defined by the protocol user.
 #[async_trait]
-pub trait CommandExecutor<C>: Sync + Send + Clone + std::fmt::Debug
+pub trait CommandExecutor<C>: Sync + Send + std::fmt::Debug
 where
     C: Command,
 {
--- a/xline/src/storage/auth_store/store.rs
+++ b/xline/src/storage/auth_store/store.rs
@@ -7,8 +7,8 @@ use std::{
     },
 };
 
-use anyhow::Result;
 use clippy_utilities::Cast;
+use engine::WriteOperation;
 use itertools::Itertools;
 use jsonwebtoken::{DecodingKey, EncodingKey};
 use log::debug;
@@ -17,12 +17,14 @@ use pbkdf2::{
     password_hash::{PasswordHash, PasswordVerifier},
     Pbkdf2,
 };
+use prost::Message;
 use tokio::sync::mpsc;
 use utils::parking_lot_lock::RwLockMap;
 
 use super::{
-    backend::{ROOT_ROLE, ROOT_USER},
+    backend::{AUTH_ENABLE_KEY, AUTH_REVISION_KEY, AUTH_TABLE, ROOT_ROLE, ROOT_USER, USER_TABLE},
     perms::{JwtTokenManager, PermissionCache, TokenClaims, TokenOperate, UserPermissions},
+    ROLE_TABLE,
 };
 use crate::{
     header_gen::HeaderGenerator,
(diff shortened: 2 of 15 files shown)


xline/src/storage/storage_api.rs
```diff
@@ -4,22 +4,6 @@ use super::ExecuteError;
 
 /// The Stable Storage Api
 pub trait StorageApi: Send + Sync + 'static + std::fmt::Debug {
-    /// Insert key-value pair into the underlying storage engine
-    ///
-    /// # Errors
-    ///
-    /// if error occurs in storage, return `Err(error)`
-    fn insert<K, V>(
```
Phoenix500526, line 12: Why do you remove this method?

xline/src/storage/storage_api.rs
```diff
@@ -50,15 +34,20 @@ pub trait StorageApi: Send + Sync + 'static + std::fmt::Debug {
...
-    /// Delete key-value pair from storage
+    /// Commit a batch of write operations
+    /// If sync is true, the write will be flushed from the operating system
+    /// buffer cache before the write is considered complete. If this
+    /// flag is true, writes will be slower.
     ///
     /// # Errors
     ///
     /// if error occurs in storage, return `Err(error)`
-    fn delete<K>(&self, table: &'static str, key: K, sync: bool) -> Result<(), ExecuteError>
```
Phoenix500526, line 58: Same as above

xline/src/storage/auth_store/store.rs
```diff
@@ -99,10 +101,39 @@ where
         }
     }
 
-    /// generate a next revision and save it into backend
-    fn commit_revision(&self) -> Result<(), ExecuteError> {
-        let _ignore = self.revision.next();
-        self.backend.save_revision(&self.revision)
+    /// generate a next revision and create a write operation to persistent it
+    fn op_inc_and_put_revision(&self) -> WriteOperation {
```
Phoenix500526, line 105: I think we shouldn't expose some lower details, like `WriteOperation`,  to the upper layer. This goes against the law of Demeter. If we need to modify the storage engine layer for some reason in the future, this change will diffuse to the upper layer, like the Store layer and its upper layer.

xline/src/storage/db.rs
```diff
@@ -104,22 +95,35 @@ where
...
-        Ok(())
+            .write_batch(ops, true)
+            .map_err(|e| ExecuteError::DbError(format!("Failed to reset database, error: {e}")))
     }
 
-    fn write_batch(&self, wr_ops: Vec<WriteOperation>, sync: bool) -> Result<(), ExecuteError> {
-        self.engine
-            .write_batch(wr_ops, sync)
-            .map_err(|e| ExecuteError::DbError(format!("Failed to write batch, error: {e}")))
+    fn put_op(&self, propose_id: &ProposeId, op: WriteOp) {
```
Phoenix500526, line 110: `put_op` is a little bit confusing. Actually, we can draw a conclusion from context that `put_op` means to store a `WriteOp` into the buffer. But `put_op` looks like to do a put operation. How about `buffer_op`? We can tell the buffer operations and the storage operations apart. And `buffer_op` implicitly indicates that it only buffers some operations, and it should be flushed.

curp/src/cmd.rs
```diff
@@ -101,7 +101,7 @@ impl ConflictCheck for u32 {
 /// Command executor which actually executes the command.
 /// It is usually defined by the protocol user.
 #[async_trait]
-pub trait CommandExecutor<C>: Sync + Send + Clone + std::fmt::Debug
```
markcty, line 104: What's the purpose of `Clone`. I thought ce is meant to be shared by `Arc` not cloning.

xline/src/storage/db.rs
```diff
@@ -104,22 +95,35 @@ where
...
+        let mut buffer = self.buffer.lock();
+        if let Some(ops) = buffer.get_mut(propose_id) {
+            ops.push(op);
+        } else {
+            let _ignore = buffer.insert(propose_id.clone(), vec![op]);
+        }
+    }
```
markcty, line 117:
```suggestion
        buffer.or_insert_with(|| vec![]).push(op);
```

themanforfree: it should be `buffer.entry(propose_id.clone()).or_insert_with(|| vec![]).push(op)`, if change to this, `propose_id` will be cloned every time

curp/src/cmd.rs
themanforfree: So I removed `Clone`

xline/src/storage/db.rs
themanforfree: Done

xline/src/storage/auth_store/store.rs
themanforfree: Fixed

xline/src/storage/storage_api.rs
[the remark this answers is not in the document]
themanforfree: we didn't need it

xline/src/storage/storage_api.rs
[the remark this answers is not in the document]
themanforfree: Same above

curp/src/cmd.rs
markcty: Sry.

markcty (approved):

Phoenix500526 (approved):

PR #204, opened 2023-03-15
add some necessary config to boot up an xline cluster in benchmark

Phoenix500526:
We set 100ms to simulate the latency on the WAN, but our retry timeout and rpc timeout is only 50ms. Therefore,we need to extent the rpc timeout and retry timeout

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -2,6 +2,8 @@ name: Benchmark
 
 on:
   workflow_dispatch: {}
+  schedule:
+    - cron: "00 00 * * 1"
 
 env:
   CI_RUST_TOOLCHAIN: 1.67.1
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -52,7 +52,7 @@ run_xline() {
     --server-wait-synced-timeout 10s \
     --client-wait-synced-timeout 10s \
     --client_propose_timeout 5s \
-    --log-level debug"
+    --cmd-workers 16"
 
     if [ ${1} -eq 1 ]; then
         cmd="${cmd} --is-leader"
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -58,8 +58,8 @@ run_xline() {
         cmd="${cmd} --is-leader"
     fi
 
-    docker exec -e RUST_LOG=debug -d node${1} ${cmd}
-    echo "docker exec -e RUST_LOG=debug -d node${1} ${cmd}"
+    docker exec -e RUST_LOG=curp,xline. -d node${1} ${cmd}
+    echo "docker exec -e RUST_LOG=curp,xline -d node${1} ${cmd}"
 }
 
 # run etcd node by index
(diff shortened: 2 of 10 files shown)


.github/workflows/benchmark.yml
```diff
@@ -1,7 +1,8 @@
 name: Benchmark
 
 on:
-  workflow_dispatch: {}
```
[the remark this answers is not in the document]
Phoenix500526: Fixed

scripts/benchmark.sh
```diff
@@ -44,14 +44,22 @@ benchmark_cmd() {
...
+    --client-wait-synced-timeout 10s \
+    --client-propose-timeout 5s \
+    --cmd-workers 16"
 
     if [ ${1} -eq 1 ]; then
         cmd="${cmd} --is-leader"
     fi
 
     docker exec -e RUST_LOG=debug -d node${1} ${cmd}
-    echo "command is: docker exec -e RUST_LOG=debug -d node${1} ${cmd}"
```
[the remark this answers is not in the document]
Phoenix500526: Fixed

Phoenix500526:
> Shall we use [educe](https://crates.io/crates/educe) crate to generate default values.

I think there is no need to use `educe` to generate default values, at least not in this pr. 
There are two reasons why we have to write these methods:
1. We need to provide a fine-grained way to configure default values when serializing and deserializing. The `serde` attribute macro provides a specific field attribute, named `[default](https://serde.rs/container-attrs.html#default--path)`, to set the default value of a single field. 
2. Be it `wait_synced_timeout` or `retry_timeout`, all these timeout objects are essential `Duration`. So what we need is to provide different contextual default values for objects of the same type, which cannot be done with `educe`.

rogercloud (dismissed):

scripts/benchmark.sh
```diff
@@ -238,7 +262,7 @@ for server in "xline" "etcd"; do
                 printf ${FORMAT} ${QPS} ${Latency} ${Time} "${cmd}" >>${result_file}
 
             done
+            stop_all
```
themanforfree, line 265: restart containers takes a long time, I recommend cleaning up the cluster inside the container

scripts/benchmark.sh
```diff
@@ -44,14 +44,22 @@ benchmark_cmd() {
 run_xline() {
     cmd="/usr/local/bin/xline \
     --name node${1} \
-    --members ${MEMBERS}"
+    --members ${MEMBERS} \
+    --storage-engine memory \
```
themanforfree, line 48: we should run benchmark with persistent storage

scripts/benchmark.sh
Phoenix500526: Fixed

scripts/benchmark.sh
Phoenix500526: Fixed

themanforfree (approved):

themanforfree (approved):

markcty (approved):

Issue #206, opened 2023-03-18
[Bug]: gc panic during a benchmark 


Phoenix500526:
### Description about the bug

I've captured a panic message when doing a benchmark. The panic message is shown below:
```bash
thread 'tokio-runtime-worker' panicked at 'assertion failed: at <= self.entries.len()', /Users/gy/.cargo/registry/src/[github.com](http://github.com/)-1ecc6299db9ec823/indexmap-1.9.2/src/map/[core.rs:184](http://core.rs:184/):9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Version

0.1.0 (Default)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

markcty:
close on #210

markcty:
close on #210

PR #210, opened 2023-03-20
Fix [Bug]: gc panic during a benchmark  #206

markcty:

Panic in gc

The cause of this bug: spec pool is also modified in cmd worker when a cmd finished after sync. So it's possible that the `last_checked_len` > `spec.len()`. So, instead of recording the length, we now record all keys in spec pool. Admittedly, the cloning of all keys in spec pool will take more time, but I think it's not a big issue since spec pool is usually small.

### DIFF
--- a/curp/src/server/gc.rs
+++ b/curp/src/server/gc.rs
@@ -1,7 +1,12 @@
-use std::time::Duration;
+use std::{collections::HashSet, time::Duration};
+
+use utils::parking_lot_lock::MutexMap;
 
 use super::spec_pool::SpecPoolRef;
-use crate::{cmd::Command, server::cmd_board::CmdBoardRef};
+use crate::{
+    cmd::{Command, ProposeId},
+    server::cmd_board::CmdBoardRef,
+};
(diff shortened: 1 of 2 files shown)


curp/src/server/gc.rs
```diff
@@ -148,4 +159,32 @@ mod tests {
...
+    async fn spec_gc_will_not_panic() {
+        let spec: SpecPoolRef<TestCommand> = Arc::new(Mutex::new(SpeculativePool::new()));
+
+        let cmd1 = Arc::new(TestCommand::default());
+        spec.lock()
+            .pool
+            .insert(cmd1.id().clone(), Arc::clone(&cmd1));
+
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        let cmd2 = Arc::new(TestCommand::default());
```
Phoenix500526, line 174: How come you sleep 100ms here?

themanforfree (approved):

Phoenix500526 (approved):

PR #211, opened 2023-03-20
Use LogIndex for log index

markcty:

Use `LogIndex`(u64) for log index. We previously use `usize` for log index. `usize` may have different size on different machines, which may result in unexpected bugs.

### DIFF
--- a/curp/src/server/raw_curp/log.rs
+++ b/curp/src/server/raw_curp/log.rs
@@ -2,12 +2,14 @@
 
 use std::{collections::HashSet, fmt::Debug, sync::Arc};
 
+use clippy_utilities::NumericCast;
 use tokio::sync::mpsc;
 use tracing::error;
 
 use crate::{
     cmd::{Command, ProposeId},
     log_entry::LogEntry,
+    LogIndex,
 };
 
 /// Curp logs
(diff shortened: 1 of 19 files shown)


curp/src/server/raw_curp/log.rs
```diff
@@ -52,17 +54,17 @@ impl<C: 'static + Command> Log<C> {
             entries, // a fake log[0]
             commit_index: 0,
             last_applied: 0,
-            base_index: 1,
+            base_index: 0,
```
Phoenix500526, line 57: Why don't we need fake log[0] entry now? Is there any use for it other than simplify the calculation?

markcty: No, we still have a fake log[0]. I changed the meaning of base_index(see the comments). I think it's more straight forward this way.

Phoenix500526 (approved):

curp/src/server/raw_curp/log.rs
```diff
@@ -165,18 +166,19 @@ impl<C: 'static + Command> Log<C> {
...
             let entry = self.get(i - 1).unwrap_or_else(|| {
                 unreachable!(
-                    "system corrupted, get log[{i}] when we only have {} log entries",
+                    "system corrupted, get log[{}] when we only have {} log entries",
+                    i - 1,
                     self.last_log_index()
                 )
             });
-            (entry.term, entry.index)
+            (entry.index, entry.term)
```
themanforfree, line 181: Why reverse the order here, I see term in front of index in other places

curp/src/server/raw_curp/mod.rs
```diff
@@ -317,11 +317,11 @@ impl<C: 'static + Command> RawCurp<C> {
...
         term: u64,
         leader_id: String,
-        prev_log_index: usize,
+        prev_log_index: LogIndex,
         prev_log_term: u64,
         entries: Vec<LogEntry<C>>,
-        leader_commit: usize,
-    ) -> Result<u64, (u64, usize)> {
+        leader_commit: LogIndex,
+    ) -> Result<u64, (u64, LogIndex)> {
```
themanforfree, line 324: term in front of index, or reverse? please unify the format

curp/src/server/raw_curp/log.rs
markcty: To unify with the format in other places

themanforfree (approved):

PR #212, opened 2023-03-21
fix bug in lock test

themanforfree:

fix bug in lock test
generate unique lease id instead use timestamp
no

### DIFF
--- a/xline/src/id_gen.rs
+++ b/xline/src/id_gen.rs
@@ -34,9 +34,10 @@ impl IdGenerator {
     }
 
     /// Generate next id
-    pub(crate) fn next(&self) -> u64 {
-        let suffix = self.suffix.fetch_add(1, Ordering::Release);
-        self.prefix | suffix
+    pub(crate) fn next(&self) -> i64 {
+        let suffix = self.suffix.fetch_add(1, Ordering::Relaxed);
+        let id = self.prefix | suffix;
+        (id & 0x7fff_ffff_ffff_ffff).cast()
     }
 }
 
--- a/xline/src/server/lock_server.rs
+++ b/xline/src/server/lock_server.rs
@@ -1,14 +1,9 @@
-#![allow(unused)]
-use std::{
-    sync::Arc,
-    time::{SystemTime, UNIX_EPOCH},
-};
+use std::sync::Arc;
 
-use clippy_utilities::{Cast, OverflowArithmetic};
+use clippy_utilities::OverflowArithmetic;
 use curp::{client::Client, cmd::ProposeId, error::ProposeError};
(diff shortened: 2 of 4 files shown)


xline/src/id_gen.rs
```diff
@@ -38,6 +38,11 @@ impl IdGenerator {
         let suffix = self.suffix.fetch_add(1, Ordering::Release);
```
Phoenix500526, line 38: `Relaxed` is enough.

xline/src/id_gen.rs
```diff
@@ -38,6 +38,11 @@ impl IdGenerator {
...
         self.prefix | suffix
     }
+
+    /// Generate next id and cast to i64
+    pub(crate) fn next_as_i64(&self) -> i64 {
```
Phoenix500526, line 43: Why don't you inline the `next` method here and rename the `next_as_i64` as `next`? Doing so can increase coverage.

Phoenix500526 (changes requested):

xline/src/id_gen.rs
themanforfree: Fixed

xline/src/id_gen.rs
themanforfree: Fixed

Phoenix500526 (approved):

xline/src/id_gen.rs
```diff
@@ -34,9 +34,10 @@ impl IdGenerator {
     }
 
     /// Generate next id
-    pub(crate) fn next(&self) -> u64 {
-        let suffix = self.suffix.fetch_add(1, Ordering::Release);
-        self.prefix | suffix
+    pub(crate) fn next(&self) -> i64 {
```
rogercloud, line 37: A test to cover this change?

xline/src/server/lock_server.rs
```diff
@@ -252,11 +236,7 @@ where
 
     /// Lease grant
     async fn lease_grant(&self, token: Option<String>) -> Result<i64, tonic::Status> {
-        let lease_id = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap_or_else(|e| panic!("SystemTime before UNIX EPOCH! {e}"))
-            .as_secs()
-            .cast(); // TODO: generate lease unique id
+        let lease_id = self.id_gen.next();
```
rogercloud, line 239: Have we double checked other id generations?

xline/src/id_gen.rs
themanforfree: already cover in the back

markcty (approved):

Issue #173, opened 2023-02-09
make gc interval configurable


Phoenix500526:
Currently, the `GC_INTERVAL` is still hard-coded in the `curp/src/server/gc.rs`, like:
```rust
/// How often we should do gc
const GC_INTERVAL: Duration = Duration::from_secs(20);

/// Run background GC tasks for Curp server
pub(super) fn run_gc_tasks<C: Command + 'static>(cmd_board: CmdBoardRef<C>, spec: SpecPoolRef<C>) {
    let _spec_pool_gc = tokio::spawn(gc_spec_pool(spec, GC_INTERVAL));
    let _cmd_board_gc = tokio::spawn(gc_cmd_board(cmd_board, GC_INTERVAL));
}
```
In fact, since the gc interval may be changed in some situations, we should put it into the `ServerTimeout` in the config.rs file and make it configurable, as other configuration field does.

Phoenix500526:
This issue is currently assigned to @myrfy001.

Phoenix500526:
Welcome, this issue is currently assigned to you. Thanks for your contribution.

Phoenix500526:
> Hi, i'm interested in this issue. May I take this good first issue? thx.

Welcome, this issue is currently assigned to you. Thanks for your contribution.

PR #213, opened 2023-03-21
read gc interval from config file

myrfy001:
  * Fixes: #173

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -397,7 +397,7 @@ impl<C: 'static + Command> CurpNode<C> {
                 Arc::clone(&cmd_board),
                 Arc::clone(&spec_pool),
                 uncommitted_pool,
-                curp_cfg,
+                Arc::clone(&curp_cfg),
                 Box::new(ce_event_tx.clone()),
                 Arc::clone(&sync_event),
                 log_tx,
@@ -416,7 +416,7 @@ impl<C: 'static + Command> CurpNode<C> {
                 Arc::clone(&cmd_board),
                 Arc::clone(&spec_pool),
                 uncommitted_pool,
-                curp_cfg,
+                Arc::clone(&curp_cfg),
                 Box::new(ce_event_tx.clone()),
                 Arc::clone(&sync_event),
                 log_tx,
@@ -433,7 +433,11 @@ impl<C: 'static + Command> CurpNode<C> {
             done_tx,
             Arc::clone(&shutdown_trigger),
         );
-        run_gc_tasks(Arc::clone(&cmd_board), Arc::clone(&spec_pool));
+        run_gc_tasks(
+            Arc::clone(&cmd_board),
+            Arc::clone(&spec_pool),
+            curp_cfg.gc_interval,
+        );
 
         let curp_c = Arc::clone(&curp);
         let shutdown_trigger_c = Arc::clone(&shutdown_trigger);
--- a/curp/src/server/gc.rs
+++ b/curp/src/server/gc.rs
@@ -8,13 +8,14 @@ use crate::{
     server::cmd_board::CmdBoardRef,
 };
 
-/// How often we should do gc
-const GC_INTERVAL: Duration = Duration::from_secs(20);
-
 /// Run background GC tasks for Curp server
-pub(super) fn run_gc_tasks<C: Command + 'static>(cmd_board: CmdBoardRef<C>, spec: SpecPoolRef<C>) {
-    let _spec_pool_gc = tokio::spawn(gc_spec_pool(spec, GC_INTERVAL));
-    let _cmd_board_gc = tokio::spawn(gc_cmd_board(cmd_board, GC_INTERVAL));
+pub(super) fn run_gc_tasks<C: Command + 'static>(
+    cmd_board: CmdBoardRef<C>,
+    spec: SpecPoolRef<C>,
+    gc_interval: Duration,
+) {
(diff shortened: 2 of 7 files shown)


utils/src/config.rs
```diff
@@ -198,10 +202,18 @@ pub fn default_curp_data_dir() -> PathBuf {
     PathBuf::from("/var/lib/curp")
 }
 
+/// default gc interval
+#[must_use]
+#[inline]
+pub fn default_gc_interval() -> Duration {
```
Phoenix500526, line 208: We can use `CTFE` to calculate the default gc interval at compile time

curp/src/server/gc.rs
```diff
@@ -1,15 +1,13 @@
...
-
-/// How often we should do gc
-const GC_INTERVAL: Duration = Duration::from_secs(20);
+use utils::config::CurpConfig;
 
 /// Run background GC tasks for Curp server
-pub(super) fn run_gc_tasks<C: Command + 'static>(cmd_board: CmdBoardRef<C>, spec: SpecPoolRef<C>) {
-    let _spec_pool_gc = tokio::spawn(gc_spec_pool(spec, GC_INTERVAL));
-    let _cmd_board_gc = tokio::spawn(gc_cmd_board(cmd_board, GC_INTERVAL));
+pub(super) fn run_gc_tasks<C: Command + 'static>(cmd_board: CmdBoardRef<C>, spec: SpecPoolRef<C>, curp_cfg: &Arc<CurpConfig>) {
```
Phoenix500526, line 8: You can directly pass the `gc_interval` to `run_gc_tasks` rather than wrapping it by an `Arc`

curp/src/server/curp_node.rs
```diff
@@ -289,7 +289,7 @@ impl<C: 'static + Command> CurpNode<C> {
             ))
         };
 
-        run_gc_tasks(Arc::clone(&cmd_board), Arc::clone(&spec_pool));
+        run_gc_tasks(Arc::clone(&cmd_board), Arc::clone(&spec_pool), &curp_cfg);
```
Phoenix500526, line 292: We can pass the `gc_interval` to the `run_gc_tasks` directly. Hence, we don't need to wrap the `curp_cfg` by `Arc`.

curp/src/server/gc.rs
Phoenix500526, line 8: Same as above.

curp/src/server/curp_node.rs
Phoenix500526, line 292: We can pass the gc_interval to the run_gc_tasks directly. Hence, we don't need to wrap the curp_cfg by Arc.

utils/src/config.rs
Phoenix500526, line 208: We can use CTFE to calculate the default gc interval at compile time

Phoenix500526 (changes requested):

curp/src/server/curp_node.rs
myrfy001: Will there be more gc related params needed in the future? If it won't, we can simply pass the interval value.

Phoenix500526: From the current situation, there will be only one param in this gc task.

Phoenix500526: > Will there be more gc related params needed in the future? If it won't, we can simply pass the interval value.

From the current situation, there will be only one param in this gc task.

Phoenix500526 (commented):
Please squash these three commits into one

utils/src/config.rs
```diff
@@ -198,10 +202,18 @@ pub fn default_curp_data_dir() -> PathBuf {
...
+    Duration::from_secs(20)
+}
+
 impl CurpConfig {
     /// Create a new server timeout
     #[must_use]
     #[inline]
+    #[allow(clippy::too_many_arguments)] // TODO: Fixme, should be changed into builder mode.
```
rogercloud, line 216: As it's a small PR, can we refator it to the builder mode here?

rogercloud (approved):

rogercloud (changes requested):
please refactor the builder mode for "too many arguments".

myrfy001:
> please refactor the builder mode for "too many arguments".

Ok, do you have any advices about how to do that? Write by hand or introducing a new dependency like `derive_builder = "0.12.0"`? I prefer using `derive_builder`.

Phoenix500526:
> > please refactor the builder mode for "too many arguments".
> 
> Ok, do you have any advices about how to do that? Write by hand or introducing a new dependency like `derive_builder = "0.12.0"`? I prefer using `derive_builder`.

I also prefer using `derive_builder`.

Phoenix500526 (approved):
LGTM

themanforfree (approved):

Phoenix500526:
Please rebase this pr @myrfy001

myrfy001:
@Phoenix500526 rebase done

Phoenix500526 (dismissed):
LGTM

themanforfree (approved):

rogercloud (dismissed):

Phoenix500526:
This pr has already been approved but is conflicted with the master branch. Please rebase it so we can merge it ASAP. @myrfy001

myrfy001:
@Phoenix500526  rebase Done

themanforfree (approved):

Phoenix500526 (approved):

Issue #202, opened 2023-03-15
Refactor lease store to


Phoenix500526:

Phoenix500526:
Resolve in pr #215

PR #215, opened 2023-03-22
refactor write operation, use reference in delete operations

themanforfree:
#202 

### DIFF
--- a/engine/src/engine_api.rs
+++ b/engine/src/engine_api.rs
@@ -3,11 +3,11 @@ use crate::error::EngineError;
 /// Write operation
 #[non_exhaustive]
 #[derive(Debug)]
-pub enum WriteOperation {
+pub enum WriteOperation<'a> {
     /// `Put` operation
     Put {
         /// The table name
-        table: &'static str,
+        table: &'a str,
         /// Key
         key: Vec<u8>,
         /// Value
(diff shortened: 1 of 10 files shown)


engine/src/engine_api.rs
```diff
@@ -1,13 +1,87 @@
 use crate::error::EngineError;
 
+// /// Write operation
```
rogercloud, line 3: Remove the useless comments.

engine/src/engine_api.rs
```diff
@@ -1,13 +1,87 @@
...
 #[derive(Debug)]
-pub enum WriteOperation {
+pub enum WriteOperation<'a> {
     /// `Put` operation
     Put {
         /// The table name
-        table: &'static str,
+        table: &'a str,
         /// Key
         key: Vec<u8>,
```
rogercloud, line 86: What's special for the `Put` operation for that the vector type is kept?

rogercloud (commented):
Actually this PR also removes the buffer_ops, which is a big part in the PR, please comment this change in the commit message.

engine/src/engine_api.rs
themanforfree: Fixed

rogercloud (approved):

PR #217, opened 2023-03-23
Normal AE will not retry.

markcty:
If ae failed, will now fallback to calibrate instead of retrying.

Also, we have only one hb_opt flag. This is not enough because failed server also needs hb. So we need hb flag for each server.

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -329,22 +329,26 @@ impl<C: 'static + Command> CurpNode<C> {
         let rpc_timeout = curp.cfg().rpc_timeout;
         let resps = hbs
             .into_iter()
-            .map(|(id, hb)| {
+            .filter_map(|(id, hb)| {
                 let connect = connects
                     .get(&id)
                     .cloned()
                     .unwrap_or_else(|| unreachable!("no server {id}'s connect"));
-                let req = AppendEntriesRequest::new_heartbeat(
+                let Ok(req) = AppendEntriesRequest::new(
                     hb.term,
                     hb.leader_id,
                     hb.prev_log_index,
                     hb.prev_log_term,
+                    hb.entries,
                     hb.leader_commit,
-                );
-                async move {
+                ) else {
+                    error!("failed to serialize ae");
+                    return None;
+                };
+                Some(async move {
                     let resp = connect.append_entries(req, rpc_timeout).await;
                     (id, resp)
-                }
+                })
             })
             .collect::<FuturesUnordered<_>>()
             .filter_map(|(id, resp)| async move {
@@ -499,6 +503,7 @@ impl<C: 'static + Command> CurpNode<C> {
         while let Some(i) = sync_rx.recv().await {
             let req = {
                 let Ok(ae) = curp.append_entries_single(i) else {
+                    error!("{} get log {i} failed", curp.id());
                     continue;
                 };
                 let req = AppendEntriesRequest::new(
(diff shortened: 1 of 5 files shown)


curp/src/server/curp_node.rs
```diff
@@ -517,59 +522,56 @@ impl<C: 'static + Command> CurpNode<C> {
...
-                    let resp = resp.into_inner();
-
-                    let result = curp.handle_append_entries_resp(
-                        connect.id(),
-                        Some(i),
-                        resp.term,
-                        resp.success,
-                        resp.hint_index,
-                    );
+        let resp = connect.append_entries(req, rpc_timeout).await;
```
Phoenix500526, line 547: In the original version, the `send_log_until_success` would retry `append_entries` infinitely until it succeeded, whereas there is no retry in the current version. Is this modification too extreme? I think it would be more friendly to specify a maximum number of retries, like 3.

curp/src/server/raw_curp/state.rs
```diff
@@ -51,8 +48,6 @@ pub(super) struct LeaderState {
     next_index: HashMap<ServerId, LogIndex>,
     /// For each server, index of highest log entry known to be replicated on server
     match_index: HashMap<ServerId, LogIndex>,
-    /// Servers that are being calibrated by the leader
-    pub(super) calibrating: HashSet<ServerId>,
```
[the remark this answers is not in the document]
markcty: It's the commit message.

> I removed calibrating field in leader state since we've already checked whether a calibration is on in calibrate_task

curp/src/server/raw_curp/mod.rs
```diff
@@ -219,21 +209,30 @@ impl<C: 'static + Command> RawCurp<C> {
...
+                let req = (
                     id.clone(),
                     AppendEntries {
                         term,
                         leader_id: self.id().clone(),
                         prev_log_index,
                         prev_log_term,
                         leader_commit: log_r.commit_index,
-                        entries: vec![],
+                        entries: log_r.get_from(next_index).unwrap_or_default().to_vec(),
```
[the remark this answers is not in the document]
markcty: The heartbeat is also a append_entires that can carry log entries. So I think it's better not to waste a heartbeat for bringing new entries to the follower.

curp/src/server/curp_node.rs
markcty: Removing retrying here does not mean that we will not retry. Actually, we will retry indefinitely in calibration task

curp/src/server/curp_node.rs
```diff
@@ -517,59 +522,56 @@ impl<C: 'static + Command> CurpNode<C> {
...
+    async fn send_log_to_follower(
         curp: Arc<RawCurp<C>>,
         connect: Arc<impl ConnectApi>,
         i: LogIndex,
         req: AppendEntriesRequest,
     ) {
-        let (rpc_timeout, retry_timeout) = (curp.cfg().rpc_timeout, curp.cfg().retry_timeout);
+        debug!("{} sends log[{i}] to follower {}", curp.id(), connect.id());
+        let rpc_timeout = curp.cfg().rpc_timeout;
         // send log[i] until succeed
```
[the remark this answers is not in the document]
markcty: Fixed

Phoenix500526 (approved):

curp/src/server/raw_curp/mod.rs
rogercloud, line 232: If we can calibrate in the heartbeat, do we still need a manually calibrate? Can we merge them into one function?

rogercloud (commented):
As we remove the infinite retry code in this PR, shall we add a new test to make sure the calibration really make do the same thing as the original "retry" function?

markcty:
> As we remove the infinite retry code in this PR, shall we add a new test to make sure the calibration really make do the same thing as the original "retry" function?

The integration test has already covered it.

curp/src/server/curp_node.rs
```diff
@@ -517,59 +522,55 @@ impl<C: 'static + Command> CurpNode<C> {
...
+        let resp = connect.append_entries(req, rpc_timeout).await;
 
-                    let result = curp.handle_append_entries_resp(
-                        connect.id(),
-                        Some(i),
-                        resp.term,
-                        resp.success,
-                        resp.hint_index,
-                    );
+        #[allow(clippy::unwrap_used)]
```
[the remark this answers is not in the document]
markcty: Fixed.

curp/src/server/raw_curp/mod.rs
markcty: Yes, we can merge them. Actually, I am planning to merge send_log_to_follower, heartbeat, and calibrate all together since all these functionalities use append_entries. And I think after merging them together, batching would be easier. Perhaps for another PR?

Phoenix500526: I can merge them in my new pr

Phoenix500526 (approved):

rogercloud (approved):

PR #222, opened 2023-03-29
Introduce batching logic in sync_task

### LINKED ISSUE #220: [Proposal] Improving Xline throughput in the high-load scenario
# Motivation
Improving Xline throughput in the high-load scenario.

# Implementation
There are three questions we should answer before we start discussing batching mechanism.

1. What does a batch look like?
    The so-called batch we mentioned above is an AppenEntries request which contains several `LogEntry` while every LogEntry has only one command. 

2. How to dispatch a batch？
    When one of the two following conditions is satisfied, a batch will be dispatched:
    1. When the size of a batch reaches its limit. 
       gRPC has imposed a 4MB cap on the size of a message. And in some raft implementations, like LogCabin, this message size cap is about 1MB([MAX_MESSAGE_LENGTH](https://github.com/logcabin/logcabin/blob/ee6c55ae9744b82b451becd9707d26c7c1b6bbfb/Protocol/Common.h#L78)). A  way better than hardcoding this limitation is to make it configurable.

    2. When the batching_timeout is exceeded. 
        Since we are using AppenEntries to optimize the Heartbeat, batching_timeout should not exceed the heartbeat_interval(300ms). Currently, the latency for a single request does not exceed 10ms. Therefore, a reasonable duration of batching_timeout could be from 15ms to 25ms. Same as batching_max_size, we can also make batching_timeout configurable. 

3. Who is in charge of collecting a batch and dispatching it？
    Currently, sync_task is responsible for replicating log entries, so we can place the  batch-collecting logic and dispatch strategy here. 

# Extra discussion
Q: How to retry if a batch dispatch failed?
A: If a batch dispatch failed, it will fall back to the calibrate and retry in there.

Q: What would happen if leadership changed during batch dispatching? 
A： The possible cases of leadership change are:
1. The original leader crashed:
  1. The original leader restarts immediately and becomes the new leader. It will execute the calibration process
  2. A follower will initiate an election and win the election finally, resulting in an increase in the current term. Since the unsent batch must not commit, the new leader can truncate the log in its calibration process. It does not violate the safety principle of the consensus algorithm.
  5. When a network partition occurs, the original leader enters the minority part, and the majority part elects a new leader. The original leader can send its batching log as usual, but these AppenEntries requests will never be accepted. Once the network partition is restored, the original leader will step down as a follower, and the new leader will truncate its log in the calibration process.

Phoenix500526:
Improving Xline throughput by batching in the high-load scenario. This pr is based on #217.  Close #220.

Improving Xline throughput by batching in the high-load scenario.

No

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -9,7 +9,7 @@ use thiserror::Error;
 use tokio::{
     sync::{broadcast, mpsc},
     task::JoinHandle,
-    time::MissedTickBehavior,
+    time::{sleep, MissedTickBehavior},
 };
 use tracing::{debug, error, info, warn};
 use utils::config::CurpConfig;
@@ -169,7 +169,7 @@ impl<C: 'static + Command> CurpNode<C> {
         let rand = thread_rng()
             .gen_range(0..heartbeat_interval.as_millis())
             .numeric_cast();
-        tokio::time::sleep(Duration::from_millis(rand)).await;
+        sleep(Duration::from_millis(rand)).await;
 
         let mut ticker = tokio::time::interval(heartbeat_interval);
         ticker.set_missed_tick_behavior(MissedTickBehavior::Delay);
@@ -490,7 +490,43 @@ impl<C: 'static + Command> CurpNode<C> {
                     }
                 }
             };
-            tokio::time::sleep(retry_timeout).await;
+            sleep(retry_timeout).await;
+        }
+    }
+
+    /// invoke `send_log_to_follower` for every follower
+    fn bcast_append_entries(
+        curp: &Arc<RawCurp<C>>,
+        connects: &HashMap<ServerId, Arc<impl ConnectApi>>,
+        first_sent_idx: LogIndex,
+        last_sent_idx: LogIndex,
+    ) {
+        let Ok(ae) = curp.append_entries_batch(first_sent_idx, last_sent_idx) else {
+            return;
+        };
+
+        let req = match AppendEntriesRequest::new(
+            ae.term,
+            ae.leader_id,
+            ae.prev_log_index,
+            ae.prev_log_term,
+            ae.entries,
+            ae.leader_commit,
+        ) {
+            Err(e) => {
+                error!("unable to serialize append entries request: {}", e);
+                return;
+            }
+            Ok(req) => req,
+        };
+
+        for connect in connects.values().cloned() {
+            let _handle = tokio::spawn(Self::send_log_to_follower(
+                Arc::clone(curp),
+                connect,
+                last_sent_idx,
+                req.clone(),
+            ));
         }
     }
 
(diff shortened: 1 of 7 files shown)


curp/src/server/curp_node.rs
```diff
@@ -225,7 +233,7 @@ impl<C: 'static + Command> CurpNode<C> {
         curp_cfg: Arc<CurpConfig>,
         tx_filter: Option<Box<dyn TxFilter>>,
     ) -> Result<Self, CurpError> {
-        let (sync_tx, sync_rx) = mpsc::unbounded_channel();
+        let (sync_tx, sync_rx) = flume::unbounded();
```
rogercloud, line 236: what's the difference of these two unbounded channels?

curp/src/server/curp_node.rs
```diff
@@ -490,49 +498,92 @@ impl<C: 'static + Command> CurpNode<C> {
...
-    /// Sync task is responsible for replicating log entries
-    async fn sync_task(
-        curp: Arc<RawCurp<C>>,
-        connects: HashMap<ServerId, Arc<impl ConnectApi>>,
-        mut sync_rx: mpsc::UnboundedReceiver<LogIndex>,
+    /// invoke `send_log_to_follower` for every follower
+    fn bcast_append_entries(
+        curp: &Arc<RawCurp<C>>,
+        connects: &HashMap<ServerId, Arc<impl ConnectApi>>,
+        batch: Vec<LogIndex>,
```
rogercloud, line 509: If the batch's contents are always the continuous number, can we do it better? For instance the first log index and the length.

curp/src/server/curp_node.rs
```diff
@@ -490,49 +498,92 @@ impl<C: 'static + Command> CurpNode<C> {
...
+        curp: Arc<RawCurp<C>>,
+        connects: HashMap<ServerId, Arc<impl ConnectApi>>,
+        sync_rx: flume::Receiver<(LogIndex, usize)>,
+    ) {
+        let batch_timeout = curp.cfg().batch_timeout;
+        let batch_max_size = curp.cfg().batch_max_size;
+        let connects = Arc::new(connects);
+        let timer = sleep(batch_timeout);
+        let batch = Mutex::new(Vec::new());
+        let batch_size = AtomicUsize::new(0);
```
rogercloud, line 553: Why do you use atomic number here? I don't think it'll be accessed parallelly.

utils/src/lib.rs
```diff
@@ -264,6 +264,42 @@ pub fn parse_rotation(s: &str) -> Result<RotationConfig, ConfigParseError> {
     }
 }
 
+/// Parse bytes from string
+/// # Errors
+/// Return error when parsing the given string to usize failed
+#[inline]
+#[allow(clippy::integer_arithmetic)]
+pub fn parse_batch_bytes(s: &str) -> Result<usize, ConfigParseError> {
+    if s.ends_with("kb") {
```
rogercloud, line 273: let's make it case insensitive.

curp/src/server/curp_node.rs
Phoenix500526: Fixed

curp/src/server/curp_node.rs
Phoenix500526: Fixed.

curp/src/server/raw_curp/mod.rs
```diff
@@ -683,27 +684,47 @@ impl<C: 'static + Command> RawCurp<C> {
         self.ctx.leader_tx.subscribe()
     }
 
-    /// Get `append_entries` request for log[i]
+    /// Get `append_entries` request for log batch
```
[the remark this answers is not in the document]
Phoenix500526: Fixed

utils/src/lib.rs
Phoenix500526: Fixed

utils/src/lib.rs
```diff
@@ -264,6 +264,42 @@ pub fn parse_rotation(s: &str) -> Result<RotationConfig, ConfigParseError> {
...
+        if let Some(value) = s.strip_suffix("kb") {
+            Ok(value.parse::<usize>()? * 1024)
+        } else {
+            Err(ConfigParseError::InvalidValue(format!(
+                "the value of size should not be empty. ({s})"
+            )))
+        }
+    } else if s.ends_with("mb") {
```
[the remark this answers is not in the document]
Phoenix500526: Fixed

curp/src/server/curp_node.rs
Phoenix500526: Fixed

Phoenix500526:
New batching is in pr #227

Issue #223, opened 2023-03-30
[Feature]: Unified AE entry


markcty:
### Description about the feature

Whatever the upper level functions may appear(batch ae, heartbeat, calibrate), they all use append_entires(AE) as bottom level transporting structure. And however the upper level functions may be triggered(needs calibrate, new log entry arrives, heartbeat timeout), they all use AE to replicate the latest log entries  to the follower. Therefore, we should unify all AE entries together in one place.

In addition, currently when new log entires arrive, we will send out AEs carrying the new log entires regardless of the follower’s condition. This is bad since different follower’s next_index may vary and followers that are lagging behind may reject the AE. What’s worse, the calibrating task of a follower maybe ongoing while we send these AEs that carry the latest log entires. These two situations make sending out latest AEs completely meaningless.

As a solution, there should be one and only one entry(function) to generate the AE for each follower. The entry(function) will automatically detect current follower’s status and send out AE on demand. When batching is added, the function will detect the ae size to hold as many log entries as possible.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #224, opened 2023-03-30
Unified AE entry

markcty:
As described in #223, this PR unify all AE generation functions to one, which will generate ae according to next_index only. And now, one sync task corresponds to one follower now.

Based on #217.

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -3,12 +3,12 @@ use std::{collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
 use clippy_utilities::NumericCast;
 use event_listener::Event;
 use futures::{pin_mut, stream::FuturesUnordered, StreamExt};
+use itertools::Itertools;
 use madsim::rand::{thread_rng, Rng};
 use parking_lot::{Mutex, RwLock};
 use thiserror::Error;
 use tokio::{
     sync::{broadcast, mpsc},
-    task::JoinHandle,
     time::MissedTickBehavior,
 };
 use tracing::{debug, error, info, warn};
@@ -18,7 +18,7 @@ use super::{
     cmd_board::{CmdBoardRef, CommandBoard},
     cmd_worker::start_cmd_workers,
     gc::run_gc_tasks,
-    raw_curp::{AppendEntries, RawCurp, TickAction, Vote},
+    raw_curp::{RawCurp, Vote},
     spec_pool::{SpecPoolRef, SpeculativePool},
     storage::{StorageApi, StorageError},
 };
@@ -32,7 +32,7 @@ use crate::{
         WaitSyncedRequest, WaitSyncedResponse,
     },
     server::storage::rocksdb::RocksDBStorage,
-    LogIndex, ServerId, TxFilter,
+    ServerId, TxFilter,
 };
 
 /// Uncommitted pool type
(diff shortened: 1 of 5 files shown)


curp/src/server/curp_node.rs
```diff
@@ -173,42 +173,101 @@ impl<C: 'static + Command> CurpNode<C> {
...
+            // a sync is either triggered by an heartbeat timeout event or when new log entries arrive
+            tokio::select! {
+                _now = ticker.tick() => {
+                    if hb_opt {
+                        hb_opt = false;
+                        continue;
+                    }
+                }
+                // TODO: use a batch timer
+                _ = sync_event.listen() => {}
```
rogercloud, line 227: The sync_event is only used in the `handle_vote_response` function. Shall we give it a more meaningful name?

rogercloud: It may miss some event during the loop as an event may be dropped if there's no listener waiting.

markcty: > The sync_event is only used in the `handle_vote_response` function. Shall we give it a more meaningful name?

It is also used in `handle_propose`. I think this line of code is miss deleted when rebasing and now I've added it. `sync_event` means the leader should trigger a sync with the follower.

markcty: > It may miss some event during the loop as an event may be dropped if there's no listener waiting.

Fixed.

curp/src/server/curp_node.rs
```diff
@@ -173,42 +173,101 @@ impl<C: 'static + Command> CurpNode<C> {
```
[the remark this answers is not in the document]
markcty: Renamed

curp/src/server/curp_node.rs
```diff
@@ -173,42 +176,109 @@ impl<C: 'static + Command> CurpNode<C> {
...
+                    if hb_opt {
+                        hb_opt = false;
+                        continue;
+                    }
+                }
+                // TODO: use a batch timer
+                _ = listener.take().unwrap() => {
+                }
+            }
+            listener = Some(sync_event.listen());
```
themanforfree, line 239: Lines 239 and 224-226 seem to be doing the same thing, I think we can remove line 239, or we can remove line 224-226 and init listener at line 219

curp/src/server/curp_node.rs
```diff
@@ -295,19 +362,26 @@ impl<C: 'static + Command> CurpNode<C> {
...
-            let tick_task = tokio::spawn(Self::tick_task(Arc::clone(&curp_c), connects.clone()));
-            let sync_task = tokio::spawn(Self::sync_task(
-                Arc::clone(&curp_c),
-                connects.clone(),
-                sync_rx,
-            ));
-            let calibrate_task = tokio::spawn(Self::calibrate_task(curp_c, connects, calibrate_rx));
+            let connects = rpc::connect(others.clone(), tx_filter).await;
+            let election_task =
+                tokio::spawn(Self::tick_election(Arc::clone(&curp_c), connects.clone()));
```
themanforfree, line 367: I suggest renaming tick_election to election_task

curp/src/server/curp_node.rs
```diff
@@ -173,42 +176,109 @@ impl<C: 'static + Command> CurpNode<C> {
...
+                Arc::clone(&curp),
+                Arc::clone(&connect),
+                Arc::clone(&sync_event),
+            )
+            .await;
+        }
+    }
+
+    /// real sync task, will stop if self is no long the leader
+    async fn _sync_task(
```
themanforfree, line 208: The leading `_` has has a special meaning in rust, so I think the naming here is not very good, I suggest use `inner_sync_task`

curp/src/server/curp_node.rs
markcty: Fixed

curp/src/server/curp_node.rs
markcty: Changed to sync_follower

curp/src/server/curp_node.rs
markcty: Fixed

Phoenix500526 (approved):

themanforfree (approved):

PR #226, opened 2023-04-04
fix: clear curp tmp directory after test

themanforfree:

After running the test, many useless directories will be left under the tmp directory
clear curp tmp directory after test
no

### DIFF
--- a/xline/tests/common/mod.rs
+++ b/xline/tests/common/mod.rs
@@ -22,6 +22,8 @@ pub struct Cluster {
     stop_tx: Option<Sender<()>>,
     /// Cluster size
     size: usize,
+    /// storage paths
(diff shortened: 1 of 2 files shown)


xline/tests/common/mod.rs
```diff
@@ -112,6 +126,12 @@ impl Cluster {
     }
 }
 
+impl Drop for Cluster {
+    fn drop(&mut self) {
+        self.stop();
```
Phoenix500526, line 131: I think we can inline `stop` method here.

themanforfree: fixed

Phoenix500526 (approved):

markcty (approved):

PR #227, opened 2023-04-04
New batching

### LINKED ISSUE #220: [Proposal] Improving Xline throughput in the high-load scenario

Phoenix500526:
Improving Xline throughput by batching in the high-load scenario. This pr is based on #224 . Close #220 

Improving Xline throughput by batching in the high-load scenario.

  No

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -307,12 +307,13 @@ impl<C: 'static + Command> CurpNode<C> {
         let mut ticker = tokio::time::interval(curp.cfg().heartbeat_interval);
         ticker.set_missed_tick_behavior(MissedTickBehavior::Delay);
         let id = connect.id();
+        let batch_timeout = curp.cfg().batch_timeout;
 
         #[allow(clippy::integer_arithmetic, clippy::unwrap_used)]
         // tokio select internal triggered,
         loop {
             // grab the listener in the beginning to prevent missing sync events
-            let listener = sync_event.listen();
+            let listener: event_listener::EventListener = sync_event.listen();
 
             let Ok(sync_action) = curp.sync(id) else {
                 return;
@@ -320,14 +321,20 @@ impl<C: 'static + Command> CurpNode<C> {
 
             match sync_action {
                 SyncAction::AppendEntries(ae) => {
-                    let result = Self::send_ae(connect.as_ref(), curp.as_ref(), ae).await;
-                    if let Err(err) = result {
-                        warn!("ae to {} failed, {err}", connect.id());
-                        if matches!(err, SendAEError::NotLeader) {
-                            return;
+                    // (hb_opt, entries) status combination
+                    // (false, empty) => send heartbeat to followers
+                    // (true, empty) => indicates that `batch_timeout` expired, and during this period there is not any log generated. Do nothing
+                    // (true | false, not empty) => send append entries
+                    if !hb_opt || !ae.entries.is_empty() {
+                        let result = Self::send_ae(connect.as_ref(), curp.as_ref(), ae).await;
+                        if let Err(err) = result {
+                            warn!("ae to {} failed, {err}", connect.id());
+                            if matches!(err, SendAEError::NotLeader) {
+                                return;
+                            }
+                        } else {
+                            hb_opt = true;
                         }
-                    } else {
-                        hb_opt = true;
                     }
                 }
                 SyncAction::Snapshot(rx) => match rx.await {
@@ -350,13 +357,13 @@ impl<C: 'static + Command> CurpNode<C> {
             // a sync is either triggered by an heartbeat timeout event or when new log entries arrive
             tokio::select! {
                 _now = ticker.tick() => {
-                    if hb_opt {
-                        hb_opt = false;
-                        continue;
+                    hb_opt = false;
+                }
+                res = tokio::time::timeout(batch_timeout, listener) => {
+                    if let Err(_e) = res {
+                        hb_opt = true;
                     }
                 }
-                // TODO: use a batch timer
-                _ = listener => {}
             }
         }
     }
@@ -374,7 +381,10 @@ impl<C: 'static + Command> CurpNode<C> {
         curp_cfg: Arc<CurpConfig>,
         tx_filter: Option<Box<dyn TxFilter>>,
     ) -> Result<Self, CurpError> {
-        let sync_event = Arc::new(Event::new());
+        let sync_events = others
+            .keys()
+            .map(|server_id| (server_id.clone(), Arc::new(Event::new())))
+            .collect();
         let (log_tx, log_rx) = mpsc::unbounded_channel();
         let shutdown_trigger = Arc::new(Event::new());
         let cmd_board = Arc::new(RwLock::new(CommandBoard::new()));
@@ -399,7 +409,7 @@ impl<C: 'static + Command> CurpNode<C> {
                 uncommitted_pool,
                 Arc::clone(&curp_cfg),
                 Box::new(ce_event_tx.clone()),
-                Arc::clone(&sync_event),
+                sync_events,
                 log_tx,
             ))
         } else {
@@ -416,9 +426,9 @@ impl<C: 'static + Command> CurpNode<C> {
                 Arc::clone(&cmd_board),
                 Arc::clone(&spec_pool),
                 uncommitted_pool,
-                Arc::clone(&curp_cfg),
+                &curp_cfg,
                 Box::new(ce_event_tx.clone()),
-                Arc::clone(&sync_event),
+                sync_events,
                 log_tx,
                 voted_for,
                 entries,
@@ -448,12 +458,12 @@ impl<C: 'static + Command> CurpNode<C> {
             let election_task =
                 tokio::spawn(Self::election_task(Arc::clone(&curp_c), connects.clone()));
             let sync_task_daemons = connects
-                .into_values()
-                .map(|connect| {
+                .into_iter()
+                .map(|(server_id, connect)| {
                     tokio::spawn(Self::sync_follower_daemon(
                         Arc::clone(&curp_c),
                         connect,
-                        Arc::clone(&sync_event),
+                        curp_c.sync_event(&server_id),
                     ))
                 })
                 .collect_vec();
--- a/curp/src/server/raw_curp/log.rs
+++ b/curp/src/server/raw_curp/log.rs
@@ -14,6 +14,10 @@ use crate::{
     LogIndex,
 };
 
+/// Given that the log vector is used to store a large number of log entries, we can specify a suitable capacity to avoid unnecessary allocation.
+/// The initial capacity of the log entries
+const LOG_INIT_CAPACITY: usize = 8192;
+
 /// Curp logs
 /// There exists a fake log entry 0 whose term equals 0
 /// For the leader, there should never be a gap between snapshot and entries
@@ -36,7 +40,7 @@ pub(super) struct Log<C: Command> {
     log_tx: mpsc::UnboundedSender<LogEntry<C>>,
     /// Prefix sum vector of log entries
     batch_index: Vec<u64>,
-    /// batch size limit
+    /// Batch size limit
     batch_limit: u64,
 }
 
@@ -53,15 +57,17 @@ impl<C: Command> Debug for Log<C> {
 }
 
 impl<C: 'static + Command> Log<C> {
-    /// Create a new log
-    pub(super) fn new(
+    /// Recover log from the given entires
+    pub(super) fn recover(
         log_tx: mpsc::UnboundedSender<LogEntry<C>>,
         entries: Vec<LogEntry<C>>,
         batch_limit: u64,
     ) -> Self {
-        let mut batch_index = vec![0u64];
+        let mut batch_index = Vec::with_capacity(entries.capacity());
+        batch_index.push(0);
         for entry in &entries {
             #[allow(clippy::expect_used)]
+            // it's in the initialization stage, panic doesn't affect much
             let entry_size =
                 serialized_size(entry).expect("log entry {entry:?} cannot be serialized");
             if let Some(cur_size) = batch_index.last() {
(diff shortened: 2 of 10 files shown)


curp/src/server/raw_curp/log.rs
```diff
@@ -181,11 +239,18 @@ impl<C: 'static + Command> Log<C> {
             (entry.index, entry.term)
         }
     }
+
+    #[cfg(test)]
+    fn set_batch_limit(&mut self, batch_limit: u64) {
+        self.batch_limit = batch_limit;
+    }
```
markcty, line 246: We can move this function to line 260

curp/src/server/curp_node.rs
```diff
@@ -323,7 +328,7 @@ impl<C: 'static + Command> CurpNode<C> {
         );
 
         // create curp state machine
-        let (voted_for, entries) = storage.recover().await?;
+        let (voted_for, entries, prefix_sum) = storage.recover().await?;
```
markcty, line 331: Should we recover the prefix sum in the `Log::new()` function not in the storage layer? Because prefix sum seems to be an optimization used by `Log` instead of somthing that the storage layer should be awared of.

curp/src/server/raw_curp/mod.rs
```diff
@@ -127,8 +127,8 @@ struct Context<C: Command> {
     election_tick: AtomicU8,
     /// Tx to send cmds to execute and do after sync
     cmd_tx: Box<dyn CEEventTxApi<C>>,
-    /// Sync event trigger
-    sync_event: Arc<Event>,
+    /// Followers sync event trigger
+    sync_event: Arc<HashMap<ServerId, Arc<Event>>>,
```
markcty, line 131: We can remove the outer Arc.

curp/src/server/curp_node.rs
Phoenix500526: The prefix sum can be viewed as a batch index,  which comes with the log entries. On the performance ground, I rebuild the prefix sum in the recovery process in that it will read the persisted log entries out and serialize them. By doing so we don't need to iterate the log entries again and calculate the serialized size of every single log entry in `Log::new`. But prefix sum is not a good name? Perhaps, batch_index would be better?

curp/src/server/raw_curp/log.rs
Phoenix500526: Fixed

curp/src/server/raw_curp/mod.rs
Phoenix500526: Fixed

curp/src/server/curp_node.rs
markcty: I still think it's a better idea to calculate the prefix sum in `log.rs`. The performance cost you mentioned is negligible in that the number of recovered log entries will be quite small with gc enabled and the calculation will be performed only once at startup. But the concept of `prefix_sum`, IMO, is still something that the storage layer should not be aware of. The prefix sum does not come with the log entires, but is calculated from the log entries.

curp/src/server/storage/mod.rs
```diff
@@ -31,7 +31,14 @@ pub(super) trait StorageApi: Send + Sync {
     /// Return `voted_for` and all log entries
```
themanforfree, line 31: Update comment with new return value

curp/src/server/curp_node.rs
```diff
@@ -280,8 +281,7 @@ impl<C: 'static + Command> CurpNode<C> {
                         continue;
                     }
                 }
-                // TODO: use a batch timer
-                _ = listener => {
+                _ = tokio::time::timeout(batch_timeout, listener) => {
```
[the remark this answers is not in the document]
Phoenix500526: Fixed

curp/src/server/storage/mod.rs
Phoenix500526: Fixed

curp/src/server/raw_curp/mod.rs
```diff
@@ -775,13 +797,14 @@ impl<C: 'static + Command> RawCurp<C> {
...
-                index,
-            );
+            if let Ok(index) = log.push_cmd(term, Arc::clone(&cmd)) {
+                debug!(
+                    "{} recovers speculatively executed cmd({}) in log[{}]",
+                    self.id(),
+                    cmd.id(),
+                    index,
+                );
+            }
```
markcty, line 807: Shall we use `expect` on `push_cmd` if it returns false? It seems like an intolerable fault if the cmds can't be recovered correctly.

scripts/benchmark.sh
```diff
@@ -6,15 +6,10 @@ MEMBERS="node1=${SERVERS[1]}:2379,node2=${SERVERS[2]}:2379,node3=${SERVERS[3]}:2
...
-    "node2  false node2=${SERVERS[2]}:2379"
-    "client false node3=${SERVERS[3]}:2379"
     "client true  ${MEMBERS}"
 )
 ETCD_TESTCASE=(
-    "node1  false node1=${SERVERS[1]}:2379"
-    "node2  false node2=${SERVERS[2]}:2379"
-    "client false node3=${SERVERS[3]}:2379"
+    "client false node1=${SERVERS[1]}:2379"
 )
```
markcty, line 13: Why are these benchmarks removed?

curp/src/server/curp_node.rs
```diff
@@ -226,6 +228,10 @@ impl<C: 'static + Command> CurpNode<C> {
             };
 
             let result = async {
+                if batch_timeout_flag && ae.entries.is_empty() {
+                    batch_timeout_flag = false;
+                    return Ok(());
+                }
```
markcty, line 234: I guess you don't need another flag here. Just use `hb_opt` flag here.

curp/src/server/raw_curp/mod.rs
Phoenix500526: Fixed

scripts/benchmark.sh
Phoenix500526: Fixed

curp/src/server/curp_node.rs
Phoenix500526: Fixed

curp/src/server/curp_node.rs
```diff
@@ -352,11 +361,13 @@ impl<C: 'static + Command> CurpNode<C> {
                 _now = ticker.tick() => {
                     if hb_opt {
                         hb_opt = false;
-                        continue;
                     }
```
markcty, line 364: No need for if. Just `hb_opt = false;`

utils/src/config.rs
```diff
@@ -146,6 +173,21 @@ pub const fn default_heartbeat_interval() -> Duration {
     Duration::from_millis(300)
 }
 
+/// default batch timeout
+#[must_use]
+#[inline]
+pub const fn default_batch_timeout() -> Duration {
+    Duration::from_millis(1)
```
markcty, line 180: Is 1 millis too extreme? It will result in a busy leader. If our default heartbeat timeout is 300ms, perhaps 50ms~100ms would be more suitable?

curp/src/server/curp_node.rs
Phoenix500526: Fixed

utils/src/config.rs
Phoenix500526: I think 50ms ~ 100ms may be too large. The original default value is 15ms. But in the benchmark practice, 1ms is better than 15ms in the conflict test case, saving about 20ms latency per wait synced request. And in the non-conflict test case, the performance is also quite good.

curp/src/server/curp_node.rs
```diff
@@ -307,27 +307,36 @@ impl<C: 'static + Command> CurpNode<C> {
...
+                            if matches!(err, SendAEError::NotLeader) {
+                                return;
+                            }
+                        } else {
+                            hb_opt = true;
                         }
                     } else {
-                        hb_opt = true;
+                        hb_opt = false;
                     }
```
markcty, line 340: And here hb_opt should not be set to false? If we skip an ae here, it doesn't mean that next ae should not be skipped. The hb_opt should only be set to false when heartbeat_timeout is reached?

curp/src/server/curp_node.rs
```diff
@@ -350,13 +359,13 @@ impl<C: 'static + Command> CurpNode<C> {
...
                 _now = ticker.tick() => {
-                    if hb_opt {
-                        hb_opt = false;
-                        continue;
+                    hb_opt = false;
+                }
+                res = tokio::time::timeout(batch_timeout, listener) => {
+                    if let Err(_e) = res {
+                        hb_opt = true;
                     }
```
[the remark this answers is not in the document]
Phoenix500526: If the `batch_timeout` expires while there is not any log generated during this period, it will send an empty append_entries request to followers. That's not what we want. Whether or not a request is sent depends on two factors: (hb_opt, ae.entries.is_empty).

curp/src/server/curp_node.rs
Phoenix500526: Fixed. The reason why I set `hb_opt` to false is that we had an if statement in the select block. And now it's gone, so we don't set it again.

utils/src/config.rs
markcty: If we want the wait sync to be as fast as possible, then I guess we don't need the batch from the beginning? We can just send out log entry as soon as possible. I guess batching is somehow a trade off between sync speed and system stability. With batch, we can have fewer append_entries sent, which can possibly result in a higher throughput when there is little conflict between cmds.

So I guess the question here is should we still take sync speed as a criteria of our benchmark? With batching, we indeed have a lower sync speed then etcd, but we can have a higher throughput when there's no conflict. If this reasoning stands, than we can have a longer `batch_timeout`.

Phoenix500526: Actually, I tried to send an `ae` immediately when the curp realize that the proposal conflicted with the proposal it has received before. But the benchmark result is not good.

xline/src/main.rs
```diff
@@ -162,43 +163,49 @@ struct ServerArgs {
     /// Open jaeger online
     #[clap(long)]
     jaeger_online: bool,
-    /// Trace level of jaeger
+    /// Trace level of  [default: info]
```
themanforfree, line 166: of what

xline/src/main.rs
```diff
@@ -162,43 +163,49 @@ struct ServerArgs {
...
+    /// Curp rpc timeout [default: 50ms]
     #[clap(long, value_parser = parse_duration)]
     rpc_timeout: Option<Duration>,
+    /// Curp append entries batch timeout [default: 1ms]
+    #[clap(long, value_parser = parse_duration)]
+    batch_timeout: Option<Duration>,
+    /// Curp append entries batch max size [default: 2MB]
+    #[clap(long, value_parser = parse_batch_bytes)]
+    batch_max_size: Option<u64>,
     /// Follower election timeout ticks
```
themanforfree, line 196: Why is this comment not updated

xline/src/main.rs
```diff
@@ -162,43 +163,49 @@ struct ServerArgs {
...
     #[clap(long, default_value_t = default_follower_timeout_ticks())]
     follower_timeout_ticks: u8,
     /// Candidate election timeout ticks
```
themanforfree, line 199: Same above, and other args too

xline/src/main.rs
Phoenix500526: Fixed

xline/src/main.rs
Phoenix500526: That's because the default value of follower election timeout ticks is set by `default_value_t`, we don't need to describe it in the comment.

xline/src/main.rs
Phoenix500526: Fixed

markcty (dismissed):

curp/src/server/raw_curp/log.rs
```diff
@@ -52,14 +57,27 @@ impl<C: 'static + Command> Log<C> {
     pub(super) fn new(
         log_tx: mpsc::UnboundedSender<LogEntry<C>>,
         entries: Vec<LogEntry<C>>,
+        batch_limit: u64,
     ) -> Self {
+        let mut batch_index = vec![0u64];
+        for entry in &entries {
+            #[allow(clippy::expect_used)]
```
rogercloud, line 64: Add a comment to explain why. For instance, "it's in the initialization stage, panic doesn't affect much".

curp/src/server/raw_curp/log.rs
```diff
@@ -122,8 +140,21 @@ impl<C: 'static + Command> Log<C> {
                 continue;
             }
 
+            let Ok(entry_size) = serialized_size(&entry) else {
```
rogercloud, line 143: Why do we use a different way to handle serialized_size result? In the above change `expect` is used.

themanforfree (approved):

curp/src/server/raw_curp/log.rs
Phoenix500526: Fixed

curp/src/server/raw_curp/log.rs
Phoenix500526: Fixed

themanforfree (approved):

themanforfree (approved):

rogercloud (approved):

PR #229, opened 2023-04-06
Refactor leader state

markcty:
now every followers' next_index and match_index corresponds to one mutex, will reduce lock contention when the leader is handling ae responses. This change won't affect correctness in the sense that each follower's state(next_index, match_index) is independent of each other.


curp/src/server/raw_curp/state.rs
```diff
@@ -44,13 +45,20 @@ pub(super) struct CandidateState<C> {
...
+
 /// Additional state for the leader, all volatile
 #[derive(Debug)]
 pub(super) struct LeaderState {
-    /// For each server, index of the next log entry to send to that server
-    next_index: HashMap<ServerId, LogIndex>,
-    /// For each server, index of highest log entry known to be replicated on server
-    match_index: HashMap<ServerId, LogIndex>,
+    /// For each server, the leader maintains its status
+    statues: HashMap<ServerId, Mutex<FollowerStatus>>,
```
Phoenix500526, line 61: Typo. It should be "status"

Phoenix500526 (changes requested):
Plz update the comment ("READ THIS BEFORE YOU START WRITING CODE FOR THIS MODULE ") in the mod.rs file.

curp/src/server/raw_curp/state.rs
markcty: Fixed

markcty:
>

markcty:
> Plz update the comment ("READ THIS BEFORE YOU START WRITING CODE FOR THIS MODULE ") in the mod.rs file.

Sure. Updated.

Phoenix500526 (approved):

PR #234, opened 2023-04-18
refactor snapshot at engine level, and fix todo in `CommandExecutor`

themanforfree:

add real implementation of snapshot and reset in the command executor.
refactor snapshot at engine level
no

### DIFF
--- a/curp/src/server/storage/rocksdb.rs
+++ b/curp/src/server/storage/rocksdb.rs
@@ -6,8 +6,8 @@ use std::{
 use async_trait::async_trait;
 use chrono::Local;
 use engine::{
-    engine_api::SnapshotApi,
-    rocksdb_engine::{RocksEngine, RocksSnapshot},
+    rocksdb_engine::RocksEngine,
+    snapshot_api::{RocksSnapshot, SnapshotProxy},
     StorageEngine, WriteOperation,
 };
 use uuid::Uuid;
@@ -81,14 +81,14 @@ impl<C: 'static + Command> StorageApi for RocksDBStorage<C> {
         Ok((voted_for, entries))
     }
 
-    async fn new_snapshot(&self) -> Result<Box<dyn SnapshotApi>, StorageError> {
+    async fn new_snapshot(&self) -> Result<SnapshotProxy, StorageError> {
         // TODO: delete outdated snapshot
         // TODO: better snapshot file naming
         let dir = self
             .data_dir
             .join(format!("snapshot-{}-{}", Local::now(), Uuid::new_v4()));
         let snapshot = RocksSnapshot::new_for_receiving(dir)?;
-        Ok(Box::new(snapshot))
+        Ok(SnapshotProxy::Rocks(snapshot))
     }
 }
 
--- a/engine/src/snapshot_api/mod.rs
+++ b/engine/src/snapshot_api/mod.rs
@@ -1,0 +1,78 @@
+pub use self::{memory_snapshot::MemorySnapshot, rocks_snapshot::RocksSnapshot};
+
+/// Memory snapshot
+mod memory_snapshot;
+/// Rocksdb snapshot
+mod rocks_snapshot;
+
+/// This trait is a abstraction of the snapshot, We can Read/Write the snapshot like a file.
+#[async_trait::async_trait]
+pub trait SnapshotApi: Send + Sync + std::fmt::Debug {
+    /// Get the size of the snapshot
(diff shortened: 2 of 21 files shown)


curp/src/server/storage/rocksdb.rs
```diff
@@ -81,14 +81,14 @@ impl<C: 'static + Command> StorageApi for RocksDBStorage<C> {
         Ok((voted_for, entries))
     }
 
-    async fn new_snapshot(&self) -> Result<Box<dyn SnapshotApi>, StorageError> {
+    async fn new_snapshot(&self) -> Result<SnapshotUnion, StorageError> {
         // TODO: delete outdated snapshot
         // TODO: better snapshot file naming
```
Phoenix500526, line 86: Shall we finish these two todos in this release?

themanforfree: this method has some problem, it use Curp storage create xline snapshot, I will refactor this in other PR

Phoenix500526: Ok, please leave an issue to track this problem.

themanforfree: #235

Phoenix500526 (dismissed):

engine/src/snapshot_api/mod.rs
```diff
@@ -0,0 +1,80 @@
+pub use self::{memory_snapshot::MemorySnapshot, rocks_snapshot::RocksSnapshot};
+
+/// Memory snapshot
+mod memory_snapshot;
+/// Rocksdb snapshot
+mod rocks_snapshot;
+
+/// This trait is a abstraction of the snapshot, We can Read/Write the snapshot like a file.
+#[async_trait::async_trait]
+#[allow(clippy::pedantic)] // In order to unify the naming
```
rogercloud, line 10: I tried remove this line, clippy doesn't warn or report errors.

engine/src/snapshot_api/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    async fn write_all(&mut self, buf: &[u8]) -> std::io::Result<()>;
+
+    /// Clean files of current snapshot
+    async fn clean(&mut self) -> std::io::Result<()>;
+}
+
+/// Union of snapshot types, hiding the difference between different implementations
+#[derive(Debug)]
+#[non_exhaustive]
+#[allow(clippy::pedantic)] // In order to unify the naming
```
rogercloud, line 31: Same as above.

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -0,0 +1,382 @@
+use std::{
```
rogercloud, line 1: Please add some tests to cover the logic in this file. Because memory_snapshot is too simple, only the rocks is required.

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -0,0 +1,382 @@
...
+    fn size(&self) -> u64 {
+        self.snap_files
+            .iter()
+            .map(|f| f.size)
+            .sum::<u64>() // sst files size
+            .overflow_add(self.meta.len().numeric_cast()) // meta size
+    }
+
+    #[inline]
+    #[allow(clippy::indexing_slicing)] // safe indexing
```
rogercloud, line 323: Please explain why it's safe. For instance, the return value of `read` function is not larger than the input buf's length.

rogercloud: And we should also try to put the constraint  in the `read` function. Do the same thing for the `write` function.

engine/src/snapshot_api/mod.rs
themanforfree: Removed

engine/src/snapshot_api/mod.rs
themanforfree: Removed

engine/src/snapshot_api/rocks_snapshot.rs
themanforfree: Added

engine/src/snapshot_api/rocks_snapshot.rs
themanforfree: Fixed

Phoenix500526 (approved):

rogercloud (approved):

PR #241, opened 2023-04-24
Use `Vec<u8>` as key and value instead of `u32` in tests

themanforfree:

Some pre-work for refactoring the engine, the next PR will use the engine to replace the store, and the engine cannot use u32 as a key
Use `Vec<u8>` as key and value instead of `u32` in Curp tests
no

### DIFF
--- a/curp/src/test_utils/test_cmd.rs
+++ b/curp/src/test_utils/test_cmd.rs
@@ -135,7 +135,7 @@ impl ConflictCheck for TestCommand {
 pub(crate) struct TestCE {
     server_id: ServerId,
     last_applied: Arc<AtomicU64>,
-    pub(crate) store: Arc<Mutex<HashMap<u32, u32>>>,
+    pub(crate) store: Arc<Mutex<HashMap<Vec<u8>, Vec<u8>>>>,
     exe_sender: mpsc::UnboundedSender<(TestCommand, TestCommandResult)>,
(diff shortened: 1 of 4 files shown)


curp/src/cmd.rs
```diff
@@ -97,7 +97,7 @@ impl ConflictCheck for String {
     }
 }
 
-impl ConflictCheck for u32 {
+impl ConflictCheck for Vec<u8> {
```
Phoenix500526, line 100: I can't get the point of why the engine cannot use an u32 integer as a key.

themanforfree: Next PR will use `Engine` as `TestCE`'s store, and the `Engine` can only save `Vec<u8>` data,  Another possible solution is to do the conversion between u32 and Vec<u8> when data is written or read. what do you think

rogercloud: Is affected by one engine implementation？Or is it a trait interface constration?

themanforfree: `Curp`'s test directly uses `MemorySnapshot`, and it serializes the store by itself and writes it to the snapshot. This is not expected, so I will replace the store in the `Curp`'s test wit `MemoryEngine`. Our `MemoryEngine` can only read and write `Vec<u8>`, you can't use `u32` directly, so we need to fix this.
And I rewrote this PR by type conversion, now it looks easier

rogercloud (approved):

Phoenix500526 (approved):

PR #244, opened 2023-04-25
fix snapshot restore test

### LINKED ISSUE #240: [Bug]: The test case "test_snapshot_and_restore" sometimes fails.
### Description about the bug

As the issue title, the test case "test_snapshot_and_restore" sometimes fails for unknown reasons in our CI workflow. I'm trying to reproduce this problem.

### Version

0.1.0 (Default)

### Relevant log output

```shell
---- test_snapshot_and_restore stdout ----
thread 'test_snapshot_and_restore' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`', xline/tests/maintenance.rs:39:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Code of Conduct

themanforfree:

Fix #240 
the maintenance_client sends a request directly to the leader, because other nodes may not have synced yet
no

### DIFF
--- a/xline/tests/maintenance.rs
+++ b/xline/tests/maintenance.rs
@@ -20,8 +20,11 @@ async fn test_snapshot_and_restore() -> Result<(), Box<dyn std::error::Error>> {
         cluster.start().await;
         let client = cluster.client().await;
         let _ignore = client.put(PutRequest::new("key", "value")).await?;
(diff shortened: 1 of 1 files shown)


Phoenix500526:
Please update the root cause of issue #240 in the comment.

Phoenix500526 (approved):

rogercloud (approved):

PR #245, opened 2023-04-26
refactor stores

themanforfree:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
remove channel between stores
* what changes does this pull request make?
share lease_collections between stores, remove LeaseMessage and the channel, remove unnecessary async
* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)
no

PR #246, opened 2023-04-27
Refactor/dependencies

### LINKED ISSUE #250: Refactor: refactor the dependencies of kv server
Currently, there is a confusion of dependencies between several components in our kv server. Their dependencies look like this:
image

IMO, this chaos may boil down to three reasons:
1. there are some useless nested types, like `KvWatcherInner` and `KvStoreBackend`. 
2. `KvStore` owns a `KvWatcher` but doesn't use it.
3. Some structures depend on a concrete type instead of an abstract interface.
 
I plan to refactor the dependencies. The dependencies after changed should look like this: 
## After changed
image

Phoenix500526:
refactor kv server dependencies, close #250, #259
This pr is based on #245 

Viewing #250 , #259 for details.

1. Merge `KvWatcherInner` and `KvWatcher` into a `KvWatcher`. All the things a `KvWatcherInner` does is spawn a watcher task and construct a `KvWatcherInner` object. A function `watcher` is better suited to do such a thing than a concrete structure `KvWatcher` in that a function is stateless, making the dependency graph clearer.
2. Merge `KvStoreBackend` and `KvStore` together. The reason why we need a `KvStoreBackend` is that the `KvStore` need to keep a `KvWatcher` as its member variable. But it's unnecessary, the `KvStore` never uses the `KvWatcher`. 
3. Place the `KvWatcher` constructed logic into XlineServer.
4. Make some data structures depend on interfaces rather than concrete data types.

No


curp/src/server/raw_curp/log.rs
```diff
@@ -424,7 +424,7 @@ mod tests {
     fn get_from_should_success() {
         let log_entry = LogEntry::new(0, 0, Arc::new(TestCommand::default()));
         let log_entry_size = serialized_size(&log_entry).unwrap();
-
+        // println!("log_entry_size = {log_entry_size}");
```
themanforfree, line 427: Remove this

Phoenix500526: Fixed

PR #247, opened 2023-04-28
Fix/watch

### LINKED ISSUE #166: [Bug]: Watch Concurrent Bug
### Description about the bug

When we call `watch` in `kvwatcher.rs`, the following things will happen:

1. get events from the requested revision
2. add watcher to the map and start listening for events

However, these two steps are not atomic operation. So if a new event occur after 1 but before 2, then we will lose this event.

### Version

0.1.0 (Default)

### Relevant log output

```shell
Not yet produced.
```

### Code of Conduct

themanforfree:
Fix #166 , and fix the atomicity of index and db operations

### DIFF
--- a/xline/src/server/command.rs
+++ b/xline/src/server/command.rs
@@ -278,6 +278,7 @@ where
             res = sync_res;
         }
         self.persistent.flush_ops(ops)?;
+        self.kv_storage.mark_index_available(res.revision());
         self.id_barrier.trigger(cmd.id());
         self.index_barrier.trigger(index);
         Ok(res)
--- a/xline/src/server/kv_server.rs
+++ b/xline/src/server/kv_server.rs
@@ -198,12 +198,18 @@ where
     }
 
     /// Validate range request before handle
-    fn check_range_request(req: &RangeRequest) -> Result<(), tonic::Status> {
+    fn check_range_request(req: &RangeRequest, current_revision: i64) -> Result<(), tonic::Status> {
         if req.key.is_empty() {
             return Err(tonic::Status::invalid_argument("key is not provided"));
         }
         if !SortOrder::is_valid(req.sort_order) || !SortTarget::is_valid(req.sort_target) {
             return Err(tonic::Status::invalid_argument("invalid sort option"));
+        }
+        if req.revision > current_revision {
+            return Err(tonic::Status::invalid_argument(format!(
+                "required revision {} is higher than current revision {}",
+                req.revision, current_revision
+            )));
         }
 
         Ok(())
@@ -234,7 +240,7 @@ where
     }
 
     /// Validate txn request before handle
-    fn check_txn_request(req: &TxnRequest) -> Result<(), tonic::Status> {
+    fn check_txn_request(req: &TxnRequest, current_revision: i64) -> Result<(), tonic::Status> {
         let opc = req
             .compare
             .len()
@@ -253,10 +259,10 @@ where
         for op in req.success.iter().chain(req.failure.iter()) {
             if let Some(ref request) = op.request {
                 match *request {
-                    Request::RequestRange(ref r) => Self::check_range_request(r),
+                    Request::RequestRange(ref r) => Self::check_range_request(r, current_revision),
                     Request::RequestPut(ref r) => Self::check_put_request(r),
                     Request::RequestDeleteRange(ref r) => Self::check_delete_range_request(r),
-                    Request::RequestTxn(ref r) => Self::check_txn_request(r),
+                    Request::RequestTxn(ref r) => Self::check_txn_request(r, current_revision),
                 }?;
             } else {
                 return Err(tonic::Status::invalid_argument("key not found"));
@@ -387,7 +393,7 @@ where
     ) -> Result<tonic::Response<RangeResponse>, tonic::Status> {
         debug!("Receive RangeRequest {:?}", request);
         let range_req = request.get_ref();
-        Self::check_range_request(range_req)?;
+        Self::check_range_request(range_req, self.kv_storage.revision())?;
         let is_serializable = range_req.serializable;
         let wrapper = match get_token(request.metadata()) {
             Some(token) => RequestWithToken::new_with_token(request.into_inner().into(), token),
@@ -463,7 +469,7 @@ where
         request: tonic::Request<TxnRequest>,
     ) -> Result<tonic::Response<TxnResponse>, tonic::Status> {
         debug!("Receive TxnRequest {:?}", request);
-        Self::check_txn_request(request.get_ref())?;
+        Self::check_txn_request(request.get_ref(), self.kv_storage.revision())?;
         let is_fast_path = false; // lock need revision of txn
         let (cmd_res, sync_res) = self.propose(request, is_fast_path).await?;
 
@@ -542,7 +548,38 @@ mod test {
             ],
             failure: vec![],
         };
-        let result = KvServer::<DB>::check_txn_request(&txn_req);
+        let result = KvServer::<DB>::check_txn_request(&txn_req, 0);
         assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_future_revision() {
+        let current_revision = 10;
+        let range_request = RangeRequest {
+            key: b"foo".to_vec(),
+            revision: 20,
+            ..Default::default()
+        };
+        let expected_err_message = tonic::Status::invalid_argument(format!(
+            "required revision {} is higher than current revision {}",
+            range_request.revision, current_revision
+        ))
+        .to_string();
+        let message = KvServer::<DB>::check_range_request(&range_request, current_revision)
+            .unwrap_err()
+            .to_string();
+        assert_eq!(message, expected_err_message);
+
+        let txn_req = TxnRequest {
+            compare: vec![],
+            success: vec![RequestOp {
+                request: Some(Request::RequestRange(range_request)),
+            }],
+            failure: vec![],
+        };
+        let message = KvServer::<DB>::check_txn_request(&txn_req, current_revision)
+            .unwrap_err()
+            .to_string();
+        assert_eq!(message, expected_err_message);
     }
 }
--- a/xline/src/server/watch_server.rs
+++ b/xline/src/server/watch_server.rs
@@ -7,6 +7,7 @@ use tracing::{debug, warn};
 
 use super::command::KeyRange;
 use crate::{
+    header_gen::HeaderGenerator,
     rpc::{
         RequestUnion, ResponseHeader, Watch, WatchCancelRequest, WatchCreateRequest, WatchRequest,
         WatchResponse,
@@ -30,6 +31,8 @@ where
     watcher: Arc<KvWatcher<S>>,
     /// Watch ID generator
     next_id_gen: Arc<WatchIdGenerator>,
+    /// Header Generator
+    header_gen: Arc<HeaderGenerator>,
 }
 
 impl<S> WatchServer<S>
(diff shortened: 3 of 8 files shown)


xline/src/storage/kv_store.rs
```diff
@@ -436,14 +442,15 @@ where
     }
 
     /// Get `KeyValue` start from a revision and convert to `Event`
-    pub(crate) fn get_event_from_revision(
+    pub(crate) fn get_events_from_revision(
         &self,
         key_range: KeyRange,
         revision: i64,
-    ) -> Result<Vec<Event>, ExecuteError> {
+    ) -> Result<(i64, Vec<Event>), ExecuteError> {
```
rogercloud, line 449: Please comment the meaning of the return value as it's complicated.

xline/src/server/watch_server.rs
```diff
@@ -49,9 +49,9 @@ where
         ST: Stream<Item = Result<WatchRequest, tonic::Status>> + Unpin,
         W: KvWatcherOps,
     {
-        let (event_tx, event_rx) = mpsc::channel(CHANNEL_SIZE);
+        let (event_tx, mut event_rx) = mpsc::channel(CHANNEL_SIZE);
         let (stop_tx, stop_rx) = flume::bounded(0);
```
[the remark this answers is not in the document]
themanforfree: Fixed

xline/src/storage/kvwatcher.rs
```diff
@@ -244,33 +260,61 @@ where
...
+    /// Mark as syncing done and notify
+    fn sync_done(&self) {
+        self.syncing.store(false, Ordering::SeqCst);
+        self.syncing_notify.notify(1);
+    }
+
+    /// If current state is syncing, wait until it is done
+    async fn wait_syncing(&self) {
+        while self.is_syncing() {
+            self.syncing_notify.listen().await;
```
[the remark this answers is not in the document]
themanforfree: Fixed

xline/src/storage/kv_store.rs
themanforfree: Fixed

xline/src/storage/revision.rs
```diff
@@ -11,6 +11,9 @@ pub(crate) struct KeyRevision {
     pub(crate) mod_revision: i64,
     /// Sub revision in one transaction
     pub(crate) sub_revision: i64,
+    /// Available flag
+    // true means this revision is written to storage
```
Phoenix500526, line 15: IMO, using the present perfect tense here would be more straightforward, like "/// true means this revision has been already written to storage".

xline/src/storage/index.rs
```diff
@@ -33,7 +45,7 @@ impl Index {
     fn get_revision(revs: &[KeyRevision], revision: i64) -> Option<Revision> {
         // TODO: handle future revision
```
Phoenix500526, line 46: Shall we handle this TODO in this pr?

xline/src/storage/index.rs
```diff
@@ -91,36 +131,42 @@ pub(crate) trait IndexOperate {
...
         match RangeType::get_range_type(key, range_end) {
-            RangeType::OneKey => index
+            RangeType::OneKey => inner
+                .index
                 .get(key)
                 .map(|revs| Self::filter_revision(revs, revision))
                 .unwrap_or_default(),
-            RangeType::AllKeys => index
+            RangeType::AllKeys => inner
+                .index
```
Phoenix500526, line 164: The test case `test_get` only covers the `RangeType::OneKey` match arm, please enrich this test case to cover all match arms.

xline/src/storage/index.rs
```diff
@@ -49,6 +61,34 @@ impl Index {
...
+        sub_revision: i64,
+    ) -> Option<(Revision, Revision)> {
+        let last_rev = Self::get_revision(revs, 0)?;
+        let del_rev = KeyRevision::new_deletion(revision, sub_revision);
+        revs.push(del_rev);
+        Some((last_rev, del_rev.as_revision()))
+    }
+
+    /// Mark the `KeyRevision` as available
+    pub(crate) fn mark_available(&self, revision: i64) {
```
Phoenix500526, line 78: The `mark_available` is only being used in the `kv_store`, so there is no point to describe it as crate visibility, super is enough.

xline/src/storage/index.rs
```diff
@@ -134,45 +180,53 @@ impl IndexOperate for Index {
...
+                    .filter_map(|revs| Self::gen_del_revision(revs, revision, sub_revision))
+                    .collect();
+                let keys = if pairs.is_empty() {
+                    vec![]
+                } else {
+                    vec![key.to_vec()]
+                };
+                (pairs, keys)
+            }
+            RangeType::AllKeys => {
```
Phoenix500526, line 200: Same as the `get_from_rev`. Plz enrich the test case `test_delete` to cover the `RangeType::AllKeys` and `RangeType::Range` match arms.

xline/src/server/command.rs
```diff
@@ -278,6 +278,7 @@ where
             res = sync_res;
         }
         self.persistent.flush_ops(ops)?;
+        self.kv_storage.mark_index_available(res.revision());
```
Phoenix500526, line 281: The `mark_index_available` will panic. Shall we deal with index error in this way? WDYT? @rogercloud

xline/src/storage/kv_store.rs
```diff
@@ -937,6 +944,33 @@ mod test {
...
+                    );
+                    let (_res, ops) = store.after_sync(&req).await.unwrap();
+                    db.flush_ops(ops).unwrap();
+                }
+            }
+        });
+        tokio::time::sleep(std::time::Duration::from_micros(500)).await;
+        let revs = store.index.get_from_rev(b"foo", b"", 1);
+        let kvs = store.get_values(&revs).unwrap();
+        assert_eq!(kvs.len(), revs.len());
```
Phoenix500526, line 971: Please add some concrete diagnostic messages for this assertion. Additionally, I don't get the point why this test can prove that the kv store process is atomic.

xline/src/storage/revision.rs
themanforfree: Fixed

xline/src/storage/index.rs
themanforfree: Fixed

xline/src/storage/index.rs
themanforfree: Fixed

xline/src/storage/index.rs
themanforfree: Fixed

xline/src/storage/kv_store.rs
themanforfree: Fixed, If the kv store is not atomic, some kvs maybe insert to index, but not flush to db, so `revs.len()` will lager than `kvs.len()`

xline/src/storage/index.rs
themanforfree: Added

xline/src/storage/kv_store.rs
```diff
@@ -922,8 +928,9 @@ mod test {
         );
         let db = DB::open(&StorageConfig::Memory)?;
         let store = init_store(db).await?;
-        let (_ignore, ops) = store.after_sync(&txn_req).await?;
+        let (sync_res, ops) = store.after_sync(&txn_req).await?;
         store.db.flush_ops(ops)?;
```
rogercloud, line 932: If these two operations always happend togethor, put them in one function to force the "pair"?

xline/src/storage/kv_store.rs
```diff
@@ -937,6 +944,37 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 10)]
+    async fn test_kv_store_atomic() {
```
rogercloud, line 948: Atomic usually means we can't insert other operations between the continuous operations in the logic view. I don't think the test tells the store is "atomic".

xline/src/storage/kv_store.rs
themanforfree: they always happened together in this module's tests, but not else where, I added a function to reduce duplication

xline/src/storage/kv_store.rs
themanforfree: Renamed

xline/src/server/command.rs
themanforfree: we can't recover it, all keys in `unavailable_cache` will be inserted into `index` first and then into `unavailable_cache`, so it is impossible to read none here

Phoenix500526: Then why don't you use the `unreachable!` macro to replace the `panic!`?

themanforfree: Fixed

xline/src/storage/kv_store.rs
```diff
@@ -960,15 +994,23 @@ mod test {
...
             );
-            let _cmd_res = store.execute(&req)?;
-            let (_sync_res, ops) = store.after_sync(&req).await?;
-            store.db.flush_ops(ops)?;
+            sync_req(&store, &req).await?;
         }
         Ok(store)
     }
 
+    async fn sync_req(
```
rogercloud, line 1002: the inner operations are not "syncing" but, execute "after_sync" step and flush to persistent store. Shall we rename it "exe_as_and_flush"?

xline/src/storage/kvwatcher.rs
```diff
@@ -258,50 +268,50 @@ where
...
-                    .flatten()
-                    .collect();
-                for watcher in watchers {
-                    #[allow(clippy::indexing_slicing)]
-                    watcher_events
-                        .entry(watcher)
-                        .or_default()
-                        .push(event.clone());
+    fn handle_kv_updates(&self, (revision, all_events): (i64, Vec<Event>)) {
+        let watcher_map_r = self.watcher_map.read();
```
rogercloud, line 279: Why do we remove the "map_read" function call?

xline/src/storage/kvwatcher.rs
```diff
@@ -332,3 +342,84 @@ impl WatchEvent {
...
+        }
+
+        assert_eq!(map.len(), 100);
+        for (k, count) in map {
+            assert_eq!(count, 1, "key {k} should be notified once");
+        }
+        handle.abort();
+    }
+
+    fn init_empty_store() -> (Arc<KvStore<DB>>, Arc<DB>, Arc<KvWatcher<DB>>) {
```
rogercloud, line 409: Shall we move the util function to the top of test mod?

themanforfree: Moved

xline/src/storage/kv_store.rs
themanforfree: Renamed

xline/src/storage/index.rs
```diff
@@ -49,6 +61,34 @@ impl Index {
...
+        sub_revision: i64,
+    ) -> Option<(Revision, Revision)> {
+        let last_rev = Self::get_revision(revs, 0)?;
+        let del_rev = KeyRevision::new_deletion(revision, sub_revision);
+        revs.push(del_rev);
+        Some((last_rev, del_rev.as_revision()))
+    }
+
+    /// Mark the `KeyRevision` as available
+    pub(super) fn mark_available(&self, revision: i64) {
```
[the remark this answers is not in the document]
themanforfree: Commented

xline/src/server/kv_server.rs
```diff
@@ -198,13 +198,19 @@ where
...
     /// Validate range request before handle
-    fn check_range_request(req: &RangeRequest) -> Result<(), tonic::Status> {
+    fn check_range_request(req: &RangeRequest, current_revision: i64) -> Result<(), tonic::Status> {
         if req.key.is_empty() {
             return Err(tonic::Status::invalid_argument("key is not provided"));
         }
         if !SortOrder::is_valid(req.sort_order) || !SortTarget::is_valid(req.sort_target) {
             return Err(tonic::Status::invalid_argument("invalid sort option"));
         }
+        if req.revision > current_revision {
```
Phoenix500526, line 208: Why do we need to check the revision here now when we didn't need to before?

themanforfree: it was a todo before
https://github.com/datenlord/Xline/pull/247#discussion_r1190788475

Phoenix500526 (approved):

rogercloud (approved):

Issue #259, opened 2023-05-06
[Bug]: Test cases get_from_should_success and recovery_from_log will fail sometime in CI


Phoenix500526:
### Description about the bug

As the title described, the test cases `get_from_should_success` and `recovery_from_log` will fail sometime.

These two cases are pasted below:
```Rust
    #[test]
    fn get_from_should_success() {
        let log_entry = LogEntry::new(0, 0, Arc::new(TestCommand::default()));
        let log_entry_size = serialized_size(&log_entry).unwrap();

        let (tx, _rx) = mpsc::unbounded_channel();
        let mut log =
            Log::<TestCommand>::new(tx, default_batch_max_size(), default_log_entries_cap());

        let _res = repeat(TestCommand::default())
            .take(10)
            .map(|cmd| log.push_cmd(1, Arc::new(cmd)).unwrap())
            .collect::<Vec<u64>>();

        set_batch_limit(&mut log, 3 * log_entry_size - 1);
        let bound_1 = log.get_range_by_batch(3);
        assert_eq!(
            bound_1,
            3..5,
            "batch_index = {:?}, batch = {}",
            log.batch_index,
            log.batch_limit
        );
        assert!(log.has_next_batch(8));
        assert!(!log.has_next_batch(9));
        ... // omit some test cases
    }

    #[test]
    fn recover_log_should_success() {
        let entry_size =
            serialized_size(&LogEntry::new(0, 0, Arc::new(TestCommand::default()))).unwrap();
        let entries = repeat(Arc::new(TestCommand::default()))
            .enumerate()
            .take(10)
            .map(|(idx, cmd)| LogEntry::new((idx + 1).numeric_cast(), 0, cmd))
            .collect::<Vec<LogEntry<TestCommand>>>();
        let (tx, _rx) = mpsc::unbounded_channel();
        let mut log =
            Log::<TestCommand>::new(tx, default_batch_max_size(), default_log_entries_cap());

        log.restore_entries(entries);
        assert_eq!(log.entries.len(), 10);
        assert_eq!(log.batch_index.len(), 11);
        assert_eq!(log.batch_index[0], 0);

        log.batch_index
            .iter()
            .enumerate()
            .for_each(|(idx, &size)| assert_eq!(size, entry_size * idx.numeric_cast::<u64>()));
    }
```

[...]

PR #260, opened 2023-05-06
Refactor/dependencies

Phoenix500526:
refactor kv server dependencies, close [#250], [#259] 
This pr is based on [#245]

Viewing #250, and #259 for more details. 

1. Merge `KvWatcherInner` and `KvWatcher` into a `KvWatcher`. All the things a `KvWatcherInner` does is spawn a watcher task and construct a `KvWatcherInner` object. A function watcher is better suited to do such a thing than a concrete structure `KvWatcher` in that a function is stateless, making the dependency graph clearer.
2. Merge `KvStoreBackend` and `KvStore` together. The reason why we need a `KvStoreBackend` is that the `KvStore` need to keep a `KvWatcher` as its member variable. But it's unnecessary, the `KvStore` never uses the `KvWatcher`.
3. Place the `KvWatcher` constructed logic into `XlineServer`.
4. Make some data structures depend on interfaces rather than concrete data types.

No

### DIFF
--- a/xline/src/storage/kvwatcher.rs
+++ b/xline/src/storage/kvwatcher.rs
@@ -183,21 +183,6 @@ impl WatcherMap {
     }
 }
 
-/// Boot up a watcher task and return an `Arc<KvWatcher<S>>`
-pub(crate) fn watcher<S: StorageApi>(
-    storage: Arc<KvStore<S>>,
-    mut kv_update_rx: mpsc::Receiver<(i64, Vec<Event>)>,
-) -> Arc<KvWatcher<S>> {
-    let kv_watcher = Arc::new(KvWatcher::new(storage));
-    let watcher = Arc::clone(&kv_watcher);
-    let _handle = tokio::spawn(async move {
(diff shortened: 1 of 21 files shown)


xline/src/storage/kvwatcher.rs
```diff
@@ -174,24 +164,19 @@ impl WatcherMap {
...
+) -> Arc<KvWatcher<S>> {
+    let kv_watcher = Arc::new(KvWatcher::new(storage));
+    let watcher = Arc::clone(&kv_watcher);
+    let _handle = tokio::spawn(async move {
+        while let Some(updates) = kv_update_rx.recv().await {
+            watcher.handle_kv_updates(updates).await;
+        }
+    });
+    kv_watcher
 }
```
themanforfree, line 180: Why not implement this method directly for `KvWatcher`

Phoenix500526: Done

themanforfree (dismissed):

themanforfree (approved):

PR #264, opened 2023-05-09
fix #263

### LINKED ISSUE #263: Refactor: improve the read and write logic for the RocksSnapshot
Currently, the `read` and `write` methods in `RocksSnapshot` look like this:
```Rust
impl RocksSnapshot {
   /// Read data from the snapshot
    async fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        if buf.is_empty() {
            return Ok(0);
        }

        if self.meta.is_current {
            let n = self.meta.data.read(buf).await?;
            if n == 0 {
                self.meta.is_current = false;
            } else {
                return Ok(n);
            }
        }

        while self.snap_file_idx < self.snap_files.len() {
            let f = if let Some(ref mut f) = self.current_file {
                f
            } else {
                let path = self.current_file_path(false);
                let reader = File::open(path).await?;
                self.current_file = Some(reader);
                self.current_file
                    .as_mut()
                    .unwrap_or_else(|| unreachable!("current_file must be `Some` here"))
            };
            let n = f.read(buf).await?;
            if n == 0 {
                let _ignore = self.current_file.take();
                self.snap_file_idx = self.snap_file_idx.overflow_add(1);
            } else {
                return Ok(n);
            }
        }
        Ok(0)
    }

    /// Write snapshot data
    async fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        if buf.is_empty() {
            return Ok(0);
        }

        let (mut next_buf, mut written_bytes) = (buf, 0);

        #[allow(clippy::indexing_slicing)] // length is checked when reading meta
        if self.meta.is_current {
            let Some(meta_len_slice) = next_buf.get(0..8) else {
                return Err(io::Error::new(
                    ErrorKind::UnexpectedEof,
                    "cannot read meta length from buffer",
                ));
            };
            let meta_len_bytes: [u8; 8] = meta_len_slice
                .try_into()
                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
            let meta_len = u64::from_le_bytes(meta_len_bytes);

            let Some(meta_slice) = next_buf.get(8..meta_len.overflow_add(8).numeric_cast()) else {
                return Err(io::Error::new(
                    ErrorKind::UnexpectedEof,
                    "cannot read meta from buffer",
                ));
            };
            let meta_bytes: Vec<u8> = meta_slice
                .try_into()
                .unwrap_or_else(|_e| unreachable!("infallible"));
            let meta = bincode::deserialize(&meta_bytes)
                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;

            self.apply_snap_meta(meta);
            let mut data = Vec::new();
            data.extend(meta_len_bytes);
            data.extend(meta_bytes);
            *self.meta.data.get_mut() = data;
            self.meta.is_current = false;
            written_bytes = written_bytes
                .overflow_add(meta_len.numeric_cast())
                .overflow_add(8);
            next_buf = &next_buf[meta_len.overflow_add(8).numeric_cast()..];
        }

        // the snap_file_idx has checked with while's pattern
        // written_len is calculated by next_buf.len() so it must be less than next_buf's length
        #[allow(clippy::indexing_slicing)]
        while self.snap_file_idx < self.snap_files.len() {
            let snap_file = &mut self.snap_files[self.snap_file_idx];
            if snap_file.size == 0 {
                return Err(io::Error::new(
                    ErrorKind::Other,
                    format!("snap file {} size is 0", snap_file.filename),
                ));
            }
            let left = snap_file.remain_size().numeric_cast();
            let (write_len, switch, finished) = match next_buf.len().cmp(&left) {
                Ordering::Greater => (left, true, false),
                Ordering::Equal => (left, true, true),
                Ordering::Less => (next_buf.len(), false, true),
            };
            snap_file.written_size = snap_file
                .written_size
                .overflow_add(write_len.numeric_cast());
            written_bytes = written_bytes.overflow_add(write_len);

            let buffer = &next_buf[0..write_len];

            let f = if let Some(ref mut f) = self.current_file {
                f
            } else {
                let path = self.current_file_path(true);
                let writer = File::create(path).await?;
                self.current_file = Some(writer);
                self.current_file
                    .as_mut()
                    .unwrap_or_else(|| unreachable!("current_file must be `Some` here"))
            };
            f.write_all(buffer).await?;

            if switch {
                next_buf = &next_buf[write_len..];
                let old = self.current_file.take();
                if let Some(mut old_f) = old {
                    old_f.flush().await?;
                    let path = self.current_file_path(false);
                    let tmp_path = self.current_file_path(true);
                    fs::rename(tmp_path, path)?;
                }
                self.snap_file_idx = self.snap_file_idx.overflow_add(1);
            }
            if finished {
                break;
            }
        }
        Ok(written_bytes)
    }
}
```
In this implementation, we should maintain the slice index manually. IMO, we can use the [bytes](https://docs.rs/bytes/1.4.0/bytes/) crate to refine it.

bsbds:
fix #263 

    improve the read and write logic for the RocksSnapshot

    change of SnapshotApi in read_exact and write_all funtions and all associated calls on these funtions

    read_exact and write_all funtions now use bytes::{Bytes, BytesMut} as parameter types

### DIFF
--- a/curp/src/rpc/connect.rs
+++ b/curp/src/rpc/connect.rs
@@ -292,7 +292,7 @@ fn install_snapshot_stream(
             let len: u64 =
                 std::cmp::min(snapshot.size() - offset, SNAPSHOT_CHUNK_SIZE).numeric_cast();
             let mut data = BytesMut::with_capacity(len.numeric_cast());
-            if let Err(e) = snapshot.read_exact(&mut data).await {
+            if let Err(e) = snapshot.read_buf_exact(&mut data).await {
                 error!("read snapshot error, {e}");
                 break;
             }
--- a/curp/src/rpc/connect.rs
+++ b/curp/src/rpc/connect.rs
@@ -2,6 +2,7 @@ use std::{collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
 
 use async_stream::stream;
 use async_trait::async_trait;
+use bytes::BytesMut;
 use clippy_utilities::NumericCast;
 use engine::SnapshotApi;
 use futures::Stream;
@@ -290,7 +291,7 @@ fn install_snapshot_stream(
         while offset < snapshot.size() {
             let len: u64 =
                 std::cmp::min(snapshot.size() - offset, SNAPSHOT_CHUNK_SIZE).numeric_cast();
-            let mut data = vec![0; len.numeric_cast()];
+            let mut data = BytesMut::with_capacity(len.numeric_cast());
             if let Err(e) = snapshot.read_exact(&mut data).await {
                 error!("read snapshot error, {e}");
                 break;
@@ -301,7 +302,7 @@ fn install_snapshot_stream(
                 last_included_index: meta.last_included_index,
                 last_included_term: meta.last_included_term,
                 offset,
-                data,
+                data: data.freeze(),
                 done: (offset + len) == snapshot.size(),
             };
 
@@ -329,6 +330,7 @@ impl Connect {
 
 #[cfg(test)]
 mod tests {
+    use bytes::Bytes;
     use engine::{EngineType, Snapshot as EngineSnapshot};
     use futures::{pin_mut, StreamExt};
     use tracing_test::traced_test;
@@ -342,7 +344,7 @@ mod tests {
         const SNAPSHOT_SIZE: u64 = 200 * 1024;
         let mut snapshot = EngineSnapshot::new_for_receiving(EngineType::Memory).unwrap();
         snapshot
-            .write_all(&mut vec![1; SNAPSHOT_SIZE.numeric_cast()])
+            .write_all(Bytes::from(vec![1; SNAPSHOT_SIZE.numeric_cast()]))
             .await
             .unwrap();
         let stream = install_snapshot_stream(
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -222,17 +222,15 @@ impl<C: 'static + Command> CurpNode<C> {
             ) {
                 return Ok(InstallSnapshotResponse::new(self.curp.term()));
             }
-            snapshot
-                .write_all(req.data.as_slice())
-                .await
-                .map_err(|err| {
-                    error!("can't write snapshot data, {err:?}");
-                    err
-                })?;
+            let req_data_len = req.data.len().numeric_cast::<u64>();
+            snapshot.write_all(req.data).await.map_err(|err| {
+                error!("can't write snapshot data, {err:?}");
+                err
+            })?;
             if req.done {
                 debug_assert_eq!(
                     snapshot.size(),
-                    req.offset + req.data.len().numeric_cast::<u64>(),
+                    req.offset + req_data_len,
                     "snapshot corrupted"
                 );
                 let meta = SnapshotMeta {
--- a/engine/src/proxy.rs
+++ b/engine/src/proxy.rs
@@ -1,5 +1,7 @@
 use std::path::{Path, PathBuf};
 
+use bytes::{Bytes, BytesMut};
+
 use crate::{
     error::EngineError,
     memory_engine::{MemoryEngine, MemorySnapshot},
@@ -155,7 +157,7 @@ impl SnapshotApi for Snapshot {
     }
 
     #[inline]
-    async fn read_exact(&mut self, buf: &mut [u8]) -> std::io::Result<()> {
+    async fn read_exact(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
         match *self {
             Snapshot::Memory(ref mut s) => s.read_exact(buf).await,
             Snapshot::Rocks(ref mut s) => s.read_exact(buf).await,
@@ -163,7 +165,7 @@ impl SnapshotApi for Snapshot {
     }
 
     #[inline]
-    async fn write_all(&mut self, buf: &[u8]) -> std::io::Result<()> {
+    async fn write_all(&mut self, buf: Bytes) -> std::io::Result<()> {
         match *self {
             Snapshot::Memory(ref mut s) => s.write_all(buf).await,
             Snapshot::Rocks(ref mut s) => s.write_all(buf).await,
@@ -335,12 +337,12 @@ mod test {
             let put = WriteOperation::new_put("kv", "key2".into(), "value2".into());
             assert!(engine.write_batch(vec![put], false).is_ok());
 
-            let mut buf = vec![0u8; snapshot.size().numeric_cast()];
+            let mut buf = BytesMut::with_capacity(snapshot.size().numeric_cast());
             snapshot.read_exact(&mut buf).await.unwrap();
 
             buf.extend([0u8; 100]); // add some padding, will be ignored when receiving
 
-            received_snapshot.write_all(&buf).await.unwrap();
+            received_snapshot.write_all(buf.freeze()).await.unwrap();
 
             assert!(recover_engine
                 .apply_snapshot(received_snapshot, &TESTTABLES)
--- a/engine/src/proxy.rs
+++ b/engine/src/proxy.rs
@@ -157,10 +157,18 @@ impl SnapshotApi for Snapshot {
     }
 
     #[inline]
-    async fn read_exact(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
+    async fn read_buf(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
         match *self {
-            Snapshot::Memory(ref mut s) => s.read_exact(buf).await,
-            Snapshot::Rocks(ref mut s) => s.read_exact(buf).await,
+            Snapshot::Memory(ref mut s) => s.read_buf(buf).await,
+            Snapshot::Rocks(ref mut s) => s.read_buf(buf).await,
+        }
+    }
+
+    #[inline]
+    async fn read_buf_exact(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
+        match *self {
+            Snapshot::Memory(ref mut s) => s.read_buf_exact(buf).await,
+            Snapshot::Rocks(ref mut s) => s.read_buf_exact(buf).await,
         }
     }
 
@@ -338,7 +346,7 @@ mod test {
             assert!(engine.write_batch(vec![put], false).is_ok());
 
             let mut buf = BytesMut::with_capacity(snapshot.size().numeric_cast());
-            snapshot.read_exact(&mut buf).await.unwrap();
+            snapshot.read_buf_exact(&mut buf).await.unwrap();
 
             buf.extend([0u8; 100]); // add some padding, will be ignored when receiving
 
--- a/engine/src/rocksdb_engine.rs
+++ b/engine/src/rocksdb_engine.rs
@@ -386,10 +386,6 @@ impl RocksSnapshot {
 
     /// Read data from the snapshot
     async fn read(&mut self, buf: &mut BytesMut) -> io::Result<usize> {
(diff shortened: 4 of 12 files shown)


engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -340,39 +335,41 @@ impl SnapshotApi for RocksSnapshot {
     }
 
     #[inline]
-    async fn read_exact(&mut self, mut buf: &mut [u8]) -> std::io::Result<()> {
-        while !buf.is_empty() {
+    async fn read_exact(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
+        let origin_cap = buf.capacity();
+        while buf.len() < buf.capacity() {
             // the return value of read function is not larger than the input buf's length.
             #[allow(clippy::indexing_slicing)]
```
Phoenix500526, line 342: Plz remove this clippy lint

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -340,39 +335,41 @@ impl SnapshotApi for RocksSnapshot {
...
             match self.read(buf).await {
                 Ok(0) => break,
-                Ok(n) => {
-                    let tmp = buf;
-                    buf = &mut tmp[n..];
-                }
+                Ok(_) => continue,
```
Phoenix500526, line 345:
```suggestion
                Ok(n) => if 0 == n { break;},
```

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -340,39 +330,43 @@ impl SnapshotApi for RocksSnapshot {
...
         }
-        if buf.is_empty() {
+        if buf.len() == buf.capacity() {
             Ok(())
         } else {
             Err(std::io::Error::new(
                 ErrorKind::UnexpectedEof,
-                "failed to fill whole buffer",
+                format!(
+                    "failed to fill whole buffer: {} {} {}",
```
Phoenix500526, line 353: The error message should be more meaningful.

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -340,39 +330,43 @@ impl SnapshotApi for RocksSnapshot {
...
             ))
         }
     }
 
     #[inline]
-    async fn write_all(&mut self, mut buf: &[u8]) -> std::io::Result<()> {
+    async fn write_all(&mut self, mut buf: Bytes) -> std::io::Result<()> {
         let mut written_size = 0;
         // the return value of write function is not larger than the input buf's length.
         #[allow(clippy::indexing_slicing)]
```
Phoenix500526, line 366: There is no point to allow `clippy::indexing_slicing`

engine/src/snapshot_api/rocks_snapshot.rs
```diff
@@ -340,39 +330,43 @@ impl SnapshotApi for RocksSnapshot {
...
         while !buf.is_empty() {
```
Phoenix500526, line 367: You can use `has_remaining` to replace `buf.is_empty` and we don't need to maintain a `written_size` variable to track how many bytes we read.

engine/src/snapshot_api/memory_snapshot.rs
```diff
@@ -34,13 +36,13 @@ impl SnapshotApi for MemorySnapshot {
     }
 
     #[inline]
-    async fn read_exact(&mut self, buf: &mut [u8]) -> std::io::Result<()> {
-        self.data.read_exact(buf).await.map(drop)
+    async fn read_exact(&mut self, buf: &mut BytesMut) -> std::io::Result<()> {
+        read_buf(&mut self.data, buf).await.map(drop)
```
themanforfree, line 40: Why use `read_buf` instead `read_exact`

bsbds: the `read_exact` function requires that the `buf` to be initialized. Howerver, `BytesMut` instances are not typically initialized, which would compel us to maintain an index if they were.

engine/src/rocksdb_engine.rs
```diff
@@ -386,13 +385,13 @@ impl RocksSnapshot {
     }
 
     /// Read data from the snapshot
-    async fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        if buf.is_empty() {
+    async fn read(&mut self, buf: &mut BytesMut) -> io::Result<usize> {
```
Phoenix500526, line 388: The `read` method doesn't depend on any specific methods of `BytesMut`, so, IMO, we can make the second parameter more general. How about `impl BufMut`?

engine/src/rocksdb_engine.rs
```diff
@@ -423,52 +422,54 @@ impl RocksSnapshot {
...
+                ));
+            };
+            let meta_bytes = buf.split_to(meta_len.numeric_cast());
             let meta = bincode::deserialize(&meta_bytes)
                 .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
 
             self.apply_snap_meta(meta);
             let mut data = Vec::new();
-            data.extend(meta_len_bytes);
+            data.extend(meta_len.to_le_bytes());
```
Phoenix500526, line 447: Shall we change the `self.meta.data` from `Cursor<Vec<u8>>` to `Cursor<Bytes>`? So we can use `chain` to link `meta_len` and `meta_bytes` as a new `Bytes` without copying meta_bytes.

bsbds:
Current implementation in `chunk_mut` for `BytesMut` only expand the capacity of the `BytesMut` when it already full. So it won't automatically expand the capacity because either the `read_buf` function only called once(unless the buf's capacity is zero), or the condition that the length is smaller than the capacity have already been checked. @themanforfree 
```rust, no_run
    #[inline]
    fn chunk_mut(&mut self) -> &mut UninitSlice {
        if self.capacity() == self.len() {
            self.reserve(64);
        }
        UninitSlice::from_slice(self.spare_capacity_mut())
    }
```
So we do rely on this specific implementation of `BytesMut`, if we use `impl BufMut` in the parameter, the `chunk_mut` may grow capacity before it gets full. @Phoenix500526

Phoenix500526:
The `chunk_mut` method implemented for `Vec<u8>` doesn't look very different from the one that `BytesMut` implemented.  FYI.
```Rust
unsafe impl BufMut for Vec<u8> {
    #[inline]
    fn chunk_mut(&mut self) -> &mut UninitSlice {
        if self.capacity() == self.len() {
            self.reserve(64); // Grow the vec
        }

        let cap = self.capacity();
        let len = self.len();

        let ptr = self.as_mut_ptr();
        unsafe { &mut UninitSlice::from_raw_parts_mut(ptr, cap)[len..] }
    }
}
```
A `Vec<u8>` will also grow its capacity when necessary. And a `&mut [u8]` also implements the `BufMut` trait, but we don't need to bother this issue.

curp/src/rpc/connect.rs
```diff
@@ -290,7 +291,7 @@ fn install_snapshot_stream(
         while offset < snapshot.size() {
             let len: u64 =
                 std::cmp::min(snapshot.size() - offset, SNAPSHOT_CHUNK_SIZE).numeric_cast();
-            let mut data = vec![0; len.numeric_cast()];
+            let mut data = BytesMut::with_capacity(len.numeric_cast());
             if let Err(e) = snapshot.read_exact(&mut data).await {
```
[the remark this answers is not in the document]
rogercloud: An optional way is to check the length after the read. @Phoenix500526

curp/src/server/curp_node.rs
```diff
@@ -222,17 +222,15 @@ impl<C: 'static + Command> CurpNode<C> {
...
                 return Ok(InstallSnapshotResponse::new(self.curp.term()));
             }
-            snapshot
-                .write_all(req.data.as_slice())
-                .await
-                .map_err(|err| {
-                    error!("can't write snapshot data, {err:?}");
-                    err
-                })?;
+            let req_data_len = req.data.len().numeric_cast::<u64>();
```
rogercloud, line 225: will it be used many times? Or does it make the code more understandable? If not, adding a local variable is a bad idea.

engine/src/proxy.rs
```diff
@@ -335,12 +337,12 @@ mod test {
...
             assert!(engine.write_batch(vec![put], false).is_ok());
 
-            let mut buf = vec![0u8; snapshot.size().numeric_cast()];
+            let mut buf = BytesMut::with_capacity(snapshot.size().numeric_cast());
             snapshot.read_exact(&mut buf).await.unwrap();
 
             buf.extend([0u8; 100]); // add some padding, will be ignored when receiving
 
-            received_snapshot.write_all(&buf).await.unwrap();
+            received_snapshot.write_all(buf.freeze()).await.unwrap();
```
rogercloud, line 345: Please refer to https://github.com/tokio-rs/bytes/issues/401 for a better understanding of `freeze` function. It's not zero-cost. We should reconsider whether to use `bytes` or not.

bsbds: There's a commit that fix this problem https://github.com/tokio-rs/bytes/commit/05e9d5cab95e5ce1d57dd498a9567639de50f841. Also since we already filled the buf, calls to `Vec::shrink_to_fit()` will not cause additional allocation.

xline/src/client/restore.rs
```diff
@@ -20,12 +21,12 @@ pub async fn restore(
     let mut snapshot_f = tokio::fs::File::open(snapshot_path).await?;
     let tmp_path = format!("/tmp/snapshot-{}", uuid::Uuid::new_v4());
     let mut rocks_snapshot = Snapshot::new_for_receiving(EngineType::Rocks((&tmp_path).into()))?;
-    let mut buf = vec![0; MAINTENANCE_SNAPSHOT_CHUNK_SIZE.cast()];
-    while let Ok(n) = snapshot_f.read(&mut buf).await {
+    let mut buf = BytesMut::with_capacity(MAINTENANCE_SNAPSHOT_CHUNK_SIZE.cast());
+    while let Ok(n) = read_buf(&mut snapshot_f, &mut buf).await {
```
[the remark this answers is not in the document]
bsbds: I believe pre-allocation will be faster as we already know the size.

bsbds:
Since we refactored the code to use bytes crate, the `BytesMut` objects are typically used uninitialized. Maybe we should consider change the method name from `read_exact` to `read_buf_exact` to match with std::io::Read trait method. https://doc.rust-lang.org/std/io/trait.Read.html#method.read_buf_exact

curp/src/server/curp_node.rs
bsbds: the `req.data` is moved when calling `write_all`, so the length must be record before the call.

engine/src/rocksdb_engine.rs
bsbds: fix this in https://github.com/datenlord/Xline/pull/264/commits/8626b05398cc93a43a67978e515212624ae93f71

curp/src/rpc/connect.rs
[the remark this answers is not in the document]
bsbds: `read_buf_exact` in https://github.com/datenlord/Xline/pull/264/commits/93316434913316a7242da316df7fd4e0e877d31c now ensures reading the same amount as the initial capacity

engine/src/rocksdb_engine.rs
bsbds: The `read_buf_mut` implementation rely on `len` and `capacity` funtions, which are not trait methods. So I think it's better keep using `BytesMut`.

Phoenix500526 (approved):

rogercloud (approved):

PR #276, opened 2023-05-15
Refactor/issue-251

### LINKED ISSUE #251: Refactor: refactor the dependencies of lease server
Currently, there is a confusion of dependencies between several components in our lease server. Their dependencies look like this:
image

I plan to refactor our lease server. This refactoring can be split into two steps:
1. Remove the dependency from `LeaseStoreBackend` to `State`. The main duty of the `State` is to check whether the current node is a leader or not. This is a kind of business logic, and `LeaseStoreBackend` shouldn't be responsible for that while the `LeaseServer` should.
2. Merge the `LeaseStore` and the `LeaseStoreBackend` together. Their definitions look like this:
```Rust
/// Lease store
#[derive(Debug)]
pub(crate) struct LeaseStore<DB>
where
    DB: StorageApi,
{
    /// Lease store Backend
    inner: Arc<LeaseStoreBackend<DB>>,
}

/// Lease store inner
#[derive(Debug)]
struct LeaseStoreBackend<DB>
where
    DB: StorageApi,
{
    /// lease collection
    lease_collection: Arc<LeaseCollection>,
    /// Db to store lease
    db: Arc<DB>,
    /// Key to revision index
    index: Arc<Index>,
    /// Current node is leader or not
    state: Arc<State>,
    /// Revision
    revision: Arc<RevisionNumber>,
    /// Header generator
    header_gen: Arc<HeaderGenerator>,
    /// KV update sender
}
```
There is no point to define a `LeaseStoreBackend` here. We can inline the `LeaseStoreBackend` into the `LeaseStore`. In doing so, we can make our dependencies in the lease_server crystal clear. After refactoring, the dependencies should look like this:
image

Phoenix500526:
Refactor lease server, based on pr #269. Closes #251 

Read #251 for more details. 

1. Inline `LeaseStoreBackend` into `LeaseStore`
2. Add a `RoleChange` trait and a new `State`, Read issue #251 for more details.
3. Add an integration test to cover

No


curp/src/server/curp_node.rs
```diff
@@ -405,12 +406,14 @@ impl<C: 'static + Command> CurpNode<C> {
...
     /// Create a new server instance
     #[inline]
+    #[allow(clippy::too_many_arguments)]
     pub(super) async fn new<CE: CommandExecutor<C> + 'static>(
         id: ServerId,
         is_leader: bool,
         others: HashMap<ServerId, String>,
         cmd_executor: Arc<CE>,
         snapshot_allocator: impl SnapshotAllocator + 'static,
+        role_change: impl RoleChange + 'static,
```
rogercloud, line 416: A structure to wrap all the arguments?

curp/src/server/mod.rs
```diff
@@ -136,12 +137,14 @@ impl<C: Command + 'static> Rpc<C> {
...
     /// Panic if storage creation failed
     #[inline]
+    #[allow(clippy::too_many_arguments)]
     pub async fn new<CE: CommandExecutor<C> + 'static>(
         id: ServerId,
         is_leader: bool,
         others: HashMap<ServerId, String>,
         executor: CE,
         snapshot_allocator: impl SnapshotAllocator + 'static,
+        role_change: impl RoleChange + 'static,
```
rogercloud, line 147: Same as above.

curp/src/server/raw_curp/mod.rs
```diff
@@ -580,6 +583,7 @@ impl<C: 'static + Command> RawCurp<C> {
         cmd_tx: Arc<dyn CEEventTxApi<C>>,
         sync_events: HashMap<ServerId, Arc<Event>>,
         log_tx: mpsc::UnboundedSender<LogEntry<C>>,
+        role_change: impl RoleChange + 'static,
```
rogercloud, line 586: Same as above.

curp/src/server/raw_curp/tests.rs
```diff
@@ -38,6 +39,14 @@ impl<C: 'static + Command> RawCurp<C> {
             .iter()
             .map(|id| (id.clone(), Arc::new(Event::new())))
             .collect();
+
+        let role_change = {
+            let mut role_change = MockRoleChange::default();
+            role_change.expect_on_election_win().returning(|| {});
```
rogercloud, line 45: Shall we add some notifier or channel inside the closure so that we can get informed when it's called?

curp/tests/common/curp_group.rs
```diff
@@ -106,7 +108,7 @@ pub struct CrashedCurpNode {
 }
 
 pub struct CurpGroup {
-    pub nodes: HashMap<ServerId, CurpNode>,
+    pub nodes: HashMap<ServerId, (CurpNode, Arc<TestRoleChangeInner>)>,
```
rogercloud, line 111: Would it be better to give it a named structure instead of a tuple?

curp/src/server/raw_curp/mod.rs
```diff
@@ -147,6 +148,8 @@ struct Context<C: Command> {
     sync_events: HashMap<ServerId, Arc<Event>>,
     /// Become leader event
     leader_event: Arc<Event>,
+    /// Leader change callback
+    role_change: Box<dyn RoleChange>,
```
rogercloud, line 152: Is static generic possible?

xline/src/server/lease_server.rs
```diff
@@ -41,10 +40,10 @@ where
     client: Arc<Client<Command>>,
     /// Server name
     name: String,
-    /// State of current node
-    state: Arc<State>,
     /// Id generator
     id_gen: Arc<IdGenerator>,
+    /// Address of all members
+    all_members: HashMap<String, String>,
```
rogercloud, line 46: I believe it's not the only server that knows the address of all members. Shall we share the info?

xline/src/server/xline_server.rs
```diff
@@ -285,29 +281,9 @@ where
...
-                    lease_storage.demote();
-                }
-            }
-        }
-    }
-
     /// Init `KvServer`, `LockServer`, `LeaseServer`, `WatchServer` and `CurpServer`
     /// for the Xline Server.
-    #[allow(clippy::type_complexity)] // it is easy to read
+    #[allow(clippy::type_complexity, clippy::too_many_lines)] // it is easy to read
```
rogercloud, line 286: Break the function into smaller ones.

xline/src/server/lease_server.rs
```diff
@@ -346,4 +298,72 @@ where
...
+    }
+
+    /// LeaseTimeToLive retrieves lease information.
+    async fn lease_time_to_live(
+        &self,
+        request: tonic::Request<LeaseTimeToLiveRequest>,
+    ) -> Result<tonic::Response<LeaseTimeToLiveResponse>, tonic::Status> {
+        debug!("Receive LeaseTimeToLiveRequest {:?}", request);
+        loop {
+            #[allow(clippy::redundant_else)]
```
rogercloud, line 330: Why do we need the redundent else?

xline/src/server/lease_server.rs
```diff
@@ -346,4 +298,72 @@ where
...
+                    id: time_to_live_req.id,
+                    ttl: lease.remaining().as_secs().cast(),
+                    granted_ttl: lease.ttl().as_secs().cast(),
+                    keys,
+                };
+                return Ok(tonic::Response::new(res));
+            } else {
+                let leader_id = self.client.get_leader_id().await;
+                let leader_addr = self.all_members.get(&leader_id).unwrap_or_else(|| {
+                    panic!(
```
rogercloud, line 353: unreachable is better?

curp/src/server/raw_curp/mod.rs
Phoenix500526: The `impl Trait` is only allowed in function and inherent method return types, not in path.

xline/src/server/lease_server.rs
Phoenix500526: Fixed

xline/src/server/lease_server.rs
Phoenix500526: Fixed

curp/tests/server_election.rs
```diff
@@ -4,26 +4,51 @@ use madsim::time::sleep;
...
+}
+
 // Election
 #[tokio::test]
 async fn election() {
     init_logger();
 
     let group = CurpGroup::new(5).await;
     let leader0 = group.get_leader().await.0;
+    check_role_state(&group, 5, &leader0);
```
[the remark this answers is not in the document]
Phoenix500526: That's a duplicated function in that the CurpClient provides a `get_leader_id` method. Therefore, you can get the leader by a curpclient driectly.

xline/src/server/xline_server.rs
Phoenix500526: I'll do it in a new pr

curp/src/server/raw_curp/tests.rs
Phoenix500526: At the moment we don't have such a requirement and we can wait until we really need it to add it.

curp/src/server/raw_curp/mod.rs
Phoenix500526: Done

xline/src/server/lease_server.rs
Phoenix500526: Whether a server need to know the address of all members or not depends on whether it needs to forward its requests to the leader. And currently, only the lease server depends on it.

curp/tests/common/curp_group.rs
Phoenix500526: Done

curp/src/server/curp_node.rs
Phoenix500526: These arguments are provided by the XlineServer. Using a structure to warp them all doesn't help a lot. There are two reasons:
1. there is not engouh semantic commonality between these arguments and it's hard to find a suitable tname for such a data structure.
2. it increase the burden on XlineServer, which still has to provide these parameters and also need to construct such a data structure via them.

curp/src/server/mod.rs
Phoenix500526: Same as above

curp/src/lib.rs
```diff
@@ -165,6 +165,9 @@ pub mod error;
 /// The command to be executed
 pub mod cmd;
 
+/// The callback trait to be invoked when the curp role changes
+pub mod role_change;
```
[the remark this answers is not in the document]
Phoenix500526: In etcd, there are three components that care about it. They are the lessor, compactor, and readIndex. This pr only cover the lessor. Our compactor is not implemented yet. And the readIndex can wait until we refactored ReadState.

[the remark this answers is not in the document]
rogercloud: We don't need to change it in this PR, just a reminder. Maybe we need a issue to track.

rogercloud (dismissed):

xline/src/server/lease_server.rs
```diff
@@ -286,47 +255,30 @@ where
...
         let request_stream = request.into_inner();
-        let stream = if self.is_leader() {
-            self.leader_keep_alive(request_stream).await
-        } else {
-            self.follower_keep_alive(request_stream).await?
+        let stream = loop {
+            if self.lease_storage.is_primary() {
+                break self.leader_keep_alive(request_stream).await;
+            }
+            let leader_id = self.client.get_leader_id().await;
```
themanforfree, line 262: the `leader_id` returned by `get_leader_id` maybe oudated, because it will read from client's state first, and call this method may have one mord RTT because of `fetch_leader`

xline/src/server/lease_server.rs
```diff
@@ -346,4 +298,70 @@ where
...
+                let res = LeaseTimeToLiveResponse {
+                    header: Some(self.lease_storage.gen_header()),
+                    id: time_to_live_req.id,
+                    ttl: lease.remaining().as_secs().cast(),
+                    granted_ttl: lease.ttl().as_secs().cast(),
+                    keys,
+                };
+                return Ok(tonic::Response::new(res));
+            }
+            let leader_id = self.client.get_leader_id().await;
```
themanforfree, line 350: Same above

xline/src/server/lease_server.rs
```diff
@@ -346,4 +298,70 @@ where
...
+            });
+            if !self.lease_storage.is_primary() {
+                let mut lease_client = LeaseClient::connect(format!("http://{leader_addr}"))
+                    .await
+                    .map_err(|e| {
+                        tonic::Status::internal(format!("Connect to leader error: {e}"))
+                    })?;
+                return lease_client.lease_time_to_live(request).await;
+            }
+        }
```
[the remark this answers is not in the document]
Phoenix500526: Let's assume that the current server is a candidate and the lease store it has is not a primary lessor. So the self.client will wait for the new leader id to return. If the candidate lost the election, it should redirect this request to the real leader, otherwise, it should deal with this request by itself.

curp/tests/common/mod.rs
```diff
@@ -1,17 +1,59 @@
...
 
+#[derive(Default, Debug)]
+pub struct TestRoleChange {
+    pub inner: Arc<TestRoleChangeInner>,
+}
+
+#[derive(Default, Debug)]
+pub struct TestRoleChangeInner {
+    is_leader: AtomicBool,
+}
```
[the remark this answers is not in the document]
Phoenix500526: Since the RawCurp depends on the `RoleChange` rather than `Arc<RoleChange>`.  Adding this inner struct is just for sharing it so that we can make sure those callbacks will be invoked at the right time and in the right place.

rogercloud:
https://github.com/datenlord/Xline/issues/251#issuecomment-1567720298 Please refer the latest update in the issue.

xline/src/server/lease_server.rs
Phoenix500526: I leave issue #299  to track this problem. It will be fixed in a new pr

xline/src/server/lease_server.rs
Phoenix500526: Same as above

rogercloud (dismissed):

xline/src/server/lease_server.rs
```diff
@@ -41,10 +40,10 @@ where
     client: Arc<Client<Command>>,
     /// Server name
     name: String,
-    /// State of current node
-    state: Arc<State>,
     /// Id generator
     id_gen: Arc<IdGenerator>,
+    /// Address of all members
+    all_members: HashMap<String, String>,
 }
```
themanforfree, line 47: Please use `ServerId` as key

xline/src/server/lease_server.rs
```diff
@@ -57,16 +56,16 @@ where
         auth_storage: Arc<AuthStore<S>>,
         client: Arc<Client<Command>>,
         name: String,
-        state: Arc<State>,
         id_gen: Arc<IdGenerator>,
+        all_members: HashMap<String, String>,
```
themanforfree, line 60: Same above

xline/src/server/lease_server.rs
Phoenix500526: Done

themanforfree (approved):

bsbds (approved):

bsbds (approved):

PR #296, opened 2023-05-29
Ci/replace codecov

Phoenix500526:
replace our code coverage tool in our current ci progress.


CONTRIBUTING.md
```diff
@@ -199,6 +199,9 @@ uses conventional commits to automatically generating CHANGELOGs.
 
 Read the [docs](https://www.conventionalcommits.org) for more details.
 
+Note: Currently, the scope in the commit message is optional. Different projects
```
rogercloud, line 202: It seems that we need to add dependency on `coverage-helper` for the test. We'd better add a reminder here.

And we should note the way to get the coverage locally.

Phoenix500526: Done

Issue #299, opened 2023-06-01
Refactor: sharing state between CurpServer and CurpClient


Phoenix500526:
Currently, we use `Client::get_leader_id` to get the current leader id from a curp cluster.  The client maintains its own state.  The `leader_id` returned by `get_leader_id` may be outdated, because it will read from the client's state first.  

Read pr #276 and issue #251  for more detail. I'll file a new pr to resolve this issue.

Phoenix500526:
Finally， in that sharing the `State` structure between `RawCurp` and `Client` will break our present encapsulation， we've decided to add another method named `get_leader_info_from_curp` to solve this issue.  The `get_leader_info_from_curp` will fetch the current leader and term from the curp server which is on the same node as the client. Therefore, from the curp server perspective, the leader_id won't be outdated.

PR #302, opened 2023-06-02
Fix/issue 299

Phoenix500526:
Fix issue #299 .

Read issue #299  for more details.

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -19,8 +19,8 @@ use crate::{
 
 /// Protocol client
 pub struct Client<C: Command> {
-    /// client id
-    id: ServerId,
+    /// local server id. Only use in an inner client.
+    local_server_id: Option<ServerId>,
     /// Current leader and term
     state: RwLock<State>,
     /// All servers's `Connect`
@@ -94,12 +94,12 @@ where
     /// Create a new protocol client based on the addresses
     #[inline]
     pub async fn new(
-        self_id: ServerId,
+        self_id: Option<ServerId>,
         addrs: HashMap<ServerId, String>,
         timeout: ClientTimeout,
     ) -> Self {
         Self {
-            id: self_id,
+            local_server_id: self_id,
             state: RwLock::new(State::new()),
             connects: rpc::connect(addrs, None).await,
             timeout,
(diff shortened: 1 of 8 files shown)


curp/src/client.rs
```diff
@@ -19,6 +19,8 @@ use crate::{
 
 /// Protocol client
 pub struct Client<C: Command> {
+    /// Self id
```
rogercloud, line 22: Is this client a inner client? If not what's the value of `self_id` for outer client?

curp/src/client.rs
```diff
@@ -48,30 +50,23 @@ struct State {
...
 
     /// Set the leader and notify a waiter
     fn set_leader(&mut self, id: ServerId) {
         debug!("client update its leader to {id}");
-        if self.leader.as_ref().map_or(true, |prev_id| prev_id != &id) {
-            let _ignored = self.leader_tx.send(id.clone()).ok(); // it's ok to have no receiver
-        }
         self.leader = Some(id);
-        self.leader_notify.notify(1);
+        self.leader_notify.notify(usize::MAX);
```
rogercloud, line 69: If you mean notifying everyone, please correct the above comment.

curp/src/client.rs
Phoenix500526: This id is the client id. The inner clients use the same id as the curp server on the same node, while the outer clients use their customized id.

curp/src/client.rs
Phoenix500526: Done

curp/src/client.rs
```diff
@@ -19,6 +19,8 @@ use crate::{
 
 /// Protocol client
 pub struct Client<C: Command> {
+    /// client id
+    id: ServerId,
```
bsbds, line 23: In the comment, should it be local server id?

xline/src/client/mod.rs
```diff
@@ -68,11 +68,12 @@ impl Client {
         use_curp_client: bool,
         timeout: ClientTimeout,
     ) -> Result<Self, ClientError> {
+        let name = String::from("client");
         let etcd_client =
             EtcdClient::connect(all_members.values().cloned().collect_vec(), None).await?;
-        let curp_client = CurpClient::new(all_members, timeout).await;
+        let curp_client = CurpClient::new(name.clone(), all_members, timeout).await;
```
bsbds, line 74: The server id should be optional, otherwise, it will be confusing to see that `fetch_local_leader_info` will use the client id to fetch the local server.

xline/src/server/lease_server.rs
```diff
@@ -259,7 +259,7 @@ where
             if self.lease_storage.is_primary() {
                 break self.leader_keep_alive(request_stream).await;
             }
-            let leader_id = self.client.get_leader_id().await;
+            let leader_id = self.client.get_leader_id_from_curp().await;
```
themanforfree, line 262: Please do this change at line 352 too

Phoenix500526: Done

curp/src/client.rs
Phoenix500526: How come?

rogercloud (approved):

themanforfree (approved):

PR #318, opened 2023-06-09
Feature/Xline client kv

bsbds:
Implement kv client in `xline-client`

based on #317

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #319, opened 2023-06-09
Feature/Xline client lease

bsbds:
Implement lease client in xline-client

based on #318

### DIFF
--- a/xline-client/src/clients/lease.rs
+++ b/xline-client/src/clients/lease.rs
@@ -1,13 +1,24 @@
-// TODO: Remove these when the placeholder is implemented.
-#![allow(dead_code)]
-
 use std::sync::Arc;
 
-use curp::client::Client as CurpClient;
-use tonic::transport::Channel;
+use curp::{client::Client as CurpClient, cmd::ProposeId};
+use futures::channel::mpsc::channel;
+use tonic::{transport::Channel, Streaming};
+use uuid::Uuid;
 use xline::server::Command;
+use xlineapi::{
+    LeaseGrantResponse, LeaseKeepAliveResponse, LeaseLeasesResponse, LeaseRevokeResponse,
+    LeaseTimeToLiveResponse, RequestWithToken,
+};
 
-use crate::AuthService;
+use crate::{
+    error::{ClientError, Result},
+    lease_gen::LeaseIdGenerator,
+    types::lease::{
+        LeaseGrantRequest, LeaseKeepAliveRequest, LeaseKeeper, LeaseRevokeRequest,
+        LeaseTimeToLiveRequest,
+    },
+    AuthService,
+};
 
 /// Client for Lease operations.
 #[derive(Clone, Debug)]
@@ -20,6 +31,8 @@ pub struct LeaseClient {
     lease_client: xlineapi::LeaseClient<AuthService<Channel>>,
     /// Auth token
     token: Option<String>,
+    /// Lease Id generator
+    id_gen: Arc<LeaseIdGenerator>,
 }
 
 impl LeaseClient {
@@ -30,6 +43,7 @@ impl LeaseClient {
         curp_client: Arc<CurpClient<Command>>,
         channel: Channel,
         token: Option<String>,
+        id_gen: Arc<LeaseIdGenerator>,
     ) -> Self {
         Self {
             name,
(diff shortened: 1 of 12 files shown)


xline-client/src/clients/lease.rs
```diff
@@ -39,6 +53,117 @@ impl LeaseClient {
...
+    ///
+    /// If `CurpClient` failed to send request
+    #[inline]
+    pub async fn revoke(&mut self, request: LeaseRevokeRequest) -> Result<LeaseRevokeResponse> {
+        let propose_id = self.generate_propose_id();
+        let request = RequestWithToken::new_with_token(
+            xlineapi::LeaseRevokeRequest::from(request).into(),
+            self.token.clone(),
+        );
+        let cmd = Command::new(vec![], request, propose_id);
```
themanforfree, line 92: Lease revoke request should conflict with other requests that have the same keys attached to this lease, but we cannot get keys at client side, maybe we need to let this request send to `LeaseServer`

bsbds: done

themanforfree (dismissed):

xline-client/src/lease_gen.rs
```diff
@@ -0,0 +1,55 @@
...
+            panic!("Failed to generate random bytes for lease id generator: {err}");
+        });
+        let prefix = u64::from_be_bytes(buf).overflowing_shl(1).0;
+        let suffix = AtomicU64::new(0);
+        Self { prefix, suffix }
+    }
+
+    /// Generate next id
+    #[inline]
+    pub fn next(&self) -> i64 {
```
Phoenix500526, line 38: Why do we return a `i64` instead of `u64`? Can a lease id be negative?

xline-client/src/lease_gen.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    /// panic if failed to generate random bytes
+    #[inline]
+    #[must_use]
+    pub fn new() -> Self {
+        let mut buf = [0u8; 8];
+        getrandom::getrandom(&mut buf).unwrap_or_else(|err| {
+            panic!("Failed to generate random bytes for lease id generator: {err}");
+        });
+        let prefix = u64::from_be_bytes(buf).overflowing_shl(1).0;
```
Phoenix500526, line 31: Here should be `overflowing_shl(8)` in that a byte equals eight bits. FYI: https://doc.rust-lang.org/std/primitive.u64.html#method.overflowing_shl

xline-client/src/lease_gen.rs
```diff
@@ -0,0 +1,55 @@
...
+        let id = self.prefix | suffix;
+        (id & 0x7fff_ffff_ffff_ffff).cast()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[test]
+    fn test_id_generator() {
```
Phoenix500526, line 49: This test case doesn't make any sense. Firstly, the number of test samples is too few. Please ensure an adequate number of test cases, for example, 256. Secondly, the coverage of the test is not comprehensive. In addition to ensuring that two consecutive next calls do not produce the same value, it is also necessary to ensure that the generated ID suffix always has meaningful values.

xline-client/src/lease_gen.rs
bsbds: The lease ID defined in protobuf api is i64. However only zero and positive value are used, I guess the negative id is reserved.

xline-client/src/lease_gen.rs
bsbds: more tests have been added

xline-client/src/lease_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+    use std::collections::HashSet;
+
+    use super::*;
+
+    #[test]
+    fn no_duplicates_in_one() {
+        let mut exist = HashSet::new();
+        let id_gen = LeaseIdGenerator::new();
+
+        for _ in 0..(1 << 20) {
```
Phoenix500526, line 55: Does the 20 here has any specific meaning?

xline-client/src/lease_gen.rs
```diff
@@ -0,0 +1,84 @@
...
+            let id = id_gen.next();
+            assert!(exist.insert(id), "id {id} is duplicated");
+        }
+    }
+
+    #[test]
+    fn no_duplicates_in_multi() {
+        let mut exist = HashSet::new();
+
+        for _ in 0..(1 << 10) {
```
Phoenix500526, line 65: Same as above. If the number 10 doesn't have any meaning, please use 1024 directly. We should keep it simple and stupid (KISS principle).

xline-client/src/lease_gen.rs
bsbds: No it just test it with a rather large number

xline-client/src/lease_gen.rs
bsbds: done, I change them to 1_000_000 and 1000 respectively

xline-client/tests/lease.rs
```diff
@@ -0,0 +1,102 @@
...
+    error::Result,
+    types::lease::{
+        LeaseGrantRequest, LeaseKeepAliveRequest, LeaseRevokeRequest, LeaseTimeToLiveRequest,
+    },
+};
+
+mod common;
+
+#[tokio::test]
+async fn grant_revoke() -> Result<()> {
```
Phoenix500526, line 12: This test case name is meaningless. Please rename it. A good test case name can be divided into three parts: the unit of work name, the expected behavior, and the scenario. You can use the pattern that looks like `xxx_should_xxx_when_xxx` or `xxx_should_success_in_normal_paht`. For example, `lease_grant_should_failed_when_xxx` or `lease_grant_should_success_in_normal_path` . Any thought of this?@rogercloud

xline-client/tests/lease.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let resp = client.grant(LeaseGrantRequest::new(123)).await?;
+    assert_eq!(resp.ttl, 123);
+    let id = resp.id;
+    client.revoke(LeaseRevokeRequest::new(id)).await?;
+    Ok(())
+}
+
+#[tokio::test]
+async fn keep_alive() -> Result<()> {
```
Phoenix500526, line 24: Same as above

xline-client/tests/lease.rs
```diff
@@ -0,0 +1,102 @@
...
+    let resp = stream.message().await?.unwrap();
+    assert_eq!(resp.id, keeper.id());
+    assert_eq!(resp.ttl, 60);
+
+    client.revoke(LeaseRevokeRequest::new(id)).await?;
+    Ok(())
+}
+
+#[tokio::test]
+async fn time_to_live() -> Result<()> {
```
Phoenix500526, line 44: Same as above

xline-client/tests/lease.rs
```diff
@@ -0,0 +1,102 @@
...
+        .await?;
+    assert_eq!(resp.id, leaseid);
+    assert_eq!(resp.granted_ttl, 60);
+
+    client.revoke(LeaseRevokeRequest::new(leaseid)).await?;
+    Ok(())
+}
+
+#[tokio::test]
+async fn leases() -> Result<()> {
```
Phoenix500526, line 66: Same as above

xline-client/tests/lease.rs
bsbds: done

Phoenix500526 (approved):

PR #321, opened 2023-06-09
Feature/Xline client watch

bsbds:
Implement watch client in xline-client

### DIFF
--- a/xline-client/src/clients/watch.rs
+++ b/xline-client/src/clients/watch.rs
@@ -1,11 +1,17 @@
-// TODO: Remove these when the placeholder is implemented.
-#![allow(dead_code)]
-
 use std::{fmt::Debug, sync::Arc};
 
+use futures::channel::mpsc::channel;
 use tonic::transport::Channel;
+use xlineapi::{self, RequestUnion, WatchResponse};
 
-use crate::AuthService;
+use crate::{
+    error::{ClientError, Result},
+    types::watch::{WatchRequest, Watcher},
+    AuthService,
+};
+
+/// Channel size for watch request stream
+const CHANNEL_SIZE: usize = 128;
 
 /// The maintenance client
 #[derive(Clone, Debug)]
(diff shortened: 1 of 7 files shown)


xline-client/src/clients/watch.rs
```diff
@@ -26,4 +29,267 @@ impl WatchClient {
...
+    ///
+    /// # Panics
+    ///
+    /// If the RPC server returns the wrong value
+    #[inline]
+    pub async fn watch(
+        &mut self,
+        request: WatchRequest,
+    ) -> Result<(Watcher, tonic::Streaming<WatchResponse>)> {
+        let (mut request_sender, request_receiver) = channel::<xlineapi::WatchRequest>(100);
```
Phoenix500526, line 50: Don't use a magic number. And I think 128 is better.

xline-client/src/clients/watch.rs
```diff
@@ -26,4 +29,267 @@ impl WatchClient {
...
+
+    /// Watches for events happening or that have happened.
+    ///
+    /// # Errors
+    ///
+    /// If sender fails to send to channel
+    #[inline]
+    pub fn watch(&mut self, request: WatchRequest) -> Result<()> {
+        let request = xlineapi::WatchRequest {
+            request_union: Some(RequestUnion::CreateRequest(request.into())),
```
Phoenix500526, line 109: What are the differences between `WatchClient::watch` and `Watcher::watch`? Both of them send the same `CreateRequest` via `self.sender`. Shall we create twice?

xline-client/tests/watch.rs
```diff
@@ -0,0 +1,43 @@
...
+    watcher.cancel()?;
+
+    let resp = stream.message().await?.unwrap();
+    assert_eq!(resp.watch_id, watcher.watch_id());
+    assert!(resp.canceled);
+
+    Ok(())
+}
+
+async fn get_cluster_client() -> Result<(Cluster, Client)> {
```
Phoenix500526, line 38: The `get_cluster_client` is only used in `watch.rs`. If you don't use the cluster returned by the `get_cluster_client`, then don't return it.

xline-client/src/clients/watch.rs
bsbds: The Watcher is used to add new watch request or cancel existing request interactively, while WatchClient::watch can only watch on a const range of keys.

xline-client/tests/watch.rs
bsbds: The cluster will stop itself when dropped, so we must move it to the test funcitons.

xline-client/src/clients/watch.rs
bsbds: done

xline-client/src/clients/watch.rs
```diff
@@ -26,4 +31,268 @@ impl WatchClient {
...
+    #[inline]
+    pub fn cancel_by_id(&mut self, watch_id: i64) -> Result<()> {
+        let request = xlineapi::WatchRequest {
+            request_union: Some(RequestUnion::CancelRequest(WatchCancelRequest { watch_id })),
+        };
+
+        self.sender
+            .try_send(request)
+            .map_err(|e| ClientError::WatchError(e.to_string()))
+    }
```
themanforfree, line 152: Can `Watcher` cancel other `Watcher`s ?

xline-client/src/clients/watch.rs
```diff
@@ -26,4 +31,268 @@ impl WatchClient {
...
+                key: key.into(),
+                ..Default::default()
+            },
+        }
+    }
+
+    /// Set `key` and `range_end` when with prefix
+    #[inline]
+    #[must_use]
+    pub fn with_prefix(mut self) -> Self {
```
themanforfree, line 195: Same as auth client

xline-client/tests/watch.rs
themanforfree: I suggest use this function in all clients tests, no need to write this function for every test

xline-client/src/clients/watch.rs
bsbds: I think a `Watcher` is only responsible for the one watch request stream created by `WatchClient::watch`. So If two watcher is created they will not interfere with each other.

xline-client/src/clients/watch.rs
bsbds: done

xline-client/tests/watch.rs
bsbds: done

Phoenix500526 (approved):

themanforfree (approved):

PR #322, opened 2023-06-09
Feature/Xline client lock

bsbds:
Implement lock client in xline-client

based on #321 

### DIFF
--- a/xline-client/src/clients/lock.rs
+++ b/xline-client/src/clients/lock.rs
@@ -1,7 +1,14 @@
-use std::sync::Arc;
+use std::{
+    pin::Pin,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+};
 
 use clippy_utilities::OverflowArithmetic;
 use curp::{client::Client as CurpClient, cmd::ProposeId};
+use futures::{Future, FutureExt};
 use tonic::transport::Channel;
 use uuid::Uuid;
 use xline::server::{Command, CommandResponse, KeyRange, SyncResponse};
@@ -85,13 +92,32 @@ impl LockClient {
                 .await?;
             lease_id = resp.id;
         }
-
         let prefix = format!(
             "{}/",
             String::from_utf8_lossy(&request.inner.name).into_owned()
         );
         let key = format!("{prefix}{lease_id:x}");
+        let lock_success = AtomicBool::new(false);
+        let lock_inner = self.lock_inner(prefix, key.clone(), lease_id, &lock_success);
+        tokio::pin!(lock_inner);
 
+        LockFuture {
+            key,
+            lock_success: &lock_success,
+            lock_client: self,
+            lock_inner,
+        }
+        .await
+    }
+
+    /// The inner lock logic
+    async fn lock_inner(
+        &self,
+        prefix: String,
+        key: String,
+        lease_id: i64,
+        lock_success: &AtomicBool,
+    ) -> Result<LockResponse> {
         let txn = Self::create_acquire_txn(&prefix, lease_id);
         let (cmd_res, sync_res) = self.propose(txn, false).await?;
         let mut txn_res = Into::<TxnResponse>::into(cmd_res.decode());
@@ -118,10 +144,7 @@ impl LockClient {
         {
             owner_res.header
         } else {
-            if let Err(e) = self.wait_delete(prefix, my_rev).await {
-                let _ignore = self.delete_key(key.as_bytes()).await;
-                return Err(e);
-            }
+            self.wait_delete(prefix, my_rev).await?;
             let range_req = RangeRequest {
                 key: key.as_bytes().to_vec(),
                 ..Default::default()
@@ -141,6 +164,8 @@ impl LockClient {
                 }
             }
         };
+
+        let _ignore = lock_success.fetch_or(true, Ordering::Relaxed);
 
         Ok(LockResponse {
             header,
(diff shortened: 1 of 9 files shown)


xline-client/src/clients/lock.rs
```diff
@@ -16,26 +30,272 @@ pub struct LockClient {
...
+    /// lease associate with the owner expires.
+    ///
+    /// # Errors
+    ///
+    /// If `CurpClient` fails to send request
+    #[inline]
+    pub async fn lock(&mut self, request: LockRequest) -> Result<LockResponse> {
+        let mut lease_id = request.inner.lease;
+        if lease_id == 0 {
+            let resp = self.lease_client.grant(LeaseGrantRequest::new(60)).await?;
```
themanforfree, line 82: Please use a const instead of magic number

xline-client/tests/lock.rs
```diff
@@ -0,0 +1,23 @@
...
+    let resp = client
+        .lock(LockRequest::new().with_name("lock-test"))
+        .await?;
+    let key = resp.key;
+    let key_str = std::str::from_utf8(&key).unwrap();
+    assert!(key_str.starts_with("lock-test/"));
+
+    client.unlock(UnlockRequest::new().with_key(key)).await?;
+    Ok(())
+}
```
iGxnon, line 23: Add more test case? Such as the lock contention.

xline-client/src/clients/lock.rs
bsbds: done

xline-client/src/clients/lock.rs
```diff
@@ -16,26 +30,272 @@ pub struct LockClient {
...
+            ),
             watch_client: WatchClient::new(channel, token.clone()),
             token,
         }
     }
+
+    /// Acquires a distributed shared lock on a given named lock.
+    /// On success, it will return a unique key that exists so long as the
+    /// lock is held by the caller. This key can be used in conjunction with
+    /// transactions to safely ensure updates to etcd only occur while holding
```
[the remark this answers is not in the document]
bsbds: fixed

xline-client/tests/lock.rs
bsbds: done

xline-client/tests/lock.rs
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    rx.recv().await.unwrap();
+    let now = tokio::time::Instant::now();
+    let resp = client
+        .lock(LockRequest::new().with_name("lock-test"))
+        .await
+        .unwrap();
+
+    assert!(now.elapsed() > Duration::from_secs(1));
```
Phoenix500526, line 56: Why is `Duration::from_secs(1)` instead of `Duration::from_secs(2)`?

bsbds: The actual wait time will be around 2secs(more or less). 1s is a rather conservative value, just to demonstrate the lock request will not return immediately.

Phoenix500526: How about this?
```suggestion
    let res = tokio::timeout(Duration::from_secs(2), client.lock(LockRequest::new().with_name("lock-test")))
    assert!(res.is_err());
    let resp = client
        .lock(LockRequest::new().with_name("lock-test"))
        .await
        .unwrap();
    assert!(resp.key.starts_with(b"lock-test/"));
```

bsbds: The duration may still need to be less than 2, because the sleep in another task may already began for a while before we wait for the timeout.

Phoenix500526: This testing way lacks persuasiveness. Or, you can try a different approach.
```suggestion
    let (tx, rx) = std::sync::mpsc::sync_channel(0);  // rendezvous channel

    tokio::spawn(async move {
        let resp = client_c
            .lock(LockRequest::new().with_name("lock-test"))
            .await
            .unwrap();
        rx.recv().await.unwrap();
        let _resp = client_c
            .unlock(UnlockRequest::new().with_key(resp.key))
            .await
            .unwrap();
    });

    let res = tokio::timeout(Duration::from_secs(2), client.lock(LockRequest::new().with_name("lock-test")))
    assert!(res.is_err());
    tx.send();
    let res = tokio::timeout(Duration::from_micros(100), client.lock(LockRequest::new().with_name("lock-test")))
    assert!(res.is_ok());
    ….
```

bsbds: After change to this timeout syntax, I encounter an issue that the final lock request will never return. I find out that it may be because the lock is not cancel safe, after it canceled by timeout, it will not delete the key associated with the lock thus another lock attemp will just wait forever.

themanforfree (approved):

xline-client/src/clients/lock.rs
```diff
@@ -16,26 +30,275 @@ pub struct LockClient {
...
+    /// lock is held by the caller. This key can be used in conjunction with
+    /// transactions to safely ensure updates to Xline only occur while holding
+    /// lock ownership. The lock is held until Unlock is called on the key or the
+    /// lease associate with the owner expires.
+    ///
+    /// # Errors
+    ///
+    /// If `CurpClient` fails to send request
+    #[inline]
+    pub async fn lock(&self, request: LockRequest) -> Result<LockResponse> {
```
iGxnon, line 79: I think we can use a `LockGuard` and implement its `Drop` trait to unlock automatically, instead of just returning a wrapper of some data. This design enforces [RAII](https://doc.rust-lang.org/rust-by-example/scope/raii.html) style in Rust.

xline-client/tests/lock.rs
bsbds: after implemented cancellation clean up, the test is refactored base on your approach

iGxnon (approved):

themanforfree (dismissed):

xline-client/src/clients/lock.rs
```diff
@@ -16,26 +37,343 @@ pub struct LockClient {
...
+                    res.header
+                }
+                Err(e) => {
+                    let _ignore = self.delete_key(key.as_bytes()).await;
+                    return Err(e);
+                }
+            }
+        };
+
+        let _ignore = lock_success.fetch_or(true, Ordering::Relaxed);
```
Phoenix500526, line 168: Why do you invoke `fetch_or` here? We can directly store it as true.

bsbds: fixed

Phoenix500526 (approved):

iGxnon (approved):

Phoenix500526 (approved):

PR #323, opened 2023-06-09
Feature/Xline client maintenance

bsbds:
Implement maintenance client in xline-client

### DIFF
--- a/xline-client/tests/maintenance.rs
+++ b/xline-client/tests/maintenance.rs
@@ -1,0 +1,27 @@
+use xline_client::{clients::maintenance::MaintenanceClient, error::Result, Client, ClientOptions};
+use xline_test_utils::Cluster;
+
+#[tokio::test]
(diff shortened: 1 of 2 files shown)


xline-client/tests/maintenance.rs
```diff
@@ -0,0 +1,27 @@
...
+            assert!(!resp.blob.is_empty());
+            if resp.remaining_bytes == 0 {
+                break;
+            }
+        }
+    }
+    Ok(())
+}
+
+async fn get_cluster_client() -> Result<(Cluster, MaintenanceClient)> {
```
Phoenix500526, line 20: There is not need to return Cluster in `get_cluster_client`

themanforfree: I think we need to return it, because the cluster will stop when it is dropped.

themanforfree (approved):

Phoenix500526 (approved):

PR #326, opened 2023-06-14
Ci/upload container image

### LINKED ISSUE #325: [Feature]:  Relaese Container Image
### Description about the feature

It would be great if we can release container image of Xline, I have find that docker image from https://hub.docker.com/r/datenlord/xline/tags and seems like it's not active.

### Code of Conduct

Phoenix500526:
Update the MAINTAINER doc and upload a container image when publishing a new release. Closes #325.  

Read #325 for more details.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,12 +48,26 @@ jobs:
           asset_path: ./target/release/xline
           asset_name: xline-linux-amd64
           asset_content_type: application/octet-stream
-
-      - name: Close the Relevant Milestone
-        id: close-milestone
-        uses: adlerhsieh/prepare-release@0.1.2
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          REPO_OWNER: 'datenlord'
-          REPO: 'Xline'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -48,12 +48,26 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          REPO_OWNER: 'datenlord'
-          REPO: 'Xline'
-          IGNORE_MILESTONE_NOT_FOUND: false
+      - run: |
+          mv ./target/release/xline ./scripts
+          mv ./target/release/benchmark ./scripts
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
```
bsbds, line 55: I think qemu is not needed as we only build linux amd64 target.

themanforfree (approved):

.github/workflows/release.yml
Phoenix500526: I think we can keep it in that we may support for other platforms.

bsbds (approved):

Issue #339, opened 2023-06-21
[Bug]: EventListener will lose event since it's not cancellation safe.


Phoenix500526:
### Description about the bug

When using select! in a loop to receive messages from multiple sources, we should make sure that the receive call is cancellation safe to avoid losing messages. Unfortunately, `EventListener` is not cancellation safe. That means some implementation in our code may lose event due to the select! cancel it. 

Here is the `task` method in `watch_server.rs`, and its implementation lead to events loss.
```Rust
    async fn task<ST, W>(
        next_id_gen: Arc<WatchIdGenerator>,
        kv_watcher: Arc<W>,
        res_tx: mpsc::Sender<Result<WatchResponse, tonic::Status>>,
        mut req_rx: ST,
        header_gen: Arc<HeaderGenerator>,
        watch_progress_notify_interval: Duration,
    ) where
        ST: Stream<Item = Result<WatchRequest, tonic::Status>> + Unpin,
        W: KvWatcherOps,
    {
        let (event_tx, mut event_rx) = mpsc::channel(CHANNEL_SIZE);
        let stop_notify = Arc::new(Event::new());
        let mut watch_handle = WatchHandle::new(
            kv_watcher,
            res_tx,
            event_tx,
            Arc::clone(&stop_notify),
            next_id_gen,
            header_gen,
        );
        let mut ticker = tokio::time::interval(watch_progress_notify_interval);
        loop {
            tokio::select! {
                req = req_rx.next() => {
                    if let Some(req) = req {
                        match req {
                            Ok(req) => {
                                watch_handle.handle_watch_request(req).await;
                            }
                            Err(e) => {
                                warn!("Receive WatchRequest error {:?}", e);
                                break;
                            }
                        }
                    } else {
                        warn!("Watch client closes connection");
                        break;
                    }
                }
                event = event_rx.recv() => {
                    if let Some(event) = event {
                        watch_handle.handle_watch_event(event).await;
                    } else {
                        panic!("Watch event sender is closed");
                    }
                }
                _ = ticker.tick() => {
                    watch_handle.handle_tick_progress().await;
                }
                _ = stop_notify.listen() => {
                    break;
                }
            }
        }
    }
}
```

[FYI: https://docs.rs/tokio/1.28.2/tokio/macro.select.html](https://docs.rs/tokio/1.28.2/tokio/macro.select.html#cancellation-safety)

### Version

0.4.1 (Default)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #338, opened 2023-06-21
unexpected coverage changes in a pr which doesn't modify code


Phoenix500526:
Currently, in some prs where the code has not been modified, like pr #332, there are unexpected fluctuations in code coverage. 

The reasons for the coverage fluctuation are as follows:
1. some upstream issues will lead to code coverage missing some code lines. FYI, [Code coverage misses .await lines](https://github.com/rust-lang/rust/issues/98712) and [Unexpected behavior in the coverage of the lines of the test functions](https://github.com/taiki-e/cargo-llvm-cov/issues/243)
2. There are some non-determinism in our tests

PR #340, opened 2023-06-21
Codecov/fluctuation

Phoenix500526:
1. remove some non-determinism in tests. #338 
2. fix bug #339 

Add some test cases to cover some code branch determinism.
Fix bug #339 and add a test `test_watch_task_should_be_terminated_when_response_tx_is_closed` to cover it.

No


xline/src/server/watch_server.rs
```diff
@@ -111,7 +113,9 @@ where
                 _ = ticker.tick() => {
                     watch_handle.handle_tick_progress().await;
                 }
-                _ = stop_notify.listen() => {
+                // To ensure that each iteration invokes the same `stop_listener` and keeps
+                // events losing due to the cancellation of `stop_listener` at bay.
+                _ = &mut stop_listener => {
```
bsbds, line 118: Why it is cancellation safe to use the same `stop_listener`

Phoenix500526: In fact, `stop_listener` is not cancellation safe. Due to the ownership of `stop_listner` is not belong to `select!`, it won't be canceled by `select!`.

themanforfree (approved):

Issue #343, opened 2023-06-27
[Bug]: lease may not synced in lease server


bsbds:
### Description about the bug

Currently the `lease_grant` method is using fast path of the CURP protocol

https://github.com/datenlord/Xline/blob/71249bda1ba141b07030d45fa059b96278aba068/xline/src/server/lease_server.rs#L236-L237

However in `lease_keep_alive`, `lease_time_to_live`, `lease_leases` methods, they will attempt to read from local storage, which may not flushed or synced yet and lead to inconsistency.

for instance in `lease_keep_alive`

https://github.com/datenlord/Xline/blob/71249bda1ba141b07030d45fa059b96278aba068/xline/src/server/lease_server.rs#L175

and in `lease_leases`
https://github.com/datenlord/Xline/blob/71249bda1ba141b07030d45fa059b96278aba068/xline/src/server/lease_server.rs#L291-L293

To resolve this issue, we must ensure these methods wait for the synced event to occur after `lease_grant` or `lease_revoke` are called. This synchronization process ensures that the client receives the most up-to-date view of the lease storage.

### Version

0.1.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bsbds:
I plan to refactor `lease_leases` and `lease_grant/revoke` by two seperate strategy.

**For `lease_leases`:**

We can treat `lease_leases` a kind of lease read request, and set it to conflict with `lease_grant` and `lease_revoke` request.

```rust
    pub fn is_lease_read_request(&self) -> bool {
        matches!(*self, RequestWrapper::LeaseLeasesRequest(_))
    }

    pub fn is_lease_write_request(&self) -> bool {
        matches!(
            *self,
            RequestWrapper::LeaseGrantRequest(_) | RequestWrapper::LeaseRevokeRequest(_)
        )
    }
```
in `ConflictCheck`
```rust
        if (this_req.is_lease_read_request() && other_req.is_lease_write_request())
            || (this_req.is_lease_write_request() && other_req.is_lease_read_request())
        {
            return true;
        }
```
then can we add the lease_leases handler to lease store.

**For `lease_grant/revoke`:**

We maintain an `unsynced_cache` to track leases that is not flushed to db yet.
```rust
#[derive(Debug)]
pub(crate) struct LeaseStore<DB>
where
    DB: StorageApi,
{
    ...
    /// cache unsynced lease id
    unsynced_cache: Arc<RwLock<HashSet<i64>>>,
[...]

PR #344, opened 2023-06-27
Fix/Lease sync issue

bsbds:
Described in: #343 

### DIFF
--- a/xline/src/server/lease_server.rs
+++ b/xline/src/server/lease_server.rs
@@ -17,8 +17,8 @@ use crate::{
     rpc::{
         Lease, LeaseClient, LeaseGrantRequest, LeaseGrantResponse, LeaseKeepAliveRequest,
         LeaseKeepAliveResponse, LeaseLeasesRequest, LeaseLeasesResponse, LeaseRevokeRequest,
-        LeaseRevokeResponse, LeaseStatus, LeaseTimeToLiveRequest, LeaseTimeToLiveResponse,
-        RequestWithToken, RequestWrapper,
+        LeaseRevokeResponse, LeaseTimeToLiveRequest, LeaseTimeToLiveResponse, RequestWithToken,
+        RequestWrapper,
     },
     storage::{storage_api::StorageApi, AuthStore, LeaseStore},
 };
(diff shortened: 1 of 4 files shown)


bsbds:
Tests is added to cover the newly added code : 20381e6ff0f1a9fecf2a1f100008ead829a99c86

xline/src/server/lease_server.rs
```diff
@@ -286,16 +287,18 @@ where
...
-            .leases()
-            .into_iter()
-            .map(|lease| LeaseStatus { id: lease.id() })
-            .collect();
-        let res = LeaseLeasesResponse {
-            header: Some(self.lease_storage.gen_header()),
-            leases,
-        };
+
+        let is_fast_path = true;
```
Phoenix500526, line 291: Why should we issue a proposal in the `lease_leases` method?

xline/src/server/lease_server.rs
```diff
@@ -170,6 +170,7 @@ where
                 debug!("Receive LeaseKeepAliveRequest {:?}", keep_alive_req);
                 // TODO wait applied index
```
Phoenix500526, line 171: If we have `wait_synced` method here, shall we keep this TODO? @themanforfree. This TODO comment is introduced in the commit 42e85d61b1496592ff4bf358331123a17b1634f2.

xline/src/server/lease_server.rs
bsbds: Because some lease grant or revoke may not flush to db yet, so we need to let lease_leases go through the curp protocol.

xline/src/storage/lease_store/mod.rs
```diff
@@ -52,6 +54,10 @@ where
     kv_update_tx: mpsc::Sender<(i64, Vec<Event>)>,
     /// Primary flag
     is_primary: AtomicBool,
+    /// cache unsynced lease id
+    unsynced_cache: Arc<RwLock<HashSet<i64>>>,
```
themanforfree, line 58: If we don't need to hold the lock during await, please use `parking_lot::RwLock`

xline/src/storage/lease_store/mod.rs
```diff
@@ -155,14 +164,41 @@ where
...
+        let lease_id = match *wrapper {
+            RequestWrapper::LeaseGrantRequest(ref req) => req.id,
+            RequestWrapper::LeaseRevokeRequest(ref req) => req.id,
+            _ => {
+                return;
+            }
+        };
+
+        _ = self.unsynced_cache.write().await.remove(&lease_id);
+        self.sync_event.notify_waiters();
```
themanforfree, line 180: The other parts of `Xline` use `eventer_listener` to implement similar abilities, which way is batter? should we unify them

xline/src/storage/lease_store/mod.rs
bsbds: done

xline/src/storage/lease_store/mod.rs
bsbds: It seems the `event_listener` is synchronous, the `tokio::sync::Notify` will be better?

themanforfree: It can also be used asynchronous

xline/src/server/lease_server.rs
themanforfree: I think we can remove it here

xline/src/storage/lease_store/mod.rs
bsbds: I think I better use `event_listener` for better consistency.

xline/src/server/lease_server.rs
bsbds: done

Phoenix500526 (approved):

themanforfree (approved):

Issue #335, opened 2023-06-20
[Feature]: Support single node cluster


liangyuanpeng:
### Description about the feature

Today, Xline seems like is not support single node cluster, so create the issue to tracing it.  I think it is a low level feature now.

# Reproduce it

1. docker network create xline --subnet=172.20.0.0/24
2. docker run -e RUST_LOG=debug --network xline -it -d --name xline -p 2379:2379 datenlord/xline:v0.4.1 xline --name node1 --members node1=172.20.0.2:2379 --data-dir /tmp/xline --storage-engine rocksdb
3.  docker logs xline and then you can see the log of always election

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #346, opened 2023-06-27
Feature/single node cluster

themanforfree:

#335 
handle the election and the propose of a single node cluster
no

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -553,6 +553,7 @@ mod tests {
 
     #[test]
     fn superquorum_should_work() {
+        assert_eq!(superquorum(1), 1);
         assert_eq!(superquorum(11), 9);
         assert_eq!(superquorum(97), 73);
         assert_eq!(superquorum(31), 24);
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -109,11 +109,12 @@ where
 
     /// The fast round of Curp protocol
     /// It broadcast the requests to all the curp servers.
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
     async fn fast_round(
         &self,
         cmd_arc: Arc<C>,
     ) -> Result<(Option<<C as Command>::ER>, bool), ProposeError> {
+        debug!("fast round for cmd({}) started", cmd_arc.id());
         let req = ProposeRequest::new(cmd_arc.as_ref())?;
         let mut rpcs: FuturesUnordered<_> = self
             .connects
@@ -181,7 +182,7 @@ where
                 },
             )??;
             if (ok_cnt >= superquorum) && execute_result.is_some() {
-                debug!("fast round succeeds");
+                debug!("fast round for cmd({}) succeed", cmd_arc.id());
                 return Ok((execute_result, true));
             }
         }
@@ -189,11 +190,12 @@ where
     }
 
     /// The slow round of Curp protocol
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
     async fn slow_round(
         &self,
         cmd: Arc<C>,
     ) -> Result<(<C as Command>::ASR, <C as Command>::ER), ProposeError> {
+        debug!("slow round for cmd({}) started", cmd.id());
         let retry_timeout = *self.timeout.retry_timeout();
         loop {
             // fetch leader id
@@ -409,6 +411,7 @@ where
     ///   If leader index is out of bound of all the connections, panic
     #[inline]
     #[allow(clippy::too_many_lines)] // FIXME: split to smaller functions
+    #[instrument(skip_all, fields(cmd_id=%cmd.id()))]
     pub async fn propose(&self, cmd: C) -> Result<C::ER, ProposeError> {
         let cmd_arc = Arc::new(cmd);
         let fast_round = self.fast_round(Arc::clone(&cmd_arc));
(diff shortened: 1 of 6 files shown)


curp/src/client.rs
```diff
@@ -181,19 +182,20 @@ where
...
+                debug!("fast round for cmd({}) succeed", cmd_arc.id());
                 return Ok((execute_result, true));
             }
         }
         Ok((execute_result, false))
     }
 
     /// The slow round of Curp protocol
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
```
Phoenix500526, line 193: Same as above

curp/src/client.rs
```diff
@@ -109,11 +109,12 @@ where
 
     /// The fast round of Curp protocol
     /// It broadcast the requests to all the curp servers.
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
```
Phoenix500526, line 112: If you want to print the `cmd.id()` instead of the `cmd`, you can use the adding field. FYI: https://docs.rs/tracing/0.1.38/tracing/attr.instrument.html#adding-fields

```suggestion
    #[instrument(skip(self), field(cmd_id = cmd_arc.id()) )]
```

xline/src/server/kv_server.rs
```diff
@@ -391,13 +386,13 @@ where
     S: StorageApi,
 {
     /// Range gets the keys in the range from the key-value store.
-    #[instrument(skip(self))]
+    #[instrument(skip_all)]
```
Phoenix500526, line 389: Same as above.
```suggestion
    #[instrument(skip(self), field(range_request = request.get_ref()))]
```

curp/src/server/raw_curp/mod.rs
```diff
@@ -812,20 +828,43 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
             debug!("Candidate {} restarts election", self.id());
         }
 
-        Vote {
-            term: st.term,
-            candidate_id: self.id().clone(),
-            last_log_index: log.last_log_index(),
-            last_log_term: log.last_log_term(),
+        // vote to self
+        let election_ends = {
```
Phoenix500526, line 839: This `election_ends` looks redundant.

```suggestion
        if cst.votes_received < min_granted {
            Some(Vote {
                term: st.term,
                candidate_id: self.id().clone(),
                last_log_index: log.last_log_index(),
                last_log_term: log.last_log_term(),
            })
        } else {
            // single node cluster
            // vote is granted by the majority of servers, can become leader
            let spec_pools = cst.sps.drain().collect();
            let mut log_w = RwLockUpgradableReadGuard::upgrade(log);
            self.recover_from_spec_pools(st, &mut log_w, spec_pools);
            self.become_leader(st);
            None
        }
```

curp/src/server/raw_curp/mod.rs
```diff
@@ -259,6 +262,14 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
             }
         });
 
+        // check if commit_index needs to be updated
+        if self.can_update_commit_index_to(&log_w, index, self.term()) && index > log_w.commit_index
+        {
+            log_w.commit_index = index;
+            debug!("{} updates commit index to {index}", self.id());
+            self.apply(&mut *log_w);
```
Phoenix500526, line 270: Why should we apply log in `handle_propose`?

curp/src/server/raw_curp/mod.rs
```diff
@@ -797,7 +808,12 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
 // Don't grab lock in the following functions(except cb or sp's lock)
 impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
     /// Server becomes a candidate
-    fn become_candidate(&self, st: &mut State, cst: &mut CandidateState<C>, log: &Log<C>) -> Vote {
+    fn become_candidate(
+        &self,
+        st: &mut State,
+        cst: &mut CandidateState<C>,
+        mut log: RwLockUpgradableReadGuard<'_, Log<C>>,
```
bsbds, line 815: Why should a lock guard be held throughout the entire function, instead of passing the lock directly and lock it when needed?

curp/src/server/raw_curp/mod.rs
themanforfree: special case for single node cluster

curp/src/server/raw_curp/mod.rs
themanforfree: because the lock may use in different level of method, pass the lock directly may require multiple locks, it's difficult to guarantee the locking order, more prone to deadlocks

curp/src/server/raw_curp/mod.rs
themanforfree: Fixed

curp/src/client.rs
themanforfree: Fixed

xline/src/server/kv_server.rs
themanforfree: Instrument is not suitable for putting longer data, because each `event` in `span` will print these contents

curp/src/client.rs
themanforfree: Fixed

Phoenix500526 (approved):

iGxnon (approved):

PR #347, opened 2023-06-28
Update pre commit config

Phoenix500526:
1. forbidden to commit directly to the main or master branch
2. run unit tests before every commit 


themanforfree:
Can we check the DCO signature and the commit message format in pre commit

Phoenix500526:
@themanforfree， the pre-commit hook is executed before committing. We don't have any commit messages yet at that time, so we can not check DCO and validate commit messages at the stage of pre-commit.

themanforfree (approved):

iGxnon (approved):

PR #349, opened 2023-06-29
chore: add commit-msg hook to validate commit message

Phoenix500526:
add commit message hook to validate commit messages locally

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 fail_fast: false
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v4.4.0
     hooks:
       - id: no-commit-to-branch
       - id: check-byte-order-marker
@@ -14,7 +14,7 @@ repos:
       - id: mixed-line-ending
       - id: trailing-whitespace
   - repo: https://github.com/psf/black
-    rev: 19.3b0
+    rev: 23.3.0
     hooks:
       - id: black
   - repo: local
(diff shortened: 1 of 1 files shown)


Phoenix500526:
please use `pre-commit install --hook-type commit-msg --hook-type pre-commit` to update the pre-commit config after this pr merges into the master.

iGxnon (approved):

.pre-commit-config.yaml
```diff
@@ -47,3 +47,15 @@ repos:
         language: rust
         files: \.rs$
         pass_filenames: false
+      - id: conventional-commit-msg-validation
+        name: commit message conventional validation
+        language: pygrep
+        entry: '^(breaking|build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test){1}(\([\w\-\.]+\))?(!)?: ([\w ])+([\s\S]*)'
```
bsbds, line 53: The `([\w ])+` seems only match word characters.

Phoenix500526: The `\w` in regular expressions represents any word characters. It matches any alphanumeric character, like a-z, A-Z, 0-9, and underscore. It's equivalent to the character class`[a-zA-z0-9_]`

bsbds: for example it will not match
```
feat: `foo`
```
maybe `^(breaking|build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test){1}(\([\w\-\.]+\))?(!)?: .+([\s\S]*)` is better?

Phoenix500526: IMO, using the wildcard "." is not a good choice. While it allows you to use backticks, you can also use other symbols as well, like #$%^ etc. How about "^(breaking|build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test){1}(\([\w\-\.]+\))?(!)?: ([\w `])+([\s\S]*)"?

bsbds: makes sense

bsbds (approved):

iGxnon (approved):

Issue #267, opened 2023-05-11
[Bug]: The test_no_root_user_do_admin_ops will pass despite the panic


Phoenix500526:
### Description about the bug

As the title described, the test case `test_no_root_user_do_admin_ops` will panic, but it still passes the test. This is not a good practice.  

### Version

0.3.0

### Relevant log output

```shell
running 1 test
thread 'tokio-runtime-worker' panicked at 'propose err SyncedError("ExecuteError(\"permission denied\")")', /home/jiawei/open-source/Xline/xline/src/server/auth_server.rs:108:21
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
   2: xline::server::auth_server::AuthServer<S>::propose::{{closure}}::{{closure}}
             at ./src/server/auth_server.rs:108:21
   3: core::result::Result<T,E>::map_err
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/result.rs:861:27
   4: xline::server::auth_server::AuthServer<S>::propose::{{closure}}
             at ./src/server/auth_server.rs:104:39
   5: xline::server::auth_server::AuthServer<S>::handle_req::{{closure}}
             at ./src/server/auth_server.rs:145:71
   6: <xline::server::auth_server::AuthServer<S> as xline::rpc::etcdserverpb::auth_server::Auth>::user_add::{{closure}}
             at ./src/server/auth_server.rs:234:40
   7: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/future/future.rs:124:9
   8: <<xline::rpc::etcdserverpb::auth_server::AuthServer<T> as tower_service::Service<http::request::Request<B>>>::call::UserAddSvc<T> as tonic::server::service::UnaryService<xline::rpc::etcdserverpb::AuthUserAddRequest>>::call::{{closure}}
             at /home/jiawei/open-source/Xline/target/debug/build/xline-f0729069f3b92904/out/etcdserverpb.rs:4259:78
   9: <core::pin::Pin<P> as core::future::future::Future>::poll
[...]

PR #355, opened 2023-06-30
Feature/Abort on panic macro

bsbds:
Closes: #267

   This PR add a new `abort_on_panic` macro. It adds a panic hook at the beginning of a funciton, which can be used in tokio tests to force test to fail when a tokio task panic. 

    This is more of a temporary solution as there is no way tokio could handle a panic in tasks properly unless the task is joined.

### DIFF
--- a/test-macros/Cargo.toml
+++ b/test-macros/Cargo.toml
@@ -6,7 +6,7 @@ description = "Macros for test configurations"
 keywords = ["Test", "Macros"]
 license = "Apache-2.0"
 readme = "README.md"
-repository = "https://github.com/datenlord/Xline/tree/master/test-macros"
+repository = "https://github.com/xline-kv/Xline/tree/master/test-macros"
 version = "0.1.0"
 edition = "2021"
 
--- a/test-macros/Cargo.toml
+++ b/test-macros/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "test-macros"
+authors = ["DatenLord <dev@datenlord.io>"]
+categories = ["Macros"]
+description = "Macros for test configurations"
+keywords = ["Test", "Macros"]
+license = "Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/datenlord/Xline/tree/master/test-macros"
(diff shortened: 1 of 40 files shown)


test-macros/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "test-macros"
+authors = ["DatenLord <dev@datenlord.io>"]
+categories = ["Macros"]
+description = "Macros for test configurations"
+keywords = ["Test", "Macros"]
+license = "Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/datenlord/Xline/tree/master/test-macros"
```
iGxnon, line 9: Please replace `datenlord` with `xline-kv` here

test-macros/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+use syn::{parse_macro_input, ItemFn, Stmt};
+
+#[proc_macro_attribute]
+pub fn abort_on_panic(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut input: syn::ItemFn = parse_macro_input!(item as ItemFn);
+
+    let panic_hook: Stmt = syn::parse_quote! {
+        std::panic::set_hook(Box::new(|info| {
+            let stacktrace = std::backtrace::Backtrace::force_capture();
+            println!("Got panic. @info:{}\n@stackTrace:{}", info, stacktrace);
```
iGxnon, line 12:
```suggestion
            println!("Got panic. @info:{}\n@stackTrace:\n{}", info, stacktrace);
```
Better for reading

test-macros/Cargo.toml
bsbds: done

test-macros/src/lib.rs
bsbds: done

iGxnon (dismissed):

Phoenix500526 (approved):

iGxnon (approved):

Issue #361, opened 2023-07-05
[Bug]: Madsim Curp integration tests sometimes fail


bsbds:
### Description about the bug

The Curp integration tests in `simulation` may fail sometimes, and setting `MADSIM_TEST_SEED` may not reproduce the error.

### Version

0.1.0

### Relevant log output

```shell
test curp::server_recovery::all_crash_and_recovery ... FAILED

failures:

---- curp::server_recovery::all_crash_and_recovery stdout ----
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `[]`,
 right: `[0]`', simulation/tests/it/curp/server_recovery.rs:388:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
context: node=0 "main", task=0 (spawned at /home/gum/.cargo/git/checkouts/madsim-e41dc49cac5b4fae/6753d73/madsim/src/sim/runtime/mod.rs:126:19)
note: run with `MADSIM_TEST_SEED=1688527910644819540` environment variable to reproduce this error
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #363, opened 2023-07-05
fix: resolve failing madsim tests

bsbds:

    Described in : #361

    1. change `all_crash_and_recovery` to `minority_crash_and_recovery`. This test does not match the assumption of current implementation(that the majority of nodes should not crash), therefore lead to a test failure.
    2. refactor some assumptions based on time synchronization, makes these test runs more stable under madsim simulated time.

### DIFF
--- a/simulation/src/curp_group.rs
+++ b/simulation/src/curp_group.rs
@@ -257,13 +257,13 @@ impl CurpGroup {
     }
 
     pub async fn get_leader(&self) -> (ServerId, u64) {
-        for _ in 0..5 {
+        loop {
             if let Some(leader) = self.try_get_leader().await {
                 return leader;
             }
             debug!("failed to get leader");
+            madsim::time::sleep(Duration::from_millis(100)).await;
         }
-        panic!("can't get leader");
     }
 
     // get latest term and ensure every working node has the same term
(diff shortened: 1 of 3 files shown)


simulation/src/curp_group.rs
```diff
@@ -257,13 +257,13 @@ impl CurpGroup {
     }
 
     pub async fn get_leader(&self) -> (ServerId, u64) {
-        for _ in 0..5 {
+        loop {
```
Phoenix500526, line 260: We should add a timeout mechanism to this loop to prevent endless waiting in some rare cases.

bsbds: As the time is simulated, it's hard to give a specific timeout for the waiting. In my testing, even a long duraion of waiting may fail to get the leader.

It seems the client will always get the leader, as long as the cluster is able to conduct a successful election?

simulation/tests/it/curp/server_election.rs
```diff
@@ -1,13 +1,13 @@
...
 use utils::config::ClientTimeout;
 
 use simulation::curp_group::CurpGroup;
 
-async fn wait_for_election() {
-    sleep(Duration::from_secs(3)).await;
+/// Wait some time for the election to finish, and get the leader to ensure that the election is
+/// completed.
+async fn wait_for_election(group: &CurpGroup) -> (String, u64) {
+    sleep_secs(15).await;
```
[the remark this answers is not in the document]
bsbds: Sleeping is rather cheap in madsim simulation. With this the method will get the leader at it first try, hopefully;)

themanforfree (approved):

iGxnon (approved):

PR #367, opened 2023-07-06
refactor: read only txn request will use read state mechanism

themanforfree:

Let read-only `TxnRequest` be handled like a normal `RangeRequest`
Add a new branch to the processing of `TxnRequest`
no


xline/src/server/kv_server.rs
```diff
@@ -470,19 +467,34 @@ where
...
-            Self::update_header_revision(&mut res, revision);
-        }
+        let res = if txn_req.is_read_only() {
+            debug!("TxnRequest is read only");
+            let is_serializable = txn_req.is_serializable();
+            let token = get_token(request.metadata());
+            let wrapper = RequestWithToken::new_with_token(request.into_inner().into(), token);
+            let propose_id = self.generate_propose_id();
+            let cmd = Self::command_from_request_wrapper(propose_id, wrapper);
+            if is_serializable {
```
bsbds, line 478: should this be `if !is_serializable`?

xline/src/server/kv_server.rs
```diff
@@ -401,7 +393,12 @@ where
         if !is_serializable {
             self.wait_read_state(&cmd).await?;
         }
-        self.serializable_range(cmd.request())
+        let res = self.do_serializable(cmd.request())?;
+        if let Response::ResponseRange(response) = res {
+            Ok(tonic::Response::new(response))
+        } else {
+            panic!("Receive wrong response {res:?} for RangeRequest");
```
Phoenix500526, line 400: Does `unreachable!` make more sense?

xline/src/server/kv_server.rs
```diff
@@ -470,19 +467,34 @@ where
...
+                debug!("Get revision {:?} for TxnRequest", revision);
+                Self::update_header_revision(&mut res, revision);
+            }
+            res
+        };
         if let Response::ResponseTxn(response) = res {
             Ok(tonic::Response::new(response))
         } else {
-            panic!("Receive wrong response {res:?} for TxnRequest");
+            panic!("Receive wrong response {res:?} for TxnResponse");
```
Phoenix500526, line 497: Same as above.

xline/tests/kv_test.rs
```diff
@@ -249,3 +249,55 @@ async fn test_kv_delete() -> Result<(), Box<dyn Error>> {
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 10)]
```
Phoenix500526, line 253: Why should we set `worker_threads` as 10? Its default value is the number of cpus on the system, so I guess keeping it as default is better.

xline/src/server/kv_server.rs
Phoenix500526: How come? We should do `wait_read_state` when a client sends us a serializable range request.

themanforfree: Fixed， it should be `if !is_serializable`, we only need to do wait_read_state when the request is not a serializable request

xline/src/server/kv_server.rs
themanforfree: Fixed

xline/src/server/kv_server.rs
themanforfree: Fixed

xline/tests/kv_test.rs
themanforfree: To unify with other tests, now I have unified all the tests without set `worker_threads`

xline/src/server/kv_server.rs
```diff
@@ -470,19 +467,34 @@ where
...
+            let token = get_token(request.metadata());
+            let wrapper = RequestWithToken::new_with_token(request.into_inner().into(), token);
+            let propose_id = self.generate_propose_id();
+            let cmd = Self::command_from_request_wrapper(propose_id, wrapper);
+            if !is_serializable {
+                self.wait_read_state(&cmd).await?;
+            }
+            self.do_serializable(cmd.request())?
+        } else {
+            let is_fast_path = false; // lock need revision of txn
```
bsbds, line 483: If we return the revision in execution result, can this be changed to use fast path

themanforfree: Yes, you can change this in #391

Phoenix500526 (commented):
Let’s assume that there are two committed log entries, `PUT A 1` and `PUT B 1`,  in our state machine log. Now, user_1 proposes a TXN read request (READ A and B) at moment A, while user_2 proposes a put request ate moment B. The moment A is before moment B, and their interval is very short. After user_1 finishes his request, what will he read? (A, B) = (1, 1) or (A, B) = (1, 2)?  It depends on the execution request of `wait_read_state`. Does it align with our initial expectations for the TXN?

xline/src/server/kv_server.rs
```diff
@@ -534,22 +533,37 @@ where
         debug!("Receive grpc request: {:?}", txn_req);
         Self::check_txn_request(
             txn_req,
-            self.kv_storage.compacted_revision(),
+            self.kv_storage.revision(),
```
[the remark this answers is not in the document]
themanforfree: Fixed

Issue #377, opened 2023-07-13
[Bug]: `test_kv_authorization` run unreachable code


themanforfree:
### Description about the bug

`propose` return `SyncedError("ExecuteError(\"permission denied\")")`, but the permission check should be in the `prepare` stage, not `after_sync`.

```rust
let cmd_res = self.client.propose(cmd).await.map_err(|err| match err {
    ProposeError::ExecutionError(e) => tonic::Status::invalid_argument(e),
    _ => unreachable!("propose err {err:?}"),
})?;
```

### Version

0.4.1 (Default)

### Relevant log output

```shell
1.462102583s DEBUG curp::server::cmd_worker::conflict_checked_mpmc: new ce event: ASReady(ProposeId("server0-3b5232b4-5cec-415a-8fc9-b8121e7111fc"), 16)
   1.462124653s DEBUG curp::server::spec_pool: cmd("server0-8c2e0075-bcc8-4c78-b0b9-c3f2b20f86ab") is not in spec pool
   1.462131213s DEBUG curp::server::spec_pool: cmd("server0-8c2e0075-bcc8-4c78-b0b9-c3f2b20f86ab") is not in spec pool
   1.462155783s DEBUG curp::server::cmd_worker: server1 cmd("server0-8c2e0075-bcc8-4c78-b0b9-c3f2b20f86ab") after sync is called
   1.462174893s DEBUG curp::server::cmd_worker: server2 cmd("server0-8c2e0075-bcc8-4c78-b0b9-c3f2b20f86ab") after sync is called
   1.462333093s DEBUG curp::server::raw_curp::state: follower server1's match_index updated to 16
`test_kv_authorization` panic! 
@info:
panicked at 'internal error: entered unreachable code: propose err cmd("server0-3b5232b4-5cec-415a-8fc9-b8121e7111fc") SyncedError("ExecuteError(\"permission denied\")")', /home/guanyu/Xline/xline/src/server/kv_server.rs:145:22
@stackTrace:
   0: auth_test::test_kv_authorization::{{closure}}
             at ./tests/auth_test.rs:94:1
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/boxed.rs:1987:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:695:13
[...]

Issue #376, opened 2023-07-13
[Bug]: `CurpGroup` related tests panic


themanforfree:
### Description about the bug

`basic_propose`, `concurrent_cmd_order`, etc. the tests  that depends in `CurpGroup` maybe panic.

### Version

0.4.1 (Default)

### Relevant log output

```shell
...
`concurrent_cmd_order` panic! 
@info:
panicked at 'failed to send exe msg: SendError((TestCommand { id: ProposeId("1"), keys: [0], exe_dur: 1s, as_dur: 0ns, exe_should_fail: false, as_should_fail: false, cmd_type: Put(0) }, ([], [])))', curp-test-utils/src/test_cmd.rs:234:14
@stackTrace:
   0: server::concurrent_cmd_order::{{closure}}
             at ./tests/server.rs:147:1
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/boxed.rs:1987:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:695:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:582:13
...
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Phoenix500526:
So, what's the root cause of this issue?

PR #378, opened 2023-07-13
fix `CurpGroup` related tests

themanforfree:
#376
#377 

### DIFF
--- a/curp/tests/common/curp_group.rs
+++ b/curp/tests/common/curp_group.rs
@@ -40,7 +40,6 @@ pub struct CurpNode {
     pub exe_rx: mpsc::UnboundedReceiver<(TestCommand, TestCommandResult)>,
     pub as_rx: mpsc::UnboundedReceiver<(TestCommand, LogIndex)>,
     pub store: Arc<Engine>,
-    pub rt: Runtime,
     pub storage_path: PathBuf,
     pub role_change_arc: Arc<TestRoleChangeInner>,
 }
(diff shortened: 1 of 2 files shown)


curp/tests/common/curp_group.rs
```diff
@@ -40,7 +40,6 @@ pub struct CurpNode {
     pub exe_rx: mpsc::UnboundedReceiver<(TestCommand, TestCommandResult)>,
     pub as_rx: mpsc::UnboundedReceiver<(TestCommand, LogIndex)>,
     pub store: Arc<Engine>,
-    pub rt: Runtime,
```
bsbds, line 43: I think a seperate runtime is needed to shutdown the RPC server when dropped. Otherwise the server may still writing to the files after `stop` is called.

curp/tests/common/curp_group.rs
```diff
@@ -79,34 +78,25 @@ impl CurpGroup {
...
-                            CurpConfigBuilder::default()
-                                .data_dir(storage_path_c)
-                                .log_entries_cap(10)
-                                .build()
-                                .unwrap(),
-                        ),
-                    ));
-                });
+
+                tokio::spawn(Rpc::run_from_listener(
```
Phoenix500526, line 85: Why do you remove the runtime here? I guess using runtime here is for renaming a tokio 
 thread which may be beneficial to debug.

xline/src/server/kv_server.rs
```diff
@@ -140,7 +140,13 @@ where
         #[allow(clippy::wildcard_enum_match_arm)]
         if use_fast_path {
             let cmd_res = self.client.propose(cmd).await.map_err(|err| match err {
-                ProposeError::ExecutionError(e) => tonic::Status::invalid_argument(e),
+                // If an error occurs when `prepare` or `execute`, `after_sync` will not be called, in this case,
```
Phoenix500526, line 143: This comment looks a little bit weird. Is it a grammar mistake? How about this?
```suggestion
                // If an error occurs during the `prepare` or `execute` stages, `after_sync` will not be invoked. In this case, 
                // `wait_synced` will return the errors generated in the first two stages. Therefore, 
                // if the response from `slow_round` arrives earlier than `fast_round`, the `propose` 
                // function will return a `SyncedError`, even though `after_sync` is not called.
```

themanforfree:
#396

PR #311, opened 2023-06-07
feat/compactor

Phoenix500526:
Read #188 for more details.

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #389, opened 2023-07-20
feat: implement client compact

bsbds:
based on: #311

    The compact method in `xline-client` can be implement after #311

### DIFF
--- a/xline-client/src/types/kv.rs
+++ b/xline-client/src/types/kv.rs
@@ -674,3 +674,51 @@ impl From<Txn> for xlineapi::TxnRequest {
         txn.req
     }
 }
+
+/// Compaction Request compacts the key-value store up to a given revision.
+/// All superseded keys with a revision less than the compaction revision will be removed.
+#[derive(Debug)]
+pub struct CompactionRequest {
+    /// The inner request
+    inner: xlineapi::CompactionRequest,
+}
+
+impl CompactionRequest {
+    /// Creates a new `CompactionRequest`
+    ///
+    /// `Revision` is the key-value store revision for the compaction operation.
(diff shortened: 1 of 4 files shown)


Phoenix500526:
Is this pr based on #399 ？

xline-client/src/types/kv.rs
```diff
@@ -674,3 +674,51 @@ impl From<Txn> for xlineapi::TxnRequest {
         txn.req
     }
 }
+
+/// Compaction Request compacts the key-value store up to a given revision.
+/// All superseded keys with a revision less than the compaction revision will be removed.
```
Phoenix500526, line 679: Compaction is not a simple deletion. It won't remove the latest revisions of keys even if their revisions are less than the compaction revision.

bsbds: an extra note is added

Phoenix500526: It's "Note that", not "Not that". You can use this as a comment: 
```suggestion
/// All keys with revisions less than the given revision will be compacted. The compaction process will remove all historical versions of these keys, except for the most recent one.  

/// For example, here is a revision list: [(A, 1), (A, 2), (A, 3), (A, 4), (A,5)]. We compact at revision 3. After the compaction, the revision list will become [ (A, 3), (A, 4), (A,5)]. All revisions less than 3 are deleted.  The latest revision, 3, will be kept. 
```

bsbds: Sorry for the typo. the comment is updated

Phoenix500526 (approved):

iGxnon (approved):

Issue #324, opened 2023-06-13
TXN optimization


themanforfree:
The following is the benchmark data of xline and etcd process txn put requests

benchmark command: 
``` bash
benchmark --clients=${clients} txn-put --key-size=8 --val-size=256 --total=${total} --key-space-size=${key-space-size}
```

### Xline
| QPS    | Latency(ms) | Time(s) | clients | total | key-space-size |
| ------ | ----------- | ------- | ------- | ----- | -------------- |
| 324.3  | 3.1         | 6.2     | 1       | 2000  | 1              |
| 2768.4 | 3.6         | 3.6     | 10      | 10000 | 1              |
| 5429.5 | 9.2         | 3.7     | 50      | 20000 | 1              |
| 5498.6 | 18.1        | 3.6     | 100     | 20000 | 1              |
| 5485.5 | 36.3        | 5.5     | 200     | 30000 | 1              |
| 331.5  | 3.0         | 6.0     | 1       | 2000  | 100000         |
| 2700.4 | 3.7         | 3.7     | 10      | 10000 | 100000         |
| 5470.5 | 9.1         | 3.7     | 50      | 20000 | 100000         |
| 5563.0 | 17.9        | 3.6     | 100     | 20000 | 100000         |
| 5530.4 | 36.1        | 5.4     | 200     | 30000 | 100000         |

### etcd
| QPS     | Latency(ms) | Time(s) | clients | total | key-space-size |
| ------- | ----------- | ------- | ------- | ----- | -------------- |
| 935.9   | 1.1         | 2.1     | 1       | 2000  | 1              |
| 3341.5  | 3.0         | 3.0     | 10      | 10000 | 1              |
| 12219.3 | 4.1         | 1.6     | 50      | 20000 | 1              |
| 25641.9 | 3.9         | 0.8     | 100     | 20000 | 1              |
| 16737.8 | 11.9        | 1.8     | 200     | 30000 | 1              |
| 812.7   | 1.2         | 2.5     | 1       | 2000  | 100000         |
| 4034.8  | 2.5         | 2.5     | 10      | 10000 | 100000         |
| 13525.4 | 3.7         | 1.5     | 50      | 20000 | 100000         |
| 24238.7 | 4.1         | 0.8     | 100     | 20000 | 100000         |
| 30863.9 | 6.5         | 1.0     | 200     | 30000 | 100000         |

---

In our current implementation of Txn, there are some optimized parts, as follows:

[...]

PR #393, opened 2023-07-21
Perf/txn

Phoenix500526:
Improve txn put performance. Read issue #324 for more details

### DIFF
--- a/xline/src/storage/index.rs
+++ b/xline/src/storage/index.rs
@@ -1,6 +1,5 @@
-use std::collections::BTreeMap;
-
 use clippy_utilities::OverflowArithmetic;
+use crossbeam_skiplist::SkipMap;
 use itertools::Itertools;
 use parking_lot::RwLock;
 use utils::parking_lot_lock::RwLockMap;
@@ -12,23 +11,14 @@ use crate::server::command::{KeyRange, RangeType};
 #[derive(Debug)]
 pub(crate) struct Index {
     /// Inner struct of `Index`
-    inner: RwLock<IndexInner>,
-}
-
-/// Inner struct of `Index`
-#[derive(Debug)]
-struct IndexInner {
-    /// index
-    index: BTreeMap<Vec<u8>, Vec<KeyRevision>>,
+    inner: SkipMap<Vec<u8>, RwLock<Vec<KeyRevision>>>,
 }
 
 impl Index {
     /// New `Index`
     pub(crate) fn new() -> Self {
         Self {
-            inner: RwLock::new(IndexInner {
-                index: BTreeMap::new(),
-            }),
+            inner: SkipMap::new(),
         }
     }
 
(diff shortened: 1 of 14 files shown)


xline/src/storage/index.rs
```diff
@@ -11,41 +12,38 @@ use crate::server::command::{KeyRange, RangeType};
 #[derive(Debug)]
 pub(crate) struct Index {
     /// Inner struct of `Index`
-    inner: Mutex<IndexInner>,
+    inner: RwLock<IndexInner>,
```
bsbds, line 15: we could consider using a skiplist to achieve lockfree concurrent access https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/

xline/src/storage/index.rs
```diff
@@ -185,89 +173,82 @@ impl IndexOperate for Index {
...
-    ) -> KeyRevision {
-        let mut inner = self.inner.lock();
-        let new_rev = if let Some(revisions) = inner.index.get_mut(key) {
-            if let Some(rev) = revisions.last() {
-                let new_rev = if rev.is_deleted() {
-                    KeyRevision::new(revision, 1, revision, sub_revision)
+    fn insert(&self, key_revisions: Vec<(Vec<u8>, KeyRevision)>) {
+        self.inner.map_write(|mut inner| {
+            for (key, revision) in key_revisions {
+                if let Some(revisions) = inner.index.get_mut(&key) {
```
bsbds, line 222: here we could use `map.entry().or_insert()` syntax

xline/src/storage/kv_store.rs
```diff
@@ -366,7 +361,7 @@ where
     ) -> Result<(Vec<KeyValue>, usize), ExecuteError> {
         let mut revisions = self.index.get(key, range_end, revision);
         let total = revisions.len();
-        if count_only {
+        if count_only || 0 == total {
```
themanforfree, line 364: `0 == total` looks weird, how about `total == 0`

xline/src/storage/index.rs
```diff
@@ -134,47 +112,57 @@ pub(super) trait IndexOperate {
...
+                    .collect()
+            }),
+            RangeType::Range => self.inner.map_read(|inner| {
+                inner
+                    .index
+                    .range(KeyRange::new(key, range_end))
+                    .filter_map(|(_k, revs)| Self::get_revision(revs, revision))
+                    .collect()
+            }),
         }
```
themanforfree, line 138: Please move `map_read` to the outside of the `match` block to reduce the duplication

xline/src/storage/index.rs
```diff
@@ -134,47 +112,57 @@ pub(super) trait IndexOperate {
...
+                    .collect()
+            }),
+            RangeType::Range => self.inner.map_read(|inner| {
+                inner
+                    .index
+                    .range(KeyRange::new(key, range_end))
+                    .flat_map(|(_k, revs)| Self::filter_revision(revs, revision))
+                    .sorted()
+                    .collect()
+            }),
```
themanforfree, line 165: Same as above

xline/src/storage/kv_store.rs
Phoenix500526: Done

xline/src/storage/index.rs
Phoenix500526: Done

xline/src/storage/index.rs
Phoenix500526: Ok, I'll do some researches and do it in another pr

xline/src/storage/index.rs
Phoenix500526: The `get` method is a hotspot. We need to keep the critical section as small as possible to minimize lock contention as possible.  The `match` block and `get_range_type` methods do not require holding the lock, which is why I removed them from the `map_read` block. As for reducing duplication, I tried making the match block return a closure - `FnOnce(RwLockReadGuard<'_, IndexInner>) -> Vec<Revision>`. This would reduce some duplication but would require using `Box<dyn FnOnce(RwLockReadGuard<'_, IndexInner>) -> Vec<Revision>>` to wrap the closure, which would result in a performance loss. After weighing the pros and cons, I still believe that using this approach is better. Although there is some seemingly duplicated code, the readability is not that bad. Additionally, the type of Key appears to be relatively stable and unlikely to change, so it is less likely that a modification in one place would be forgotten to be modified in another copy. Any thoughts of this? @rogercloud

xline/src/storage/index.rs
Phoenix500526: Same as above

xline/src/storage/index.rs
```diff
@@ -205,72 +197,71 @@ impl IndexOperate for Index {
...
-                };
-                revisions.push(new_rev);
-                new_rev
+    fn insert(&self, key_revisions: Vec<(Vec<u8>, KeyRevision)>) {
+        for (key, revision) in key_revisions {
+            if let Some(entry) = self.inner.get::<[u8]>(key.as_ref()) {
+                entry.value().map_write(|mut revs| revs.push(revision));
             } else {
-                panic!("Get empty revision list for key {key:?}");
+                _ = self.inner.insert(key, RwLock::new(vec![revision]));
```
themanforfree, line 239: Could we use `get_or_insert`?

xline/src/storage/index.rs
```diff
@@ -281,48 +272,52 @@ impl IndexOperate for Index {
...
+                revisions.push(KeyRevision::new(
+                    create_revision,
+                    version,
+                    revision,
+                    sub_revision,
+                ));
+            });
     }
 
     fn compact(&self, at_rev: i64) -> Vec<KeyRevision> {
```
bsbds, line 288: Is it possible that `insert` is called after `delete` that the old revisions may never be compacted.

xline/src/storage/index.rs
Phoenix500526: There is no need to do so because even if the `get_or_insert` success succeeds, you still have to check whether the returned value contains the given revision. If the given revision doesn’t exist, you still need to push it manually. Therefore, it won’t save much code.

xline/src/storage/index.rs
bsbds: my mistaken, resolved.

iGxnon (approved):

Issue #80, opened 2022-11-16
[Xline] check permissions in a more efficient way


themanforfree:
In the current implementation, we traverse a `KeyRange` vector that may have duplicates

PR #394, opened 2023-07-21
refactor: request validation check

bsbds:
Described in: #398
Based on: #399

    Currently some of the validation checks are done in the RPC server. However, these checks are not applied in the storage request handlers (eg. `handle_range_request`), so that the request from outside client will not be checked.

    This pull request applied the validation check in both RPC server and the storage handlers.

### DIFF
--- a/xline-client/tests/kv.rs
+++ b/xline-client/tests/kv.rs
@@ -254,16 +254,13 @@ async fn compact_should_remove_previous_revision() -> Result<()> {
         .await?;
     assert_eq!(rev1_resp.kvs[0].value, b"1");
 
-    client.compact(CompactionRequest::new(4)).await?;
+    client.compact(CompactionRequest::new(3)).await?;
 
     // after compacting
     let rev0_resp = client
         .range(RangeRequest::new("compact").with_revision(2))
-        .await?;
-    assert!(
-        rev0_resp.kvs.is_empty(),
-        "kvs should be empty after compaction"
-    );
+        .await;
+    assert!(rev0_resp.is_err(), "should be err after compaction");
     let rev1_resp = client
         .range(RangeRequest::new("compact").with_revision(3))
         .await?;
--- a/xline-client/tests/kv.rs
+++ b/xline-client/tests/kv.rs
@@ -260,7 +260,10 @@ async fn compact_should_remove_previous_revision() -> Result<()> {
     let rev0_resp = client
         .range(RangeRequest::new("compact").with_revision(2))
         .await;
-    assert!(rev0_resp.is_err(), "should be err after compaction");
+    assert!(
+        rev0_resp.is_err(),
+        "client.range should receive an err after compaction, but it receives: {rev0_resp:?}"
+    );
     let rev1_resp = client
         .range(RangeRequest::new("compact").with_revision(3))
         .await?;
--- a/xline/src/lib.rs
+++ b/xline/src/lib.rs
@@ -161,6 +161,8 @@ mod revision_number;
 mod rpc {
     pub use xlineapi::*;
 }
+/// Request validation module
+mod request_validation;
 /// Xline server
 pub mod server;
 /// State of current node
--- a/xline/src/lib.rs
+++ b/xline/src/lib.rs
@@ -163,6 +163,8 @@ mod rpc {
 }
 /// Request validation module
 mod request_validation;
+/// Revision check
+mod revision_check;
 /// Xline server
 pub mod server;
 /// State of current node
--- a/xline/src/request_validation.rs
+++ b/xline/src/request_validation.rs
@@ -153,7 +153,7 @@ pub(crate) fn check_range_compacted(
     range_revision: i64,
     compacted_revision: i64,
 ) -> Result<(), ValidationError> {
-    (range_revision >= compacted_revision)
+    (range_revision >= compacted_revision || range_revision <= 0)
             .then_some(())
             .ok_or(ValidationError::new(format!(
                 "required revision {range_revision} has been compacted, compacted revision is {compacted_revision}"
--- a/xline/src/request_validation.rs
+++ b/xline/src/request_validation.rs
@@ -4,9 +4,8 @@ use thiserror::Error;
 
 use crate::{
     rpc::{
-        AuthRoleAddRequest, AuthRoleGrantPermissionRequest, AuthUserAddRequest, CompactionRequest,
-        DeleteRangeRequest, PutRequest, RangeRequest, Request, RequestOp, SortOrder, SortTarget,
-        TxnRequest,
+        AuthRoleAddRequest, AuthRoleGrantPermissionRequest, AuthUserAddRequest, DeleteRangeRequest,
+        PutRequest, RangeRequest, Request, RequestOp, SortOrder, SortTarget, TxnRequest,
     },
     server::KeyRange,
     storage::ExecuteError,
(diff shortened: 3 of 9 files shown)


xline/src/storage/execute_error.rs
```diff
@@ -26,6 +26,11 @@ impl ExecuteError {
         Self::KvError("key not found".to_owned())
     }
 
+    /// Key not found
```
Phoenix500526, line 29: Please modify the comment. And why should we consider an invalid request error as a kind of KvError?

bsbds: the code is now refactored, please refer to #398

xline/src/storage/kv_store.rs
```diff
@@ -840,6 +865,72 @@ where
...
+                    }
+                }
+                Ok(())
+            }
+            UniRequest::RequestPut(_) | UniRequest::RequestDeleteRange(_) => Ok(()),
+        }
+    }
+
+    /// Validate compact request before handle
+    fn check_compaction_request(
```
Phoenix500526, line 909: We can merge `check_compaction_request` and `check_revision` together.

xline/src/storage/kv_store.rs
```diff
@@ -496,6 +498,13 @@ where
 
     /// Handle `RangeRequest`
     fn handle_range_request(&self, req: &RangeRequest) -> Result<RangeResponse, ExecuteError> {
+        req.validation()?;
+        Self::check_revision(
+            &UniRequest::RequestRange(req.clone()),
+            self.compacted_revision(),
+            self.revision(),
+        )?;
```
themanforfree, line 506: We already validated it in `kv_server`, Is it necessary to validate again here

xline/src/storage/kv_store.rs
```diff
@@ -545,6 +554,8 @@ where
 
     /// Handle `PutRequest`
     fn handle_put_request(&self, req: &PutRequest) -> Result<PutResponse, ExecuteError> {
+        req.validation()?;
```
themanforfree, line 557: Same as above

xline/src/storage/kv_store.rs
```diff
@@ -566,6 +577,8 @@ where
         &self,
         req: &DeleteRangeRequest,
     ) -> Result<DeleteRangeResponse, ExecuteError> {
+        req.validation()?;
```
themanforfree, line 580: Same as above

xline/src/storage/kv_store.rs
```diff
@@ -580,6 +593,13 @@ where
 
     /// Handle `TxnRequest`
     fn handle_txn_request(&self, req: &TxnRequest) -> Result<TxnResponse, ExecuteError> {
+        req.validation()?;
+        Self::check_revision(
+            &UniRequest::RequestTxn(req.clone()),
+            self.compacted_revision(),
+            self.revision(),
+        )?;
+
```
themanforfree, line 602: Same as above

xline/src/storage/kv_store.rs
```diff
@@ -602,16 +622,21 @@ where
     }
 
     /// Handle `CompactionRequest`
-    fn handle_compaction_request(&self, req: &CompactionRequest) -> CompactionResponse {
+    fn handle_compaction_request(
+        &self,
+        req: &CompactionRequest,
+    ) -> Result<CompactionResponse, ExecuteError> {
+        Self::check_compaction_request(req, self.compacted_revision(), self.revision())?;
```
themanforfree, line 629: Same as above

xline/src/server/auth_server.rs
```diff
@@ -206,18 +207,7 @@ where
...
-        let need_password = user_add_req
-            .options
-            .as_ref()
-            .map_or(true, |o| !o.no_password);
-        if need_password && user_add_req.password.is_empty() {
-            return Err(tonic::Status::invalid_argument(
-                "password is required but not provided",
-            ));
-        }
+        user_add_req.validation()?;
```
iGxnon, line 210: Can we directly move requests validations that are stateless into xline-client?
This way, we won't need to perform double checks at both the Store and Server.

xline/src/storage/kv_store.rs
```diff
@@ -496,6 +498,13 @@ where
 
     /// Handle `RangeRequest`
     fn handle_range_request(&self, req: &RangeRequest) -> Result<RangeResponse, ExecuteError> {
+        req.validation()?;
+        Self::check_revision(
+            &UniRequest::RequestRange(req.clone()),
```
iGxnon, line 503: It seems that when we introduce `UniRequest` in `check_revision`, we have to clone the request here.
Can we optimize it? Or we could add a TODO here.

xline-client/tests/kv.rs
```diff
@@ -254,16 +254,13 @@ async fn compact_should_remove_previous_revision() -> Result<()> {
...
     // after compacting
     let rev0_resp = client
         .range(RangeRequest::new("compact").with_revision(2))
-        .await?;
-    assert!(
-        rev0_resp.kvs.is_empty(),
-        "kvs should be empty after compaction"
-    );
+        .await;
+    assert!(rev0_resp.is_err(), "should be err after compaction");
```
rogercloud, line 263: Error message should be "XXXX should be err after compaction". XXXX stands for some actions.

xline/src/request_validation.rs
```diff
@@ -0,0 +1,414 @@
...
+        let _ignore_success = check_intervals(&self.success)?;
+        let _ignore_failure = check_intervals(&self.failure)?;
+
+        Ok(())
+    }
+}
+
+/// Check if puts and deletes overlap
+fn check_intervals(ops: &[RequestOp]) -> Result<(HashSet<&[u8]>, Vec<KeyRange>), ValidationError> {
+    // TODO: use interval tree is better?
```
rogercloud, line 103: Open an issue to track it?

xline/src/request_validation.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    for op in ops {
+        if let Some(Request::RequestTxn(ref req)) = op.request {
+            // handle child txn request
+            let (success_puts, mut success_dels) = check_intervals(&req.success)?;
+            let (failure_puts, mut failure_dels) = check_intervals(&req.failure)?;
+
+            for k in &success_puts {
+                if !puts.insert(k) {
+                    return Err(ValidationError::new("duplicate key given in txn request"));
```
rogercloud, line 125: It's not coverred in the test. Please double check to make sure that every error case is covered in UT.

xline/src/storage/kv_store.rs
bsbds: Currently the validation in kv store is responsible for checking request from xline client which use curp to propose. 

For a etcd client it's might a bit redundant for validating twice, but it may improve performance to validate the reuqest in `kv_server` as we can return immediately if validation has failed, instead of sending to `kv_store` in execution stage and then return a failure.

xline/src/server/auth_server.rs
Phoenix500526: Whether we can do it depends on whether `etcd` validates these stateless requests on its client side. If it doesn't, we have to validate these requests on the server side to be compatible with etcd.

xline/src/server/kv_server.rs
```diff
@@ -185,25 +183,36 @@ where
...
-        } else {
-            Self::check_range_compacted(req.revision, compacted_revision)
+        debug_assert!(
+            compacted_revision <= current_revision,
+            "compacted revision should not larger than current revision"
+        );
+        match *request {
+            UniRequest::RequestRange(ref r) => {
+                if r.revision > current_revision {
+                    Err(tonic::Status::invalid_argument(format!(
```
Phoenix500526, line 199: Here should be `tonic::Status::out_of_range`.

xline/src/server/kv_server.rs
```diff
@@ -632,8 +504,8 @@ mod test {
             range_request_with_future_rev.revision, current_revision
         ))
         .to_string();
-        let message = KvServer::<DB>::check_range_request(
-            &range_request_with_future_rev,
+        let message = KvServer::<DB>::check_revision(
+            &UniRequest::RequestRange(range_request_with_future_rev),
```
Phoenix500526, line 508: We don't need to match the error message. We can match the `tonic::Status::Code` directly

xline/src/lib.rs
```diff
@@ -163,6 +163,8 @@ mod rpc {
 }
 /// Request validation module
 mod request_validation;
+/// Revison check
```
Phoenix500526, line 166: "Revison" should be "Revision".

xline/src/request_validation.rs
Phoenix500526: Is this issue #80 ? @themanforfree

themanforfree: No, we didn't use interval tree now, There was no suitable rust implementation for interval trees when I write this part, so I leave a `TODO` here. and I'm not sure that using this interval tree will be better than the current implementation.

xline/src/storage/kv_store.rs
bsbds: I refactored the revison check code, now it only takes a reference

xline/src/server/auth_server.rs
Phoenix500526: The etcdctl will do some validation before it issues a get command. FYI: https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/get_command.go#L98

xline/src/request_validation.rs
bsbds: I opened this in #409

xline/src/request_validation.rs
bsbds: now it should be covered by 38eab5e0dfb759ba5d36c646f05331f091231172

rogercloud (approved):

themanforfree (approved):

Issue #397, opened 2023-07-26
[Refactor] Execution error types


bsbds:
## Background
In xline storage, we use `ExecuteError` to represent various error types when executing the command.
The current implementation is like this
```rust
/// Error met when executing commands
#[derive(Error, Debug, Clone)]
pub enum ExecuteError {
    /// Kv error
    #[error("kv error: {0}")]
    KvError(String),
    /// Lease error
    #[error("lease error: {0}")]
    LeaseError(String),
    /// Auth error
    #[error("auth error: {0}")]
    AuthError(String),
    /// Db error
    #[error("db error: {0}")]
    DbError(String),
    /// Permission denied
    #[error("permission denied")]
    PermissionDenied,
}

impl ExecuteError {
    /// Key not found
    pub(crate) fn key_not_found() -> Self {
        Self::KvError("key not found".to_owned())
    }

    /// Lease not found
    pub(crate) fn lease_not_found(lease_id: i64) -> Self {
        Self::LeaseError(format!("lease {lease_id} not found"))
    }

    ...
}
```

and in `cmd_worker` we map the error to a string and return it to the client.
```rust
                    ce.execute(cmd.as_ref(), index)
                        .await
                        .map_err(|e| e.to_string())
 ```

This type of error handling is sufficient in server only implementation. However, as we introduced the outside client, we cannot recognize the specific error unless we match the entire error string. Thus it will be much more convenient to have distinct error types of execution result.

### Step 1
We could refactor the `ExecuteError` to more fine grained error typs like:
```rust
pub enum ExecuteError {
    InvalidRequest(String),
    KeyNotFound,
    ...
}
```
### Step 2

Currently the error type is associated to the `CommandExecutor` trait, I suggest to move it to the `Command` trait. As we need to specific the generic in `curp`, it's not nice to always have two generics.
```rust
#[async_trait]
pub trait CommandExecutor<C>: Sync + Send + std::fmt::Debug
where
    C: Command,
{
    /// Error type
    type Error: std::fmt::Debug + Send + Sync + Clone + std::error::Error;
    
    ...
```

### Step 3

Then we just need to refactor the relevant code in `curp` to include the execution error type as part of the `ProposeError`
```rust
pub enum ProposeError<C: Command> {
    ExecutionError(C::Error),
    ...
}
 ```

Phoenix500526:
Resolved in pr #399

PR #399, opened 2023-07-26
refactor: execution error types

bsbds:
Described in: #397 

### DIFF
--- a/curp-external-api/src/cmd.rs
+++ b/curp-external-api/src/cmd.rs
@@ -11,6 +11,9 @@ use crate::LogIndex;
 pub trait Command:
     Sized + Sync + Send + DeserializeOwned + Serialize + std::fmt::Debug + Clone + ConflictCheck
 {
+    /// Error type
+    type Error: Send + Sync + Clone + std::error::Error + Serialize + DeserializeOwned;
+
     /// K (key) is used to tell confliction
     /// The key can be a single key or a key range
     type K: std::fmt::Debug
@@ -40,7 +43,7 @@ pub trait Command:
 
     /// Prepare the command
     #[inline]
-    fn prepare<E>(&self, e: &E, index: LogIndex) -> Result<Self::PR, E::Error>
+    fn prepare<E>(&self, e: &E, index: LogIndex) -> Result<Self::PR, Self::Error>
     where
         E: CommandExecutor<Self> + Send + Sync,
     {
@@ -49,7 +52,7 @@ pub trait Command:
 
     /// Execute the command according to the executor
     #[inline]
-    async fn execute<E>(&self, e: &E, index: LogIndex) -> Result<Self::ER, E::Error>
+    async fn execute<E>(&self, e: &E, index: LogIndex) -> Result<Self::ER, Self::Error>
     where
         E: CommandExecutor<Self> + Send + Sync,
     {
@@ -63,7 +66,7 @@ pub trait Command:
         e: &E,
         index: LogIndex,
         prepare_res: Self::PR,
-    ) -> Result<Self::ASR, E::Error>
+    ) -> Result<Self::ASR, Self::Error>
     where
         E: CommandExecutor<Self> + Send + Sync,
     {
@@ -123,14 +126,11 @@ pub trait CommandExecutor<C>: Sync + Send + std::fmt::Debug
 where
     C: Command,
 {
-    /// Error type
-    type Error: std::fmt::Debug + Send + Sync + Clone + std::error::Error;
-
     /// Prepare the command
-    fn prepare(&self, cmd: &C, index: LogIndex) -> Result<C::PR, Self::Error>;
+    fn prepare(&self, cmd: &C, index: LogIndex) -> Result<C::PR, C::Error>;
 
     /// Execute the command
-    async fn execute(&self, cmd: &C, index: LogIndex) -> Result<C::ER, Self::Error>;
+    async fn execute(&self, cmd: &C, index: LogIndex) -> Result<C::ER, C::Error>;
 
     /// Execute the after_sync callback
     async fn after_sync(
@@ -138,14 +138,14 @@ where
         cmd: &C,
         index: LogIndex,
         prepare_res: C::PR,
-    ) -> Result<C::ASR, Self::Error>;
+    ) -> Result<C::ASR, C::Error>;
 
     /// Index of the last log entry that has been successfully applied to the command executor
-    fn last_applied(&self) -> Result<LogIndex, Self::Error>;
+    fn last_applied(&self) -> Result<LogIndex, C::Error>;
 
     /// Take a snapshot
-    async fn snapshot(&self) -> Result<Snapshot, Self::Error>;
+    async fn snapshot(&self) -> Result<Snapshot, C::Error>;
 
     /// Reset the command executor using the snapshot or to the initial state if None
-    async fn reset(&self, snapshot: Option<(Snapshot, LogIndex)>) -> Result<(), Self::Error>;
+    async fn reset(&self, snapshot: Option<(Snapshot, LogIndex)>) -> Result<(), C::Error>;
 }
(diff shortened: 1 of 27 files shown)


xline/src/storage/execute_error.rs
```diff
@@ -1,134 +1,103 @@
...
-    }
-
+    #[error("root user does not have root role")]
+    RootRoleNotExist,
     /// Permission not granted
-    pub(crate) fn permission_not_granted() -> Self {
-        Self::AuthError("permission not granted to the role".to_owned())
-    }
-
+    #[error("permission not granted to the role")]
```
Phoenix500526, line 62: What are the differences between `PermissionNotGranted` and `PermissionNotGiven`?

xline/src/storage/execute_error.rs
```diff
@@ -1,134 +1,103 @@
...
-        ))
+    #[error("token's revision {0} is older than current revision {1}")]
+    TokenOldRevision(i64, i64),
+
+    /// Db error
+    #[error("db error: {0}")]
+    DbError(String),
+
+    /// Permission denied Error
+    #[error("permission denied")]
```
Phoenix500526, line 88: It looks like there is nothing different between the consequences of `PermissionDenied`、`PermissionNotGiven` and `PermissionNotGranted`. In these scenarios, users are unable to perform their operations. We can keep the `PermissionDenied` error and discard the other two.

bsbds: The PermissionNotGiven is indeed redundant, it's should be a invalid reuqest error, I will fix it in #394. 

The PermissionDenied and PermissionNotGranted actually have different meanings. 
`PermissionDenied` means the user doesn't have the admin permission in one of its roles.
`PermissionNotGranted` means a role does not have a permission to a specific key.

curp/src/client.rs
```diff
@@ -241,8 +241,16 @@ where
                 SyncResult::Error(SyncError::Timeout) => {
                     return Err(ProposeError::SyncedError("wait sync timeout".to_owned()));
                 }
-                SyncResult::Error(e) => {
-                    return Err(ProposeError::SyncedError(format!("{e:?}")));
+                SyncResult::Error(SyncError::NoSuchCmd(id)) => {
```
rogercloud, line 244: Do you consider refactoring SyncedError just like ExeError in this PR? String is not a good practice for an error type after all.

iGxnon (dismissed):

curp-external-api/src/cmd.rs
```diff
@@ -11,6 +11,16 @@ use crate::LogIndex;
 pub trait Command:
     Sized + Sync + Send + DeserializeOwned + Serialize + std::fmt::Debug + Clone + ConflictCheck
 {
+    /// The error type that is returned in execution
```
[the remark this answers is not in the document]
Phoenix500526: I guess the main reason might be to obtain user-defined error types in some related methods of `ConnectApi` trait. It may look weird to introduce a dependency between `CommandExecutor` and `ConnectApi`. `Command` seems more appropriate. So, the Error type was moved from CE to Command. After all, there's no need to keep the error type in both Command and CE.

Phoenix500526:
I think this implementation is too invasive for the curp consensus. `ProposeError::ExecutionError` refers to errors gnenrated during the execution phase of CURP protocol. In fact, the client's command may not be turly executed during this phase. It will be executed during the after sync phase.  Therefore, I don't think it's a good practice to include `C::Error` in `ProposeError::ExecutionError`. Can we instead combine the specific command's execution error and the specific command's response together?  For example, distribute the `ExecuteError` to ER (CommandResponse) and ASR (SyncResponse)?

curp/src/client.rs
bsbds: The SyncedError comes from the `SyncError` in curp. So it's a curp's internal error and is like unrecoverable for the client, that a single error type is enough?

curp-external-api/src/cmd.rs
```diff
@@ -63,14 +73,24 @@ pub trait Command:
...
-    ) -> Result<Self::ASR, E::Error>
+    ) -> Result<Self::ASR, Self::Error>
     where
         E: CommandExecutor<Self> + Send + Sync,
     {
         <E as CommandExecutor<Self>>::after_sync(e, self, index, prepare_res).await
     }
 }
 
+/// Creates the error from a arbitrary string
```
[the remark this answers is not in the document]
bsbds: fixed

bsbds:
I have pushed another refactored version @Phoenix500526 @rogercloud @iGxnon

curp-external-api/src/cmd.rs
```diff
@@ -11,6 +11,15 @@ use crate::LogIndex;
 pub trait Command:
     Sized + Sync + Send + DeserializeOwned + Serialize + std::fmt::Debug + Clone + ConflictCheck
 {
+    /// Error type
+    type Error: std::fmt::Debug
```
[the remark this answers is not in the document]
bsbds: done

curp/src/client.rs
```diff
@@ -164,23 +164,31 @@ where
...
+                                    execute_result.is_none(),
+                                    "should not set exe result twice"
+                                );
+                                execute_result = Some(er);
+                            }
+                            Err(e) => {
+                                return Err(e);
+                            }
+                        }
                     }
```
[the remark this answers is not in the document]
bsbds: done

curp/src/error.rs
```diff
@@ -1,17 +1,10 @@
...
-/// Rpc Error
-#[allow(clippy::module_name_repetitions)] // this-error generate code false-positive
-#[non_exhaustive]
-#[derive(Error, Debug)]
-pub enum RpcError {
-    /// Met I/O error during rpc communication
-    #[error("meet io related error")]
-    IoError(#[from] io::Error),
-}
+use crate::rpc::SyncError;
```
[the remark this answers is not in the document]
bsbds: done

curp/src/error.rs
```diff
@@ -46,31 +39,31 @@ pub enum ProposeError {
...
     #[error("encode error: {0}")]
     EncodeError(String),
 }
 
-impl From<tonic::transport::Error> for ProposeError {
+/// The error met during propose phase
+#[derive(Error, Debug, Serialize, Deserialize)]
+#[allow(clippy::module_name_repetitions)] // this-error generate code false-positive
+#[error("rcp error {0}")]
+pub struct RpcError(String);
```
Phoenix500526, line 54: What's the difference between `RpcError` and `RpcServiceError`?

xline/src/storage/execute_error.rs
```diff
@@ -1,134 +1,124 @@
...
+    PermissionDenied,
+}
+
+impl From<ExecuteError> for tonic::Status {
+    #[inline]
+    fn from(err: ExecuteError) -> Self {
+        let code = match err {
+            ExecuteError::InvalidRequest(_)
+            | ExecuteError::KeyNotFound
```
Phoenix500526, line 96: The `tonic::Code::NotFound` indicates that some requested entity was not found. Therefore, the `KeyNotFound` should be converted to tonic::CodeNotFound. FYI: https://docs.rs/tonic/0.9.2/tonic/enum.Code.html

xline/src/storage/execute_error.rs
```diff
@@ -1,134 +1,124 @@
...
+            | ExecuteError::AuthFailed
+            | ExecuteError::PermissionNotGiven
+            | ExecuteError::InvalidAuthManagement
+            | ExecuteError::TokenNotProvided => tonic::Code::InvalidArgument,
+            ExecuteError::LeaseExpired(_)
```
Phoenix500526, line 101: `LeaseExpired` should be converted to `DeadlineExceeded `. And the `LeaseAlreadyExists` should be converted to `tonic::Code::AlreadyExists `

curp/src/error.rs
bsbds: `RpcServiceError` is not used by any code, I removed it from `ServerError`.

Phoenix500526 (dismissed):

Phoenix500526 (approved):

Issue #188, opened 2023-03-05
[Proposal] Compaction Feature Designed Doc


Phoenix500526:
# Motivation

Xline introduces the MVCC mechanism to keep multiple historical versions of each key value for some features, such as watch and transaction. As more and more modifications are made,  the space these historical versions consume, including memory and disk, will grow infinitely and finally threaten to cause OOM.

That's why we need compaction to remove some outdated historical versions. We need to support two different compacted strategies:
1. Manual Compaction： allow users to compact historical versions manually via a Compaction API, and it should be compatible with [the etcdctl COMPACTION API](https://github.com/etcd-io/etcd/tree/main/etcdctl#compaction-options-revision)
2. Automatic Compaction: provides two compaction configs,  --auto-compaction-mode and --auto-compaction-retention. There are two different compaction strategies：

  - Periodic Compaction Mode：--auto-compaction-mode=periodic --auto-compaction-retention=30m  automatically compact with 30-minute retention window
  - Revision Compaction: --auto-compaction-mode=revision --auto-compaction-retention=1000 automatically compact on "latest revision" -  1000 every 5-minute (when latest revision is 30000, compact on revision 29000).

# Architecture Overview
<img width="849" alt="image" src="https://user-images.githubusercontent.com/9337239/222961433-80f1cbf8-c3d7-4021-a132-e1cbb4e10ed9.png">

### Overview
```Rust
/// Compact Strategy interface
pub trait CompactStrategy {
    /// Run a 
    fn run();
    fn pause();
    fn resume();
    fn stop();
}

/// AutoCompactor definition
pub struct AutoCompactor<CS: CompactStrategy> {
    strategy: CS;
}
```

When XlineServer starts, a Compactor object is created and the internal Compact Strategy is set according to the configuration file. This is implemented by the Strategy Design Pattern. 

### Periodic Strategy
```Rust
// Periodic compacts the log by purging revisions older than
// the configured retention time.
struct Periodic {
    // indicates whether the current Compact Strategy is paused or not.
    paused: AtomicBool,
    // revision buffer
    revs: Vec<Option<Revision>>,
    // KvServer is the real one who does the Compact process.
    kv: Arc<KvServer>,
    // period
    period: Duration,
    ...
}

impl CompactStrategy for Periodic {
    ...
}
```

[...]

PR #401, opened 2023-07-27
Feature/auto compactor

Phoenix500526:
1. implement auto-compaction logic, read #188 for more details
2. break down the dependencies between CurpServer and CurpClient.


xline/src/storage/compact/periodic_compactor.rs
```diff
@@ -0,0 +1,305 @@
...
+        let shutdown_trigger = self.shutdown_trigger.listen();
+        let (sample_frequency, sample_total) = sample_config(self.period);
+        let mut ticker = tokio::time::interval(sample_frequency);
+        let mut revision_window = RevisionWindow::new(sample_total);
+        tokio::pin!(shutdown_trigger);
+        loop {
+            revision_window.sample(self.revision_getter.get());
+            tokio::select! {
+                _ = ticker.tick() => {
+                    if let Some(last_compacted_rev) = self.do_compact(last_revision, &revision_window).await {
```
bsbds, line 194: If a leader retires, a new leader is elected and do some compaction. After that the original leader is re-elected, will it try to compact the outdated revisions store in the buffer?

xline/src/storage/compact/periodic_compactor.rs
```diff
@@ -0,0 +1,305 @@
...
+        Self {
+            ring_buf: Vec::with_capacity(retention),
+            cursor: retention.overflow_sub(1),
+            retention,
+        }
+    }
+
+    /// Store the revision into the inner ring buffer
+    #[allow(clippy::integer_arithmetic)]
+    fn sample(&mut self, revision: i64) {
```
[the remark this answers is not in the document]
Phoenix500526: Done

xline/src/storage/compact/periodic_compactor.rs
Phoenix500526: No, the outdated revision will be overridden since the revision window keeps sampling every tick.

xline/src/storage/compact/revision_compactor.rs
```diff
@@ -0,0 +1,152 @@
...
+            shutdown_trigger,
+            retention,
+        })
+    }
+
+    /// perform auto compaction logic
+    async fn do_compact(&self, last_revision: Option<i64>) -> Option<i64> {
+        if !self.is_leader.load(Relaxed) {
+            return None;
+        }
```
themanforfree, line 56: I recommend move this to the run loop, and use a event to notioy the loop when current node is not leader

xline/src/storage/compact/mod.rs
```diff
@@ -1,19 +1,100 @@
...
+        AutoCompactConfig::Periodic(period) => {
+            PeriodicCompactor::new_arc(is_leader, client, revision_getter, shutdown_trigger, period)
+        }
+        AutoCompactConfig::Revision(retention) => RevisionCompactor::new_arc(
+            is_leader,
+            client,
+            revision_getter,
+            shutdown_trigger,
+            retention,
+        ),
```
themanforfree, line 84: Can we get these two modes to work together

xline/src/server/watch_server.rs
```diff
@@ -301,31 +289,46 @@ where
...
+            WatchResponse {
+                header: Some(ResponseHeader {
+                    revision: watch_event.revision(),
+                    ..ResponseHeader::default()
+                }),
+                watch_id,
+                events,
+                ..WatchResponse::default()
+            }
         };
```
themanforfree, line 330: Please build `WatchResponse` before `if else` block

xline/src/storage/compact/mod.rs
Phoenix500526: I don't get the point of the "work together" here. Do you mean to enable both of them at the same time?

xline/src/storage/compact/revision_compactor.rs
Phoenix500526: We don't need another event to do the same thing in that the `State` will invoke the `on_election_win` method to notify the auto compactor. If we move this to the run loop, we cannot cover this branch in our test.

xline/src/server/watch_server.rs
Phoenix500526: How come?  The `if else` block is responsible for returning different `WatchResponse` based on whether the given `watch_event` is compacted or not.

xline/src/server/watch_server.rs
```diff
@@ -198,17 +198,6 @@ where
...
-        if req.start_revision < compacted_revision {
-            let result = Err(tonic::Status::invalid_argument(format!(
-                "required revision {} has been compacted, compacted revision is {}",
-                req.start_revision, compacted_revision
-            )));
-            if self.response_tx.send(result).await.is_err() {
-                self.stop_notify.notify(1);
-            }
-            return;
-        }
```
iGxnon, line 211: Why do we need to remove these?

xline/src/storage/compact/periodic_compactor.rs
```diff
@@ -0,0 +1,282 @@
...
+                now.elapsed().as_secs()
+            );
+            target_revision
+        }
+    }
+}
+
+/// Calculate the sample frequency and the total amount of samples.
+fn sample_config(period: Duration) -> (Duration, usize) {
+    let one_hour = Duration::from_secs(60.overflow_mul(60));
```
iGxnon, line 130: Making a constant here might be better

xline/src/storage/compact/periodic_compactor.rs
iGxnon: What is the consequence of compacting the outdated revisions? The servers may have different durations for ticks.

xline/src/storage/compact/periodic_compactor.rs
Phoenix500526: Done

xline/src/server/watch_server.rs
Phoenix500526: For compatibility with etcd. When using etcdctl to watch a key that has already been compacted, the creation of the watcher can still succeed. However, this watcher will fail during the synchronization process and return the corresponding WatchResponse.

xline/src/storage/compact/mod.rs
themanforfree: Yes, I mean to enable both of them at the same time

xline/src/server/watch_server.rs
themanforfree: we can build the response first, and modify it in different branches to reduce code duplication

xline/src/storage/compact/mod.rs
Phoenix500526: No, they are different strategies. You can only enable either periodic mode or revision mode at the same time.

xline/src/server/watch_server.rs
Phoenix500526: Done

iGxnon (approved):

xline/src/storage/compact/revision_compactor.rs
```diff
@@ -0,0 +1,162 @@
...
+        if target_revision <= 0 || Some(target_revision) <= last_revision {
+            return None;
+        }
+
+        let now = Instant::now();
+        info!(
+            "starting auto revision compaction, revision = {}, retention = {}",
+            target_revision, self.retention
+        );
+        // TODO: add more error processing logic
```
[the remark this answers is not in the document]
Phoenix500526: Done

bsbds (approved):

iGxnon (approved):

PR #405, opened 2023-07-28
docs: xlinectl document

bsbds:

    add document to xlinectl.

### DIFF
--- a/xlinectl/README.md
+++ b/xlinectl/README.md
@@ -1,3 +1,138 @@
-# Xlinectl
+# xlinectl
 
 This crate provides a command line client for Xline.
+
+## Global Options
+- endpoints <SERVER_NAME ADDR>... -- Set Xline endpoints, which are separated by ','
+    ```bash
+    # connect to servers with specific addresses
+    ./xlinectl --endpoints "server0 10.0.0.1:2379, server1 10.0.0.2:2379, server2 10.0.0.3:2379"
+    ```
+- user <USERNAME[:PASSWD]> -- The name of the user, this provide a shorthand to set password
+    ```bash
+    # connect to servers using user `foo` with password `bar`
+    ./xlinectl --user foo:bar
+    ```
+- password <PASSWD> -- The password of the user, should exist if password not set in `--user`
+    ```bash
+    # connect to servers using user `foo` with password `bar`
+    ./xlinectl --user foo --password bar
+    ```
+- wait_synced_timeout <TIMEOUT> -- The timeout for Curp client waiting synced(in secs) [default: 2]
+- propose_timeout <TIMEOUT> -- The timeout for Curp client proposing request(in secs) [default: 1]
+- retry_timeout <TIMEOUT> -- The timeout for Curp client retry interval(in millis) [default: 50]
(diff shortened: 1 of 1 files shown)


xlinectl/README.md
```diff
@@ -1,3 +1,138 @@
...
+    ```
+- wait_synced_timeout <TIMEOUT> -- The timeout for Curp client waiting synced(in secs) [default: 2]
+- propose_timeout <TIMEOUT> -- The timeout for Curp client proposing request(in secs) [default: 1]
+- retry_timeout <TIMEOUT> -- The timeout for Curp client retry interval(in millis) [default: 50]
+- printer_type <TYPE> -- The format of the result that will be printed [default: SIMPLE] [possible values: SIMPLE, FIELD]
+
+## Output Format
+
+All command output will first print the response header, and then print the response fields
+
```
iGxnon, line 29: I think we can provide an output format example.

xlinectl/README.md
```diff
@@ -1,3 +1,138 @@
...
+    ```
+- password <PASSWD> -- The password of the user, should exist if password not set in `--user`
+    ```bash
+    # connect to servers using user `foo` with password `bar`
+    ./xlinectl --user foo --password bar
+    ```
+- wait_synced_timeout <TIMEOUT> -- The timeout for Curp client waiting synced(in secs) [default: 2]
+- propose_timeout <TIMEOUT> -- The timeout for Curp client proposing request(in secs) [default: 1]
+- retry_timeout <TIMEOUT> -- The timeout for Curp client retry interval(in millis) [default: 50]
+- printer_type <TYPE> -- The format of the result that will be printed [default: SIMPLE] [possible values: SIMPLE, FIELD]
```
Phoenix500526, line 24: We should add some output format examples here. Does the FIELD type look like JSON format?

xlinectl/README.md
```diff
@@ -1,3 +1,138 @@
...
+    ```bash
+    # connect to servers using user `foo` with password `bar`
+    ./xlinectl --user foo:bar
+    ```
+- password <PASSWD> -- The password of the user, should exist if password not set in `--user`
+    ```bash
+    # connect to servers using user `foo` with password `bar`
+    ./xlinectl --user foo --password bar
+    ```
+- wait_synced_timeout <TIMEOUT> -- The timeout for Curp client waiting synced(in secs) [default: 2]
```
Phoenix500526, line 21: Does TIMEOUT need to specify the unit?  `--wait_synced_timeout 2s` and `--wait_synced_timeout 2`, which one is correct?

xlinectl/README.md
bsbds: added

xlinectl/README.md
bsbds: I added the SIMPLE format document. The FIELD type is more like a human readable format, and too verbose to put in document.

xlinectl/README.md
bsbds: no, it should be consistent with the default type.

iGxnon (approved):

Phoenix500526 (approved):

PR #407, opened 2023-08-01
fix/refactor: fix propose doesn't handle SyncedError and reduce code duplication

themanforfree:

fix propose doesn't handle `SyncedError`
return an `Error` instead of `unreachable`, and do some refactor to  reduce code duplication
no


xline/src/server/command.rs
```diff
@@ -598,6 +600,75 @@ impl CurpCommand for Command {
...
+                .into_iter()
+                .map(|k| KeyRange::new(k, ""))
+                .collect()
+        }
+        _ => vec![],
+    };
+    Command::new(keys, wrapper, propose_id)
+}
+/// Propose request and get result with fast/slow path
+pub(super) async fn propose(
```
Phoenix500526, line 641: Can we move this method to the Client?

themanforfree: Fixed

xline/src/server/command.rs
```diff
@@ -598,6 +599,60 @@ impl CurpCommand for Command {
...
+            vec![KeyRange::new(req.key.as_slice(), req.range_end.as_slice())]
+        }
+        RequestWrapper::TxnRequest(ref req) => req
+            .compare
+            .iter()
+            .map(|cmp| KeyRange::new(cmp.key.as_slice(), cmp.range_end.as_slice()))
+            .collect(),
+        RequestWrapper::LeaseRevokeRequest(ref req) => {
+            let Some(lease_storage) = lease_storage else {
+                panic!("lease_storage is None when creating command of LeaseRevokeRequest")
```
Phoenix500526, line 627: The panic message should focus more on the expectation rather than the fact. "is None" is a fact while "should be Some(_)" is this method's expectation. I prefer using expectation in the panic message.

themanforfree: Fixed

Phoenix500526 (approved):

xline/src/server/lock_server.rs
```diff
@@ -1,34 +1,38 @@
...
+    storage::storage_api::StorageApi,
 };
 
 /// Default session ttl
 const DEFAULT_SESSION_TTL: i64 = 60;
 
 /// Lock Server
 #[derive(Debug)]
-pub(super) struct LockServer {
+pub(super) struct LockServer<S> {
```
bsbds, line 35: could we just move the generic `S` to the `propose` method, it seems only this method need this generic.

themanforfree: The main reason for adding `S` here is `command_from_request_wrapper`

bsbds: sorry I forget it's an internal method

bsbds (approved):

iGxnon (approved):

PR #391, opened 2023-07-20
refactor: return revision in execution result

bsbds:
Described in #390 

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

    This PR modified `header_gen`, the header generator can always return the latest revision.

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #408, opened 2023-08-01
refactor: read only txn request will use read state mechanism

themanforfree:
Reopen #367 

(Accidentally closed)

### DIFF
--- a/xline/src/server/kv_server.rs
+++ b/xline/src/server/kv_server.rs
@@ -79,15 +79,12 @@ where
         if let Some(response) = response_op.response {
             response
         } else {
-            panic!("Receive empty ResponseOp");
+            unreachable!("Receive empty ResponseOp");
         }
     }
 
-    /// Execute `RangeRequest` in current node
-    fn serializable_range(
-        &self,
-        wrapper: &RequestWithToken,
-    ) -> Result<tonic::Response<RangeResponse>, tonic::Status> {
+    /// serializable execute request in current node
+    fn do_serializable(&self, wrapper: &RequestWithToken) -> Result<Response, tonic::Status> {
         self.auth_storage
             .check_permission(wrapper)
             .map_err(|err| tonic::Status::invalid_argument(err.to_string()))?;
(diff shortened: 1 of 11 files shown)


xlineapi/src/lib.rs
```diff
@@ -671,6 +671,23 @@ impl TxnRequest {
         self.success.iter().all(read_only_checker) && self.failure.iter().all(read_only_checker)
     }
 
+    /// Checks whether a given `TxnRequest` is serializable or not.
+    pub fn is_serializable(&self) -> bool {
```
Phoenix500526, line 675: Please add a test to cover this method.

themanforfree: Fixed

xlineapi/src/lib.rs
```diff
@@ -781,4 +781,82 @@ mod test {
...
+            compare: vec![],
+            success: vec![RequestOp {
+                request: serializable_req.clone(),
+            }],
+            failure: vec![RequestOp {
+                request: serializable_req.clone(),
+            }],
+        };
+
+        assert!(serializable_txn_req.is_serializable());
```
Phoenix500526, line 802: We should cover the happy path and the sad path here. Given that `is_serializable` is a recursive implementation, we'd better create a dedicated unit test in order to cover all possible branches.

Phoenix500526 (dismissed):

xline/src/server/kv_server.rs
```diff
@@ -310,19 +309,34 @@ where
...
+            let token = get_token(request.metadata());
+            let wrapper = RequestWithToken::new_with_token(request.into_inner().into(), token);
+            let propose_id = self.generate_propose_id();
+            let cmd = command_from_request_wrapper::<S>(propose_id, wrapper, None);
+            if !is_serializable {
+                self.wait_read_state(&cmd).await?;
+            }
+            self.do_serializable(cmd.request())?
+        } else {
+            let is_fast_path = false; // lock need revision of txn
```
bsbds, line 325: after #391 the revision will be returned in execution result, so we can use fast path here for better performance

themanforfree: Fixed

Phoenix500526 (approved):
LGTM

bsbds (approved):

Issue #306, opened 2023-06-05
[Feature]: Membership change


themanforfree:
### Description about the feature

**Membership change has two different situations**
1. Change only one node at a time, in this case there must be duplicate nodes in majority of the old and new configurations, it is impossible to elect two leaders at the same time, we can handle this situation simply, after the new configuration is committed, the new configuration can be used to make decisions.
![image](https://github.com/datenlord/Xline/assets/56149350/3ca61b1e-3634-4e1f-aa7c-fbeb19607912)
2. Change multiple nodes at a time,  the following situation may occur. nodes using different configurations may have two disjoint majorities, which means that two leaders may be elected, the solution is introduce an intermediate state, after the leader receives the configuration change request, the new configuration node set is `C_new`, and the old configuration node set is `C_old`. After `C_new` is committed, it enters the joint consensus state. When an election occurs, `C_old` and `C_new` jointly decide, and then the leader creates an empty Configuration change log, after this log committed, cluster can use `C_new` alone to make decisions, and the cluster completes this configuration change node, and this configuration change is complete.
![image](https://github.com/datenlord/Xline/assets/56149350/a3f09a8b-af9f-4bd9-8d63-4b31b2a1411e)

**Learner**
When adding a new node to the cluster, the new node does not have the current data of the cluster. At this time, the leader needs to sync the data to the new node. This process may take a while, which will increase the risk of cluster unavailability. Therefore, the Learner node is introduced. Nodes only sync logs, do not participate in voting, wait until the Leaner node catches up with the progress of the leader, and then promote it to a normal node.

**I plan to implement the membership change in three stages**
1.  implement single node change, and prepare some struct for multiple nodes change.
2. implement Learner node.
3. implement multiple nodes change.

### Tests

- [x] The member id and cluster id calculated by all nodes should be consistent when Initialize the cluster.
- [x] The new node should pull member IDs directly from the existing cluster and use them, and should not calculate IDs by itself.
- [x] Client should use new configuration  after  cluster member change

### Related papers
1. https://raft.github.io/raft.pdf
2. https://web.stanford.edu/~ouster/cgi-bin/papers/OngaroPhD.pdf

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

themanforfree:
More details of stage 1

1. use `Arc` to share `LogEntry`, then we can make `LogEntry` instead of `Command` transmitted in conflict checked mpmc channel, this can make curp manage content other than Command.
[...]

PR #412, opened 2023-08-07
refactor: build client without `ServerId`

themanforfree:
#306 

build client without `ServerId`
add client builder
add `fetch_cluster` rpc
allows the client to build without `ServerId`, which automatically fetch the cluster information from the server
yes, `Client::new` is removed, use client builder instead

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -24,10 +24,6 @@ use crate::{
 /// Client builder
 #[derive(Debug)]
 pub struct Builder<C> {
-    /// Server addresses
-    addrs: Vec<String>,
-    /// Server addresses and ids (used in an inner client)
-    all_members: HashMap<ServerId, String>,
     /// Local server id (used in an inner client)
     local_server_id: Option<ServerId>,
     /// Client timeout
@@ -41,8 +37,6 @@ impl<C> Default for Builder<C> {
     #[must_use]
     fn default() -> Self {
         Self {
-            addrs: Vec::new(),
-            all_members: HashMap::new(),
             local_server_id: None,
             timeout: None,
             phantom: PhantomData,
@@ -59,27 +53,6 @@ where
     #[must_use]
     pub fn new() -> Self {
         Self::default()
-    }
-
-    /// Set addresses of servers
-    /// # Panics
-    /// Panics when set `all_members` before
-    #[inline]
-    #[must_use]
-    pub fn addrs(self, addrs: Vec<String>) -> Self {
-        Self { addrs, ..self }
-    }
-
-    /// Set server id and all addresses of servers
-    /// # Panics
-    /// Panics when set `addrs` before
-    #[inline]
-    #[must_use]
-    pub fn all_members(self, all_members: HashMap<ServerId, String>) -> Self {
-        Self {
-            all_members,
-            ..self
-        }
     }
 
     /// Set client timeout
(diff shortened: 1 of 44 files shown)


curp/src/client.rs
```diff
@@ -1,30 +1,198 @@
...
+            local_server_id: Some(local_server_id),
+            ..self
+        }
+    }
+
+    /// Build client
+    /// # Errors
+    /// Return error when meet rpc error or missing some arguments
+    #[inline]
+    pub async fn build(self) -> Result<Client<C>, ClientBuildError> {
```
Phoenix500526, line 109: We can use `derive_builder` crate to reduce some boilerplate code.

curp/src/client.rs
```diff
@@ -1,30 +1,198 @@
...
+                Self::build_from_all_members(
+                    timeout,
+                    self.local_server_id,
+                    self.all_members,
+                    None,
+                    0,
+                )
+                .await
+            }
+            (true, true) => Err(ClientBuildError::invalid_aurguments(
```
Phoenix500526, line 127: Why should we return Error when the user provides the `addr` and the `all_members`? It's okay to ignore one of them and build Client with the other.

curp/src/client.rs
```diff
@@ -116,9 +282,15 @@ where
     ) -> Result<(Option<<C as Command>::ER>, bool), CommandProposeError<C>> {
         debug!("fast round for cmd({}) started", cmd_arc.id());
         let req = ProposeRequest::new(cmd_arc.as_ref()).map_err(Into::<ProposeError>::into)?;
-        let mut rpcs: FuturesUnordered<_> = self
-            .connects
-            .values()
+
+        let (connects, superquorum) = self.state.map_read(|state_r| {
```
Phoenix500526, line 286: Is this a good practice here? I'm afraid that doing so may be a disservice to the performance in that it might exacerbate our lock contention that are already severe.

curp/src/client.rs
themanforfree: `derive_builder` does not support custom build method

curp/src/client.rs
themanforfree: Fixed with other implementation

curp/src/client.rs
```diff
@@ -1,30 +1,198 @@
...
+
+    /// Build client
+    /// # Errors
+    /// Return error when meet rpc error or missing some arguments
+    #[inline]
+    pub async fn build(self) -> Result<Client<C>, ClientBuildError> {
+        let Some(timeout) = self.timeout else {
+            return Err(ClientBuildError::invalid_aurguments("timeout is required"));
+        };
+        match (self.addrs.is_empty(), self.all_members.is_empty()) {
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/client.rs
Phoenix500526: FYI: https://docs.rs/derive_builder/0.12.0/derive_builder/#build-method-customization

Phoenix500526: FYI: https://docs.rs/derive_builder/0.12.0/derive_builder/#build-method-customization

curp/src/client.rs
themanforfree: This does disservice some performance, but the old implementation can’t meet the needs of membership changes, do you have a better suggestion?

curp/src/client.rs
themanforfree: Fixed

curp/src/client.rs
Phoenix500526: We can move the `connects` back to the client, but implement it with some lock-free hashmap, like dashmap.

curp/src/client.rs
```diff
@@ -1,36 +1,111 @@
...
+                let mut protocol_client = ProtocolClient::connect(addr).await?;
+                let mut req = tonic::Request::new(FetchClusterRequest::new());
+                req.set_timeout(timeout);
+                let fetch_cluster_res = protocol_client.fetch_cluster(req).await?.into_inner();
+                Ok::<FetchClusterResponse, ClientBuildError>(fetch_cluster_res)
+            };
+            Box::pin(fut)
+        });
+
+        let res = select_all(futs).await.0?;
```
bsbds, line 96: As the operation in `fut` may fail, we should consider select the first successful future that returned. If none of them success, return an error instead.

curp/src/client.rs
themanforfree: Fixed

curp/src/client.rs
themanforfree: Fixed

Phoenix500526 (approved):

bsbds (approved):

PR #413, opened 2023-08-08
refactor: use u64 as ServerId

themanforfree:
#306 
Based on #412 
If we use `String` as the server ID, it is not compatible with the `etcd`'s membership change interface, and using String as the server ID will require clone in many cases, So we need to turn server ID into u64.
use u64 as ServerId instead of String, and  refactor check vote logic
yes, All server ID related parts need to be replaced with `u64`

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -67,7 +67,7 @@ impl<C: Command> Builder<C> {
         };
         let connects = rpc::connect(all_members).await.collect();
         let client = Client::<C> {
-            local_server_id: self.local_server_id.clone(),
+            local_server_id: self.local_server_id,
             state: RwLock::new(State::new(None, 0)),
             timeout,
             connects,
@@ -125,7 +125,7 @@ impl<C: Command> Builder<C> {
             .await?;
         let connects = rpc::connect(res.all_members).await.collect();
         let client = Client::<C> {
-            local_server_id: self.local_server_id.clone(),
+            local_server_id: self.local_server_id,
             state: RwLock::new(State::new(res.leader_id, res.term)),
             timeout,
             connects,
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -17,12 +17,13 @@ use crate::{
     error::{
         ClientBuildError, CommandProposeError, CommandSyncError, ProposeError, RpcError, SyncError,
     },
+    members::ServerId,
     rpc::{
         self, connect::ConnectApi, protocol_client::ProtocolClient, FetchClusterRequest,
         FetchClusterResponse, FetchLeaderRequest, FetchReadStateRequest, ProposeRequest,
         ReadState as PbReadState, SyncResult, WaitSyncedRequest,
     },
-    LogIndex, ServerId,
+    LogIndex,
 };
 
 /// Protocol client
@@ -241,19 +242,19 @@ where
                         // reset term only when the resp has leader id to prevent:
                         // If a server loses contact with its leader, it will update its term for election. Since other servers are all right, the election will not succeed.
                         // But if the client learns about the new term and updates its term to it, it will never get the true leader.
-                        if let Some(ref leader_id) = resp.leader_id {
+                        if let Some(leader_id) = resp.leader_id {
                             state.update_to_term(resp.term());
-                            state.set_leader(leader_id.clone());
+                            state.set_leader(leader_id);
                             execute_result = None;
                         }
                     }
                     Ordering::Equal => {
-                        if let Some(ref leader_id) = resp.leader_id {
+                        if let Some(leader_id) = resp.leader_id {
                             if state.leader.is_none() {
-                                state.set_leader(leader_id.clone());
+                                state.set_leader(leader_id);
                             }
                             assert_eq!(
-                                state.leader.as_ref(),
+                                state.leader,
                                 Some(leader_id),
                                 "there should never be two leader in one term"
                             );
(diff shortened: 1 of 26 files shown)


xlinectl/src/main.rs
```diff
@@ -148,6 +148,7 @@
         clippy::integer_arithmetic
     )
 )]
+#![allow(unused)]
```
iGxnon, line 151: What's the reason?

curp/src/server/raw_curp/mod.rs
```diff
@@ -478,14 +478,13 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         let mut cst_w = self.cst.lock();
         debug!("{}'s vote is granted by server {}", self.id(), id);
 
-        cst_w.votes_received += 1;
+        let _ig = cst_w.votes_received.insert(id, vote_granted);
```
bsbds, line 481: The `vote_granted` is always true here, as it has been checked in L474. Please explain the usecase of the `votes_received` hasmap.

curp/src/members.rs
```diff
@@ -1,134 +1,172 @@
...
     #[inline]
     pub fn peers_len(&self) -> usize {
-        self.peers.len()
+        self.members.len().overflow_sub(1)
+    }
+
+    /// Get id by name
+    #[must_use]
+    #[inline]
+    pub fn get_id_by_name(&self, name: &str) -> Option<ServerId> {
```
bsbds, line 166: We should be cautious about this method, as there's may multiple members who have the same name. Maybe limit this to test use only.

xlinectl/src/main.rs
themanforfree: Removed in #412

curp/src/client.rs
```diff
@@ -42,23 +118,30 @@ impl<C: Command> Debug for Client<C> {
...
+#[derive(Debug, Default)]
 struct State {
     /// Current leader
     leader: Option<ServerId>,
     /// Current term
     term: u64,
     /// When a new leader is set, notify
     leader_notify: Arc<Event>,
+    /// All servers's `Connect`
+    connects: HashMap<ServerId, Arc<dyn ConnectApi>>,
```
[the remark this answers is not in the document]
themanforfree: Fixed in #412

curp/src/rpc/connect.rs
```diff
@@ -59,9 +59,9 @@ pub(crate) async fn connect(
 /// Connect interface
 #[cfg_attr(test, automock)]
 #[async_trait]
-pub(crate) trait ConnectApi: Send + Sync + 'static {
+pub(crate) trait ConnectApi: Send + Sync + 'static + Debug {
```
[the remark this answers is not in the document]
themanforfree: Removed in #412

curp/src/members.rs
themanforfree: Added `#[cfg(test)]`

curp/src/server/raw_curp/mod.rs
themanforfree: Fixed, This is to prepare for the introduction of joint config in the future, there may be two configurations that use the same `votes_received` to check the vote result

curp/src/server/raw_curp/mod.rs
```diff
@@ -991,7 +1001,7 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
 
     /// Get quorum: the smallest number of servers who must be online for the cluster to work
     fn quorum(&self) -> u64 {
-        (self.ctx.cluster_info.peers_len() / 2 + 1).numeric_cast()
+        ((self.ctx.cluster_info.peers_len() + 1) / 2 + 1).numeric_cast()
```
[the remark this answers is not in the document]
themanforfree: Added

curp/tests/server.rs
```diff
@@ -41,6 +43,22 @@ async fn basic_propose() {
...
+async fn fetch_cluster() {
+    init_logger();
+    let group = CurpGroup::new(3).await;
+
+    let all_addrs = group.all.values().cloned().collect::<Vec<_>>();
+    let _client = Builder::<TestCommand>::default()
+        .timeout(ClientTimeout::default())
+        .build_from_addrs(all_addrs)
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
themanforfree: Renamed

curp/src/members.rs
```diff
@@ -1,134 +1,172 @@
...
-    /// calculate the member id
-    #[must_use]
-    #[inline]
-    pub fn gen_member_id(&self, cluster_name: &str) -> u64 {
-        let ts = SystemTime::now()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .unwrap_or_else(|e| unreachable!("SystemTime before UNIX EPOCH! {e}"))
-            .as_secs();
+    /// Calculate the member id
+    fn calculate_member_id(address: &str, cluster_name: &str, timestamp: Option<u64>) -> ServerId {
```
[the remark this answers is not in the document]
themanforfree: yes, A timestamp is not required fo computing the member ID when initializing the cluster, because to ensure that all nodes calculate the same ID, New nodes added need to be calculated based on the timestamp

Phoenix500526 (approved):

iGxnon (approved):

PR #414, opened 2023-08-11
CI/Replace Dockerhub to GitHub Container Registry

iGxnon:

Test workflow: https://github.com/iGxnon/Xline/actions/runs/5830042498

    Replace dockerhub to GitHub container registry  

    See above

    No

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,7 @@ on:
 
 env:
   CI_RUST_TOOLCHAIN: 1.70.0
-  REGISTRY_IMAGE: datenlord/xline
+  IMAGE_ID: ghcr.io/xline-kv/xline
 
 jobs:
   push_image:
@@ -76,11 +76,12 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
 
-      - name: Login to DockerHub
+      - name: Login to GHCR
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          registry: ghcr.io
+          username: xline-kv
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build Docker Image
         id: build_image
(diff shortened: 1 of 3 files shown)


Phoenix500526 (dismissed):

.github/workflows/docker.yml
```diff
@@ -76,19 +76,25 @@ jobs:
...
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          registry: ghcr.io
+          username: ${{ env.REPOSITORY }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Image ID
+        run: |
+          IMAGE_ID=ghcr.io/${{ env.REPOSITORY }}/xline
+          echo IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]') >> $GITHUB_ENV
```
themanforfree, line 89: Why set up image ID at here, Could we set up it directly in `env`

.github/workflows/docker.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 env:
   CI_RUST_TOOLCHAIN: 1.70.0
-  REGISTRY_IMAGE: datenlord/xline
+  REPOSITORY: ${{ github.actor }}
```
themanforfree, line 8: The explanation of `github.actor` in GitHub Docs is as follows.
> The username of the user that triggered the initial workflow run.  

Should we hardcode it to `datenlord`?

iGxnon: In order to test in my forked repository, I haven't hard-coded the `REPOSITORY`. None of these CIs will trigger in the PRs, so should the `github.actor` here always be `xline-kv`?

.github/workflows/docker.yml
iGxnon: If we avoid hard-coding the `REPOSITORY`, we must convert the image ID to lowercase before pushing it to GHCR.

.github/workflows/docker.yml
themanforfree: This CI may be triggered manually, So the `github.actor` should be the username of the trigger.

iGxnon: Fixed

themanforfree (approved):

Phoenix500526 (approved):

PR #418, opened 2023-08-19
fix: CURP TLA+ quorum size calculation & property check

EAimTY:
The previous TLA+ specification of CURP had several issues:

1. The method for calculating the quorum size was incorrect
2. In the stability property check, the preservation of order among incommutable commands was not verified

The first problem is fixed by calculating quorum size using the maximum number of faulty participants `f`.

To solve the second problem, the detailed messaging abstraction between server and client is removed in favor of a much simpler system-wide consensus decision abstraction. The client-side details are omitted since they are not the main focus of this model.


rogercloud (approved):

Phoenix500526 (dismissed):

Phoenix500526 (commented):
Please add signed-off information in your commit

Phoenix500526 (approved):

Issue #330, opened 2023-06-14
Improve the readability of boot up errors in Xline


Phoenix500526:
At present, there are two ways to boot up an Xline instance: using command-line arguments or a configuration file. Typically, when no cmd line arguments are provided, the configuration information is read from `/etc/xline_server.conf`. However, if this file doesn't exist, it will throw out an error message, "Error: No such file or directory (os error 2)", which lacks readability. We can improve its readability by adding the configuration file name in this error message.

MarkGaox:
Hi, can you assign this to me?

Phoenix500526:
@MarkGaox Sure, thank you for your contribution.

PR #432, opened 2023-08-27
fix: improve readability of bootstrap errors

MarkGaox:

#330 
When the configuration file doesn't exist, Xline will throw out an error message, "Error: No such file or directory (os error 2)", which lacks readability. We need to improve its readability by adding the configuration file name in this error message.

Add error messages when reading of the config file fails.

### DIFF
--- a/xline/src/main.rs
+++ b/xline/src/main.rs
@@ -162,6 +162,7 @@ use utils::{
         TraceConfig, XlineServerConfig,
     },
     parse_batch_bytes, parse_duration, parse_log_level, parse_members, parse_rotation,
+    ConfigParseError,
 };
 use xline::{server::XlineServer, storage::db::DB};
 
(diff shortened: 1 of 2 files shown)


xline/src/main.rs
```diff
@@ -472,7 +472,12 @@ async fn main() -> Result<()> {
     let config: XlineServerConfig = if env::args_os().len() == 1 {
         let path =
             env::var("XLINE_SERVER_CONFIG").unwrap_or_else(|_| "/etc/xline_server.conf".to_owned());
-        let config_file = fs::read_to_string(&path).await?;
+        let config_file = match fs::read_to_string(&path).await {
```
Phoenix500526, line 475: I recommend using `map_err` to convert the error message rather than panic directly.

Phoenix500526 (commented):
Please add singed-off information in the commit message.

iGxnon (approved):
LGTM

Phoenix500526 (approved):
LGTM

PR #435, opened 2023-08-29
fix: check the password on leader

iGxnon:

    `AuthServer::authenticate` should not check password locally.
   
    Move the `check_password` to the leader.

    No

### DIFF
--- a/xline/src/storage/auth_store/store.rs
+++ b/xline/src/storage/auth_store/store.rs
@@ -283,6 +283,7 @@ where
         if !self.is_enabled() {
             return Err(ExecuteError::AuthNotEnabled);
         }
+        self.check_password(&req.name, &req.password)?;
         let token = self.assign(&req.name)?;
         Ok(AuthenticateResponse {
             header: Some(self.header_gen.gen_auth_header()),
@@ -868,7 +869,7 @@ where
         &self,
         username: &str,
         password: &str,
-    ) -> Result<i64, ExecuteError> {
+    ) -> Result<(), ExecuteError> {
         if !self.is_enabled() {
             return Err(ExecuteError::AuthNotEnabled);
         }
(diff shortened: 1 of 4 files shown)


xline/src/storage/auth_store/store.rs
```diff
@@ -868,7 +869,7 @@ where
         &self,
         username: &str,
         password: &str,
-    ) -> Result<i64, ExecuteError> {
```
Phoenix500526, line 871: Why should we remove the revision from the returned value?

iGxnon: Before, we kept the revision because locally checking the password lacked timeliness. We need to check the revision after log synced and then return the auth result. Now, checking password is performed on the leader, and it's always up-to-date, so there's no need to return the revision anymore.

Phoenix500526: In etcd, if a user modifies the password, it will make the auth_rev change. This can cause the previously assigned token to become invalid. In the previous implementation of the authenticate method, a comparison is made between this revision and the auth_rev in `auth_store`. However, in the current implementation, it seems this comparison might not exist. Could this potentially lead to some issues?

iGxnon: Some operations that modify `auth_store` are proposed in **slow_path**, including `user_change_password` which invalidates tokens. This means that the modifications will be synchronized into the `auth_store` of the majority of nodes before response. Then, requests with this token will not receive the expected `PR` during the `prepare` phase. Furthermore, we won't lose the latest revision even when a leader crash.

Phoenix500526 (approved):

Phoenix500526 (dismissed):
LGTM

xline/src/storage/auth_store/store.rs
bsbds: It seems the modifications is persistent asynchronously to the auth store after the follower response to the ae request. Will this cause problem when the leader crashes?

themanforfree (approved):
LGTM

Phoenix500526 (approved):
LGTM

themanforfree (approved):

PR #433, opened 2023-08-28
feature: add `apply_conf_change` method for `RawCurp`

themanforfree:
Please briefly answer these questions:
#306 
* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
apply config change to curp
* what changes does this pull request make?
refactor data structures via `DashMap`, and modify them when apply_conf_change
* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)
no

PR #437, opened 2023-08-30
Propose conf change

themanforfree:
described in  #306 
based on #433 ,https://github.com/xline-kv/curp-proto/pull/10

propose config change
implement `propose_conf_change` rpc
no

### DIFF
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -127,7 +127,6 @@ impl ClusterInfo {
             .addrs = addrs.into();
     }
 
-    // TODO
     /// Get server addresses via server id
     #[must_use]
     #[inline]
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -6,35 +6,10 @@ use std::{
 use dashmap::{mapref::one::Ref, DashMap};
 use itertools::Itertools;
 
-use crate::rpc::PbMember;
+use crate::Member;
 
 /// Server Id
 pub type ServerId = u64;
-
-/// Cluster member
-#[derive(Debug, Clone)]
-pub struct Member {
-    /// Server id of the member
-    id: ServerId,
-    /// Name of the member
-    name: String,
-    /// Addresses of the member
-    addrs: Vec<String>,
-    /// Learner
-    is_learner: bool,
-}
-
-impl From<Member> for PbMember {
-    #[inline]
-    fn from(member: Member) -> Self {
-        Self {
-            id: member.id,
-            name: member.name,
-            addrs: member.addrs,
-            is_learner: member.is_learner,
-        }
-    }
-}
 
 /// Cluster member
 impl Member {
@@ -84,7 +59,7 @@ impl Member {
 }
 
 /// cluster members information
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ClusterInfo {
     /// cluster id
     cluster_id: u64,
@@ -108,12 +83,7 @@ impl ClusterInfo {
             if name == self_name {
                 member_id = id;
             }
-            let member = Member {
-                id,
-                name,
-                addrs,
-                is_learner: false, // TODO restore learner here?
-            };
+            let member = Member::new(id, name, addrs, false);
             let _ig = members.insert(id, member);
         }
         debug_assert!(member_id != 0, "self_id should not be 0");
@@ -157,14 +127,12 @@ impl ClusterInfo {
             .addrs = addrs.into();
     }
 
+    // TODO
     /// Get server addresses via server id
     #[must_use]
     #[inline]
     pub fn addrs(&self, id: ServerId) -> Option<Vec<String>> {
-        self.members
-            .iter()
-            .find(|t| t.id == id)
-            .map(|t| t.addrs.clone())
+        self.members.get(&id).map(|t| t.addrs.clone())
     }
 
     /// Get the current member
@@ -206,8 +174,8 @@ impl ClusterInfo {
     }
 
     /// Calculate the member id
-    #[must_use]
     #[inline]
+    #[must_use]
     pub fn calculate_member_id(
         mut addrs: Vec<String>,
         cluster_name: &str,
@@ -268,6 +236,13 @@ impl ClusterInfo {
             .iter()
             .map(|t| (t.id, t.addrs.clone()))
             .collect()
+    }
+
+    /// Get all members
+    #[must_use]
+    #[inline]
+    pub fn members(&self) -> Vec<Member> {
+        self.members.iter().map(|t| t.value().clone()).collect()
     }
 
     /// Get length of peers
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -114,17 +114,21 @@ impl ClusterInfo {
 
     /// Remove a member
     #[inline]
-    pub fn remove(&self, id: &ServerId) {
-        _ = self.members.remove(id);
+    #[must_use]
+    pub fn remove(&self, id: &ServerId) -> Option<Member> {
+        self.members.remove(id).map(|(_id, m)| m)
     }
 
-    /// Update a member
+    /// Update a member and return old addrs
     #[inline]
-    pub fn update(&self, id: &ServerId, addrs: impl Into<Vec<String>>) {
-        self.members
+    pub fn update(&self, id: &ServerId, addrs: impl Into<Vec<String>>) -> Vec<String> {
+        let mut addrs = addrs.into();
+        let mut member = self
+            .members
             .get_mut(id)
-            .unwrap_or_else(|| unreachable!("member {} not found", id))
-            .addrs = addrs.into();
+            .unwrap_or_else(|| unreachable!("member {} not found", id));
+        std::mem::swap(&mut addrs, &mut member.addrs);
+        addrs
     }
 
     /// Get server addresses via server id
--- a/curp/src/rpc/connect.rs
+++ b/curp/src/rpc/connect.rs
@@ -86,9 +86,9 @@ pub(crate) async fn connect(
 /// Convert a vec of addr string to a vec of `InnerConnect`
 pub(crate) async fn inner_connect(
     members: HashMap<ServerId, Vec<String>>,
-) -> Result<impl Iterator<Item = (ServerId, ConnectApiWrapper)>, tonic::transport::Error> {
+) -> Result<impl Iterator<Item = (ServerId, InnerConnectApiWrapper)>, tonic::transport::Error> {
     connect_impl!(InnerProtocolClient<Channel>, InnerConnectApi, members)
-        .map(|iter| iter.map(|(id, connect)| (id, ConnectApiWrapper::new_from_arc(connect))))
+        .map(|iter| iter.map(|(id, connect)| (id, InnerConnectApiWrapper::new_from_arc(connect))))
 }
 
 /// Connect interface between server and clients
@@ -177,18 +177,18 @@ pub(crate) trait InnerConnectApi: Send + Sync + 'static {
     ) -> Result<tonic::Response<InstallSnapshotResponse>, RpcError>;
 }
 
-/// Connect Api Wrapper
+/// Inner Connect Api Wrapper
 /// The solution of [rustc bug](https://github.com/dtolnay/async-trait/issues/141#issuecomment-767978616)
 #[derive(Clone)]
-pub(crate) struct ConnectApiWrapper(Arc<dyn InnerConnectApi>);
+pub(crate) struct InnerConnectApiWrapper(Arc<dyn InnerConnectApi>);
 
-impl ConnectApiWrapper {
-    /// Create a new `ConnectApiWrapper` from `Arc<dyn ConnectApi>`
+impl InnerConnectApiWrapper {
+    /// Create a new `InnerConnectApiWrapper` from `Arc<dyn ConnectApi>`
     pub(crate) fn new_from_arc(connect: Arc<dyn InnerConnectApi>) -> Self {
         Self(connect)
     }
 
-    /// Create a new `ConnectApiWrapper` from id and addrs
+    /// Create a new `InnerConnectApiWrapper` from id and addrs
     pub(crate) async fn connect(
         id: ServerId,
         mut addrs: Vec<String>,
@@ -205,7 +205,7 @@ impl ConnectApiWrapper {
                 .await;
         }
         let client = InnerProtocolClient::new(channel);
-        let connect = ConnectApiWrapper::new_from_arc(Arc::new(Connect {
+        let connect = InnerConnectApiWrapper::new_from_arc(Arc::new(Connect {
             id,
             rpc_connect: client,
             change_tx,
@@ -215,13 +215,13 @@ impl ConnectApiWrapper {
     }
 }
 
-impl Debug for ConnectApiWrapper {
+impl Debug for InnerConnectApiWrapper {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("ConnectApiWrapper").finish()
+        f.debug_struct("InnerConnectApiWrapper").finish()
     }
 }
 
-impl Deref for ConnectApiWrapper {
+impl Deref for InnerConnectApiWrapper {
     type Target = Arc<dyn InnerConnectApi>;
 
     fn deref(&self) -> &Self::Target {
--- a/curp/src/rpc/connect.rs
+++ b/curp/src/rpc/connect.rs
@@ -46,10 +46,25 @@ const DEFAULT_BUFFER_SIZE: usize = 1024;
 /// Connect implementation
 macro_rules! connect_impl {
     ($client:ty, $api:path, $members:ident) => {
-        futures::future::join_all($members.into_iter().map(|(id, mut addrs)| async move {
+        futures::future::join_all(
+            $members
+                .into_iter()
+                .map(|(id, mut addrs)| single_connect_impl!($client, $api, id, addrs)),
+        )
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>, tonic::transport::Error>>()
+        .map(IntoIterator::into_iter)
+    };
+}
+
+/// Single Connect implementation
+macro_rules! single_connect_impl {
+    ($client:ty, $api:path, $id:ident, $addrs:ident) => {
+        async move {
             let (channel, change_tx) = Channel::balance_channel(DEFAULT_BUFFER_SIZE);
             // Addrs must start with "http" to communicate with the server
-            for addr in &mut addrs {
+            for addr in &mut $addrs {
                 if !addr.starts_with("http://") {
                     addr.insert_str(0, "http://");
                 }
@@ -60,17 +75,13 @@ macro_rules! connect_impl {
             }
             let client = <$client>::new(channel);
             let connect: Arc<dyn $api> = Arc::new(Connect {
-                id,
+                $id,
                 rpc_connect: client,
                 change_tx,
-                addrs: Mutex::new(addrs),
+                addrs: Mutex::new($addrs),
             });
-            Ok((id, connect))
-        }))
-        .await
-        .into_iter()
-        .collect::<Result<Vec<_>, tonic::transport::Error>>()
-        .map(IntoIterator::into_iter)
+            Ok(($id, connect))
+        }
     };
 }
 
@@ -193,25 +204,9 @@ impl InnerConnectApiWrapper {
         id: ServerId,
         mut addrs: Vec<String>,
     ) -> Result<Self, tonic::transport::Error> {
-        let (channel, change_tx) = Channel::balance_channel(DEFAULT_BUFFER_SIZE);
-        // Addrs must start with "http" to communicate with the server
-        for addr in &mut addrs {
-            if !addr.starts_with("http://") {
-                addr.insert_str(0, "http://");
-            }
-            let endpoint = Endpoint::from_shared(addr.clone())?;
-            let _ig = change_tx
-                .send(tower::discover::Change::Insert(addr.clone(), endpoint))
-                .await;
-        }
-        let client = InnerProtocolClient::new(channel);
-        let connect = InnerConnectApiWrapper::new_from_arc(Arc::new(Connect {
-            id,
-            rpc_connect: client,
-            change_tx,
-            addrs: Mutex::new(addrs),
-        }));
-        Ok(connect)
+        let (_id, connect) =
+            single_connect_impl!(InnerProtocolClient<Channel>, InnerConnectApi, id, addrs).await?;
+        Ok(InnerConnectApiWrapper::new_from_arc(connect))
     }
 }
 
--- a/curp/src/rpc/connect.rs
+++ b/curp/src/rpc/connect.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
-    fmt::Debug,
+    fmt::{Debug, Formatter},
+    ops::Deref,
     sync::Arc,
     time::Duration,
 };
@@ -85,8 +86,9 @@ pub(crate) async fn connect(
 /// Convert a vec of addr string to a vec of `InnerConnect`
 pub(crate) async fn inner_connect(
     members: HashMap<ServerId, Vec<String>>,
-) -> Result<impl Iterator<Item = (ServerId, Arc<dyn InnerConnectApi>)>, tonic::transport::Error> {
+) -> Result<impl Iterator<Item = (ServerId, ConnectApiWrapper)>, tonic::transport::Error> {
     connect_impl!(InnerProtocolClient<Channel>, InnerConnectApi, members)
+        .map(|iter| iter.map(|(id, connect)| (id, ConnectApiWrapper::new_from_arc(connect))))
 }
 
 /// Connect interface between server and clients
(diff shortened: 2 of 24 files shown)


curp/src/server/raw_curp/mod.rs
```diff
@@ -328,6 +342,68 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+        &self,
+        conf_change: ConfChangeEntry,
+    ) -> ((Option<ServerId>, u64), Result<(), ConfChangeError>) {
+        debug!(
+            "{} gets conf change for with id {}",
+            self.id(),
+            conf_change.id()
+        );
+
+        let st_r = self.st.read();
```
rogercloud, line 356: Shall we block all the other commands from fast path by putting conf_change to the spec_pool, and make it conflict with all the other commands?

curp/src/rpc/connect.rs
```diff
@@ -167,6 +177,58 @@ pub(crate) trait InnerConnectApi: Send + Sync + 'static {
     ) -> Result<tonic::Response<InstallSnapshotResponse>, RpcError>;
 }
 
+/// Connect Api Wrapper
+/// The solution of [rustc bug](https://github.com/dtolnay/async-trait/issues/141#issuecomment-767978616)
+#[derive(Clone)]
+pub(crate) struct ConnectApiWrapper(Arc<dyn InnerConnectApi>);
```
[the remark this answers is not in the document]
themanforfree: Removing "Wrapper" directly will conflict with the `ConnectApi` trait, and I don't think it exposes the implementation details. But this name will be ambiguous, so I renamed it to `InnerConnectApiWrapper `

curp/src/server/raw_curp/log.rs
```diff
@@ -306,6 +307,20 @@ impl<C: 'static + Command> Log<C> {
         Ok(entry)
     }
 
+    /// Pack the cmd into a log entry and push it to the end of the log, return its index
+    pub(super) fn push_conf_change(
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/server/raw_curp/mod.rs
themanforfree: We can only insert `Command` into `spec_pool`, but `conf_change` is't a `Command`. and the `conf_change` vertex is conflict with other vertexes, their execution order will not be swapped.

curp/src/server/raw_curp/state.rs
```diff
@@ -221,29 +222,28 @@ impl MajorityConfig {
...
+    fn quorum(&self) -> usize {
+        self.voters.len() / 2 + 1
     }
+}
 
-    /// Remove a voter
-    pub(super) fn remove(&mut self, id: ServerId) -> bool {
-        self.voters.remove(&id)
-    }
+impl Deref for MajorityConfig {
```
bsbds, line 236: Why do we need Deref here? I believe it should only be used for smart pointers.

curp/src/rpc/connect.rs
```diff
@@ -167,6 +177,58 @@ pub(crate) trait InnerConnectApi: Send + Sync + 'static {
     ) -> Result<tonic::Response<InstallSnapshotResponse>, RpcError>;
 }
 
+/// Inner Connect Api Wrapper
+/// The solution of [rustc bug](https://github.com/dtolnay/async-trait/issues/141#issuecomment-767978616)
```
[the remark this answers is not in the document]
themanforfree: The `ConnectApi` did not trigger this rustc bug, so it doesn't need to be wrapped.

curp/src/rpc/connect.rs
```diff
@@ -167,6 +177,58 @@ pub(crate) trait InnerConnectApi: Send + Sync + 'static {
...
+        }
+        let client = InnerProtocolClient::new(channel);
+        let connect = InnerConnectApiWrapper::new_from_arc(Arc::new(Connect {
+            id,
+            rpc_connect: client,
+            change_tx,
+            addrs: Mutex::new(addrs),
+        }));
+        Ok(connect)
+    }
```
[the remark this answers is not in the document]
themanforfree: We need it when a new node join to cluster

curp/src/rpc/mod.rs
```diff
@@ -449,11 +436,12 @@ impl FetchReadStateResponse {
...
 }
 
-#[allow(dead_code)] // TODO: remove this when we implement conf change
 #[allow(clippy::as_conversions)] // ConfChangeType is so small that it won't exceed the range of i32 type.
 impl ConfChange {
     /// Create a new `ConfChange` to add a node
-    pub(crate) fn add(node_id: ServerId, address: String) -> Self {
+    #[must_use]
+    #[inline]
+    pub fn add(node_id: ServerId, address: String) -> Self {
```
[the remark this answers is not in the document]
themanforfree: The node ID is calculate by client, only need the new node address and current timestamp

curp/src/rpc/mod.rs
```diff
@@ -480,13 +472,72 @@ impl ConfChange {
     }
 
     /// Create a new `ConfChange` to add a learner node
-    pub(crate) fn add_learner(node_id: ServerId, address: String) -> Self {
+    #[must_use]
+    #[inline]
+    pub fn add_learner(node_id: ServerId, address: String) -> Self {
```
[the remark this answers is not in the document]
themanforfree: Same as above

curp/src/rpc/connect.rs
```diff
@@ -234,31 +291,45 @@ impl ConnectApi for Connect<ProtocolClient<Channel>> {
         client.propose(req).await.map_err(Into::into)
     }
 
-    /// Send `WaitSyncedRequest`
-    #[instrument(skip(self), name = "client propose")]
-    async fn wait_synced(
+    /// Send `ShutdownRequest`
+    async fn shutdown(
```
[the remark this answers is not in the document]
themanforfree: Added

curp/src/server/cmd_board.rs
```diff
@@ -83,6 +92,16 @@ impl<C: Command> CommandBoard<C> {
         self.notify_asr(id);
     }
 
+    /// Insert er to internal buffer
+    pub(super) fn insert_conf(&mut self, id: &ProposeId, conf_r: Result<bool, ConfChangeError>) {
+        assert!(
+            self.conf_buffer.insert(id.clone(), conf_r).is_none(),
+            "er should not be inserted twice"
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/server/cmd_board.rs
```diff
@@ -171,4 +210,18 @@ impl<C: Command> CommandBoard<C> {
...
         }
     }
+
+    /// Wait for an execution result
+    pub(super) async fn wait_for_conf(
+        cb: &CmdBoardRef<C>,
+        id: &ProposeId,
+    ) -> Result<bool, ConfChangeError> {
+        loop {
+            if let Some(er) = cb.map_read(|cb_r| cb_r.conf_buffer.get(id).cloned()) {
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/server/cmd_worker/mod.rs
```diff
@@ -160,7 +159,20 @@ async fn worker_as<
...
             true
         }
-        EntryData::ConfChange(_) => false, // TODO: implement conf change
+        EntryData::ConfChange(ref conf_change) => {
+            let changes = conf_change.changes().to_owned();
+            let res = curp.apply_conf_change(changes).await;
+            let is_ok = res.is_ok();
+            let shutdown_self = res.as_ref().is_ok_and(|r| *r);
+            if let Err(e) = ce.set_last_applied(entry.index) {
+                error!("failed to set last_applied, {e}");
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/server/raw_curp/state.rs
themanforfree: Simplified operation, `cfg.voters.insert(id)` -> `cfg.insert(id)`

curp/src/rpc/mod.rs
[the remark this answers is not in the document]
iGxnon: does the `curp client` calculate it?

etcd api:

```protobuf
message MemberAddRequest {
  // peerURLs is the list of URLs the added member will use to communicate with the cluster.
  repeated string peerURLs = 1;
  // isLearner indicates if the added member is raft learner.
  bool isLearner = 2;
}
```

curp/src/server/raw_curp/state.rs
bsbds: See https://doc.rust-lang.org/std/ops/trait.Deref.html

> Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and [DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, Deref should only be implemented for smart pointers to avoid confusion.

curp/src/rpc/mod.rs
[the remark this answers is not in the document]
themanforfree: For etcd: ClusterServer will calculate it

curp/src/members.rs
```diff
@@ -107,12 +83,7 @@ impl ClusterInfo {
             if name == self_name {
                 member_id = id;
             }
-            let member = Member {
-                id,
-                name,
-                addrs,
-                is_learner: false, // TODO restore learner here?
```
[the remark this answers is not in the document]
themanforfree: we need't restore it here, this method only used to init cluster, there will be no learner at this time

curp/src/rpc/mod.rs
```diff
@@ -500,3 +551,68 @@ impl ShutdownResponse {
...
+    pub(crate) fn new_propose(error: ProposeError) -> Self {
+        Self::Propose(error.into())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[test]
+    fn test_conf_change_constructor() {
```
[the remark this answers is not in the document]
themanforfree: Removed

curp/src/server/raw_curp/mod.rs
```diff
@@ -868,95 +964,38 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
     }
 
     /// Apply conf changes and return true if self node is removed
-    #[allow(unused)] // TODO: remove this when we implement conf change
-    pub(super) fn apply_conf_change(
+    pub(super) async fn apply_conf_change(
```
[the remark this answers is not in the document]

curp/src/server/raw_curp/state.rs
themanforfree: Fixed

curp/src/server/raw_curp/mod.rs
```diff
@@ -328,6 +342,68 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+        }
+
+        if !self
+            .ctx
+            .cb
+            .map_write(|mut cb_w| cb_w.sync.insert(conf_change.id().clone()))
+        {
+            return (
+                info,
+                Err(ConfChangeError::new_propose(ProposeError::Duplicated)),
```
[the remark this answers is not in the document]
themanforfree: The new leader will recover this by `recover_from_spec_pools`

curp/src/server/curp_node.rs
```diff
@@ -354,60 +370,135 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
...
+            ));
+
+            while let Some(h) = sync_task_rx.recv().await {
+                let leader_retired = h
+                    .await
+                    .unwrap_or_else(|e| unreachable!("sync follower task panicked: {e}"));
+                if leader_retired {
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
themanforfree: Since the sync task is mutable, we can't join all task handles at the same time. In the current implementation, whenever the current handle of sync task returns true, it will break and not wait for another task handle. It may not be the fastest, but it doesn't require block too long.

curp/src/rpc/connect.rs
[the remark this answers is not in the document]
iGxnon: use `update_addrs` instead?

curp/tests/server.rs
```diff
@@ -379,3 +379,64 @@ async fn propose_remove_node_failed() {
...
+        .config(ClientConfig::default())
+        .build_from_all_members(group.all(), Some(follower_id))
+        .await
+        .unwrap();
+
+    let id = generate_propose_id("test");
+    let node_id = group.nodes.keys().next().copied().unwrap();
+    let changes = vec![ConfChange::update(node_id, "new_addr".to_owned())];
+    let conf_change = ProposeConfChangeRequest::new(id, changes);
+    let members = client
```
Phoenix500526, line 433: It seems you propose the `ProposeConfChangeRequest` to a follower, not a learner. Is it correct?

curp/src/server/raw_curp/mod.rs
```diff
@@ -328,6 +342,68 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         (info, Ok(()))
     }
 
+    /// Handle `propose_conf_change` request
+    pub(super) fn handle_propose_conf_change(
```
Phoenix500526, line 346: This method looks similar to 'handle_propose`. Can we extract the common part of these two methods as a new method to avoid duplication?

Phoenix500526 (changes requested):

curp/src/rpc/connect.rs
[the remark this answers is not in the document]
themanforfree: add a new node, not update an existing node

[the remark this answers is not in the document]
iGxnon: then use the macros `connect_impl` ?

curp/src/server/cmd_board.rs
```diff
@@ -83,6 +92,16 @@ impl<C: Command> CommandBoard<C> {
         self.notify_asr(id);
     }
 
+    /// Insert er to internal buffer
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/tests/server.rs
```diff
@@ -291,3 +297,84 @@ async fn shutdown_rpc_should_shutdown_the_cluster() {
...
+        .unwrap()
+        .unwrap();
+    assert_eq!(members.len(), 5);
+    let member = members.iter().find(|m| m.id == node_id);
+    assert!(member.is_some_and(|m| &m.addrs == &["new_addr"]));
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[abort_on_panic]
+async fn propose_remove_node_failed() {
```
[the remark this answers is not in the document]
themanforfree: Renamed

curp/tests/server.rs
themanforfree: the test name should be `follower`, Fixed

curp/src/server/raw_curp/mod.rs
themanforfree: Fixed

Phoenix500526 (commented):
LGTM

curp/src/server/raw_curp/mod.rs
rogercloud: If conf_change can't be inserted into `spec_pool`, does it go through the CURP protocol? I think it should.

curp/src/server/cmd_worker/mod.rs
```diff
@@ -156,11 +155,26 @@ async fn worker_as<
...
             true
         }
-        EntryData::ConfChange(_) => false, // TODO: implement conf change
+        EntryData::ConfChange(ref conf_change) => {
+            let changes = conf_change.changes().to_owned();
+            if let Err(e) = ce.set_last_applied(entry.index) {
+                error!("failed to set last_applied, {e}");
+                return false;
+            }
+            let res = curp.apply_conf_change(changes);
```
bsbds, line 169: It seems the server only swith to new conf change in after sync stage, and the follower node does not switch to new conf immediately after it received a conf change log?

themanforfree: Yes, this implementation is not exactly the same as the raft paper, we use similar logic as etcd.

curp/src/server/raw_curp/mod.rs
```diff
@@ -883,79 +885,30 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+    pub(super) fn connects(&self) -> &DashMap<ServerId, InnerConnectApiWrapper> {
+        &self.ctx.connects
+    }
+
+    /// Insert connect
+    pub(super) fn insert_connect(&self, connect: InnerConnectApiWrapper) {
+        let _ig = self.ctx.connects.insert(connect.id(), connect);
+    }
+
+    /// Get connect
```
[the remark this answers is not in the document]
themanforfree: Updated

curp/src/rpc/connect.rs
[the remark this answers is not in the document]
themanforfree: FIxed

curp/src/server/raw_curp/mod.rs
themanforfree: Curp itself does not have the ability of membership change. This ability is provided by the back-end raft. It can be understood that whether or not `conf_change` conflicts with other requests, it goes directly into slow path and waits for the completion of the back-end raft protocol before returning to the client. So there is no need to insert `conf_change` into the `spec_pool`

curp/src/server/cmd_worker/mod.rs
bsbds: In this implementation, how does the leader guarantee there's no overlapped configuration change in etcd?

themanforfree: The leader will confirm that the follower has been applied conf change via commit index, Therefore, the leader does not resend conf change, resulting in overlap.

Phoenix500526 (approved):

Phoenix500526 (approved):
LGTM

iGxnon (dismissed):
LGTM

curp/src/server/cmd_board.rs
```diff
@@ -25,7 +22,7 @@ pub(super) struct CommandBoard<C: Command> {
     /// Store all notifiers for conf change results
     conf_notifier: HashMap<ProposeId, Event>,
     /// Store all conf change results
```
Phoenix500526, line 24: Please update the comment.

curp/src/server/cmd_worker/mod.rs
```diff
@@ -155,25 +156,26 @@ async fn worker_as<
...
                 return false;
             }
-            let res = curp.apply_conf_change(changes);
-            let is_ok = res.is_ok();
-            let shutdown_self = res.as_ref().is_ok_and(|r| *r);
-            cb.write().insert_conf(entry.id(), res);
+            let change = conf_change
+                .changes()
+                .first()
+                .unwrap_or_else(|| unreachable!("conf change should always have at one change"));
```
Phoenix500526, line 171: typo. "conf change should always have at least one change"

curp/tests/server.rs
```diff
@@ -374,6 +374,7 @@ async fn propose_remove_node_should_failed_when_cluster_nodes_equals_to_three()
     let changes = vec![ConfChange::remove(node_id)];
     let conf_change = ProposeConfChangeRequest::new(id, changes);
     let res = client.propose_conf_change(conf_change).await.unwrap();
+    println!("{res:?}");
```
Phoenix500526, line 377: Please remove this debug print.

curp/src/server/raw_curp/mod.rs
```diff
@@ -1182,17 +1205,16 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
 
     /// Switch to a new config and return true if self node is removed
```
Phoenix500526, line 1275: Please update the comment here.

curp/src/server/raw_curp/log.rs
```diff
@@ -173,6 +205,12 @@ impl<C: Command> Debug for Log<C> {
     }
 }
 
+// TODO
```
Phoenix500526, line 208: TODO what?

curp/src/server/cmd_board.rs
```diff
@@ -25,7 +22,7 @@ pub(super) struct CommandBoard<C: Command> {
...
-    pub(super) conf_buffer: IndexMap<ProposeId, Result<bool, ConfChangeError>>,
+    pub(super) conf_buffer: IndexSet<ProposeId>,
```
Phoenix500526, line 25: Why should the conf_buffer be an IndexSet?

themanforfree: Because the result of conf change is already merged into `handle_propose_conf_change`, we don't need pass this result from `Filter` to `propose_conf_change`

curp/src/server/raw_curp/log.rs
themanforfree: Removed

curp/src/server/raw_curp/mod.rs
themanforfree: Updated

curp/tests/server.rs
themanforfree: Removed

curp/src/server/cmd_worker/mod.rs
themanforfree: Fixed

curp/src/server/cmd_board.rs
themanforfree: Updated

curp/src/server/cmd_worker/mod.rs
```diff
@@ -160,7 +160,22 @@ async fn worker_as<
...
             true
         }
-        EntryData::ConfChange(_) => false, // TODO: implement conf change
+        EntryData::ConfChange(ref conf_change) => {
+            if let Err(e) = ce.set_last_applied(entry.index) {
+                error!("failed to set last_applied, {e}");
+                return false;
+            }
+            let change = conf_change.changes().first().unwrap_or_else(|| {
+                unreachable!("conf change should always have at least one change")
```
iGxnon, line 169: In ClusterServer, using an empty conf change reads out members linearly, and there seems to be an issue here.

iGxnon: Could we just return true here?

curp/src/server/raw_curp/mod.rs
```diff
@@ -153,6 +165,12 @@ struct Context<C: Command, RC: RoleChange> {
     leader_event: Arc<Event>,
     /// Leader change callback
     role_change: RC,
+    /// Conf change tx, used to update sync tasks
+    change_tx: flume::Sender<ConfChange>,
+    /// Conf change tx, used to update sync tasks
```
iGxnon, line 170: Conf change rx

curp/src/server/cmd_worker/mod.rs
themanforfree: We shouldn't use it like you described, after we support joint consensus in the future, empty changes means leave joint.

curp/src/server/raw_curp/mod.rs

curp/src/server/raw_curp/mod.rs
```diff
@@ -369,27 +392,39 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+            };
+            let changes = conf_change.changes().to_owned();
+            self.fallback_conf_change(changes, info.addrs, info.name, info.is_learner);
+        }
+        // apply conf change entries
+        for e in cc_entries {
+            let EntryData::ConfChange(ref cc) = e.entry_data else {
+                unreachable!("cc_entry should be conf change entry");
+            };
+            let (addrs,name,is_learner) = self.apply_conf_change(cc.changes().to_owned())
```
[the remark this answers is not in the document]
themanforfree: Added

curp/src/server/raw_curp/log.rs
```diff
@@ -252,7 +292,14 @@ impl<C: 'static + Command> Log<C> {
             {
                 continue;
             }
-
+            for e in self.entries.range(pi..) {
+                if matches!(e.entry_data, EntryData::ConfChange(_)) {
+                    let _ig = need_fallback_indexes.insert(e.index);
```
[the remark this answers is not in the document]
themanforfree: Yes, only first truncation will enter this loop. the latter `self.entries.truncate(pi)` has the same problem. I think the better way is refactor the `try_append_entries`, just find the truncate index and truncate once, I will create a ISSUE to track this, It will not be fixed in this PR

curp/src/server/raw_curp/mod.rs
```diff
@@ -668,11 +705,15 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
                 role_change,
+                change_tx,
+                change_rx,
+                connects,
             },
             shutdown_trigger,
         };
         if is_leader {
             let mut st_w = raw_curp.st.write();
+            st_w.term = 1;
```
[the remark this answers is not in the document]
themanforfree: `term = 0` is the initialization status of the cluster, it should not has a leader

curp/src/server/raw_curp/tests.rs
```diff
@@ -10,6 +10,8 @@ use utils::config::{
 };
 
 use super::*;
+use crate::rpc::connect::MockInnerConnectApi;
+use crate::ProposeConfChangeRequest;
```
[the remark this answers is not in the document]
themanforfree: Merged

curp/src/server/gc.rs
```diff
@@ -143,28 +143,23 @@ mod tests {
 
         tokio::time::sleep(Duration::from_millis(100)).await;
         let cmd1 = Arc::new(TestCommand::default());
-        spec.lock()
-            .pool
-            .insert(cmd1.id().clone(), Arc::clone(&cmd1));
+        spec.lock().pool.insert(cmd1.id().clone(), cmd1.into());
```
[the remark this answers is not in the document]
themanforfree: Added

Phoenix500526 (approved):
LGTM

iGxnon (approved):
LGTM

PR #441, opened 2023-09-05
fix: `wait_for_er` & `wait_for_er_asr` loops without a limit

iGxnon:

    `leader_retires` will invoke `CommandBoard::clear`, which will clear all command results in command board and then notify those wait task. 

    ```rust
    pub(super) fn release_notifiers(&mut self) {
        self.er_notifiers
            .drain()
            .for_each(|(_, event)| event.notify(usize::MAX));
        self.asr_notifiers
            .drain()
            .for_each(|(_, event)| event.notify(usize::MAX));
    }

    /// Clear
    pub(super) fn clear(&mut self) {
        self.er_buffer.clear();
        self.asr_buffer.clear();
        self.release_notifiers();
    }
    ```

    ```rust
    pub(super) async fn wait_for_er(
        cb: &CmdBoardRef<C>,
        id: &ProposeId,
    ) -> Result<C::ER, C::Error> {
        loop {
            if let Some(er) = cb.map_read(|cb_r| cb_r.er_buffer.get(id).cloned()) {
                return er;
            }
            let listener = cb.write().er_listener(id);
            listener.await;
        }
    }
    ```

    However, because it cannot obtain C::ER, it will once again enter a wait state and will not recover.
    Under certain extreme conditions, such as rapid GC of command board, this situation can also occur.

   `wait_for_xxx` will only wait once and return an `Option<...>`

### DIFF
--- a/curp/src/server/cmd_board.rs
+++ b/curp/src/server/cmd_board.rs
@@ -89,11 +89,7 @@ impl<C: Command> CommandBoard<C> {
             .er_notifiers
             .entry(id.clone())
             .or_insert_with(Event::new);
-        let listener = event.listen();
-        if self.er_buffer.contains_key(id) {
-            event.notify(usize::MAX);
-        }
-        listener
+        event.listen()
     }
 
     /// Get a listener for shutdown
@@ -107,11 +103,7 @@ impl<C: Command> CommandBoard<C> {
             .asr_notifiers
             .entry(id.clone())
             .or_insert_with(Event::new);
-        let listener = event.listen();
-        if self.asr_buffer.contains_key(id) {
-            event.notify(usize::MAX);
-        }
-        listener
+        event.listen()
     }
 
     /// Notify execution results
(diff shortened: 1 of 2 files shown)


curp/src/server/cmd_board.rs
```diff
@@ -120,36 +112,56 @@ impl<C: Command> CommandBoard<C> {
...
-            listener.await;
+    ) -> Option<(Result<C::ER, C::Error>, Option<Result<C::ASR, C::Error>>)> {
+        match cb.map_read(|cb_r| cb_r.get_er_asr(id)) {
+            (Some(er), None) if er.is_err() => return Some((er, None)),
+            (Some(er), Some(asr)) => return Some((er, Some(asr))),
+            _ => {}
+        }
+        let listener = cb.write().asr_listener(id);
+        listener.await;
+        match cb.map_read(|cb_r| cb_r.get_er_asr(id)) {
```
Phoenix500526, line 161: This `match` block is duplicated to that one in line 154.

curp/src/server/curp_node.rs
```diff
@@ -189,7 +191,9 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
         let id = req.id()?;
         debug!("{} get wait synced request for cmd({id})", self.curp.id());
 
-        let (er, asr) = CommandBoard::wait_for_er_asr(&self.cmd_board, &id).await;
+        let Some((er, asr)) = CommandBoard::wait_for_er_asr(&self.cmd_board, &id).await else {
+            return Ok(WaitSyncedResponse::new_from_result::<C>(None, None)?);
```
Phoenix500526, line 195: Why should we consider this case as a result rather than an error?

iGxnon: The result `(None, None)` would be converted into `WaitSyncError::Other("can't get er result".to_owned())`. Return `Err()` will be convert into `tonic::Status::internal` error. The `Err()` return will result in a retry in client and continuously fail to obtain results and keep retrying.

 Error handling across RPC looks a bit messy, and some parts don't comply with Rust's conventions. I think the overall error-handling code needs to be refactored.

Phoenix500526: Agree

curp/src/server/cmd_board.rs
```diff
@@ -133,18 +125,38 @@ impl<C: Command> CommandBoard<C> {
...
         id: &ProposeId,
-    ) -> Result<C::ER, C::Error> {
-        loop {
-            if let Some(er) = cb.map_read(|cb_r| cb_r.er_buffer.get(id).cloned()) {
-                return er;
-            }
-            let listener = cb.write().er_listener(id);
-            listener.await;
+    ) -> Option<Result<C::ER, C::Error>> {
+        if let Some(er) = cb.map_read(|cb_r| cb_r.get_er(id)) {
```
bsbds, line 154: could we just remove L154-L156?

curp/src/server/cmd_board.rs
```diff
@@ -157,18 +169,18 @@ impl<C: Command> CommandBoard<C> {
...
+            (Some(er), Some(asr)) => return Some((er, Some(asr))),
+            _ => {}
+        }
+        let listener = cb.write().asr_listener(id);
+        listener.await;
+        match cb.map_read(|cb_r| cb_r.get_er_asr(id)) {
+            (Some(er), None) if er.is_err() => Some((er, None)),
+            (Some(er), Some(asr)) => Some((er, Some(asr))),
+            _ => None,
         }
```
themanforfree, line 184: Could we let `get_er_asr` return `Option<(Result<C::ER, C::Error>, Option<Result<C::ASR, C::Error>>)` directly?

curp/src/server/cmd_board.rs
iGxnon: The listener will only be notified during the insert result. If a wait sync request misses a notification once, it will continue to wait at L158

iGxnon:
The current solution still causes blockage. I planned to fix this issue in #450.
Based on #450, it can be determined whether a request is currently being executed, thus eliminating the problem of wait sync blocking for outdated requests.

PR #449, opened 2023-09-14
Add learner

themanforfree:
#306 

### DIFF
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -318,7 +318,7 @@ mod tests {
         let node1_url = node1.self_addrs();
         assert!(!peers.contains_key(&node1_id));
         assert_eq!(peers.len(), 2);
-        assert_eq!(node1.voters_len(), peers.len() + 1); // TODO fix test
+        assert_eq!(node1.voters_len(), peers.len() + 1);
 
         let peer_urls = peers.values().collect::<Vec<_>>();
 
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -106,6 +106,13 @@ impl ClusterInfo {
             .collect()
     }
 
+    /// Get all members vec
+    #[must_use]
+    #[inline]
+    pub fn all_members_vec(&self) -> Vec<Member> {
+        self.members.iter().map(|t| t.value().clone()).collect()
+    }
+
     /// Insert a member
     #[inline]
     pub fn insert(&self, member: Member) {
@@ -241,18 +248,11 @@ impl ClusterInfo {
             .collect()
     }
 
-    /// Get all members
-    #[must_use]
-    #[inline]
-    pub fn members(&self) -> Vec<Member> {
-        self.members.iter().map(|t| t.value().clone()).collect()
-    }
-
     /// Get length of peers
     #[must_use]
     #[inline]
-    pub fn members_len(&self) -> usize {
-        self.members.len()
+    pub fn voters_len(&self) -> usize {
+        self.members.iter().filter(|t| !t.is_learner).count()
     }
 
     /// Get id by name
@@ -304,7 +304,7 @@ mod tests {
         let node1_url = node1.self_addrs();
         assert!(!peers.contains_key(&node1_id));
         assert_eq!(peers.len(), 2);
-        assert_eq!(node1.members_len(), peers.len() + 1);
+        assert_eq!(node1.voters_len(), peers.len() + 1); // TODO fix test
 
         let peer_urls = peers.values().collect::<Vec<_>>();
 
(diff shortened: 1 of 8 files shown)


curp/src/server/raw_curp/mod.rs
```diff
@@ -310,6 +313,16 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
                 Err(ConfChangeError::new_propose(ProposeError::NotLeader)),
             );
         }
+        if let Some(change) = conf_change.changes().first() {
+            if let ConfChangeType::Promote = change.change_type() {
+                let learner_index = self.lst.get_match_index(change.node_id);
+                let leader_index = self.log.read().last_log_index();
+                if leader_index.overflow_sub(learner_index) > MAX_PROMOTE_GAP {
+                    return (info, Err(ConfChangeError::LearnerNotCatchUp(())));
+                }
```
bsbds, line 322: Should the learner be automatically promoted? Or can the user force promote a learner without satisfying the gap?

curp/src/server/raw_curp/state.rs
```diff
@@ -210,22 +241,61 @@ pub(super) struct MajorityConfig {
...
+        let res1 = self.majority_config.voters.remove(&id);
+        let res2 = self.learners.remove(&id);
+        debug_assert_ne!(
+            res1, res2,
+            "a node should not exist in both voters and learners"
+        );
+        res1 || res2
+    }
+
+    /// Check if a voter exists
```
bsbds, line 287: the "voter" should be "server"

curp/src/server/curp_node.rs
```diff
@@ -723,21 +725,24 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
     async fn bcast_vote(curp: &RawCurp<C, RC>, vote: Vote) {
         debug!("{} broadcasts votes to all servers", curp.id());
         let rpc_timeout = curp.cfg().rpc_timeout;
+        let voters = curp.voters();
```
bsbds, line 728: Should we hold the voters lock here? The conf might changed before we broadcast vote.

curp/src/server/raw_curp/mod.rs
```diff
@@ -310,6 +313,16 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+            }
+        }
```
iGxnon, line 324: Should these be checked in `check_new_config`?

iGxnon: It seems that the conf change was inserted into logs before we checked for failures. Is there a problem with this?

curp/src/server/raw_curp/mod.rs
themanforfree: Currently, we need to do promote manually. In some scenarios, the user may want to keep the node in the learner state as a read-only or standby node. and learner should.
For the second question, We can add this feature, but why do we need it? if user want to add a new node and do not care about this gap, they and directly add a voter node.

curp/src/server/raw_curp/state.rs
themanforfree: Fixed

curp/src/server/curp_node.rs
themanforfree: Fixed, increase the scope of voter lock

curp/src/members.rs
```diff
@@ -304,7 +311,7 @@ mod tests {
         let node1_url = node1.self_addrs();
         assert!(!peers.contains_key(&node1_id));
         assert_eq!(peers.len(), 2);
-        assert_eq!(node1.members_len(), peers.len() + 1);
+        assert_eq!(node1.voters_len(), peers.len() + 1); // TODO fix test
```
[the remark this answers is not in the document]
themanforfree: Forgot to remove, Removed

curp/src/server/raw_curp/mod.rs

curp/src/server/raw_curp/mod.rs
bsbds: makes sense, thx.

Phoenix500526 (approved):
LGTM

curp/src/server/curp_node.rs
bsbds: I mean maybe we need to forbid apply conf change during the entire vote procedure. Because if the conf is applied during the `handle_vote_resp`, the vote result may be incorrect.

themanforfree: It seems impossible to vote or apply conf change at the same time, if a candidate receive an conf change log, this node will become a follower and apply this conf change, the vote response will be ignored.

bsbds (approved):

PR #453, opened 2023-09-21
refactor: use `Logindex` instead of `ProposeId` in `cmd_vid`

themanforfree:

In the current implementation, all log entries require propose id, but some log, such as shutdown and no-op log that will be introduced in the future, do not need propose id.
 use Logindex instead of ProposeId in cmd_vid, after this refactor, now the log entry no longer has to include ProposeId
no

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -371,7 +371,7 @@ where
 
     /// The shutdown rpc of curp protocol
     #[instrument(skip_all)]
-    pub async fn shutdown(&self, id: ProposeId) -> Result<(), ProposeError> {
+    pub async fn shutdown(&self) -> Result<(), ProposeError> {
         let retry_count = *self.config.retry_count();
         for _ in 0..retry_count {
             let leader_id = match self.get_leader_id().await {
@@ -386,7 +386,7 @@ where
                 .get_connect(leader_id)
                 .unwrap_or_else(|| unreachable!("leader {leader_id} not found"))
                 .shutdown(
-                    ShutdownRequest::new(id.clone()),
+                    ShutdownRequest::default(),
                     *self.config.wait_synced_timeout(),
                 )
                 .await
@@ -402,7 +402,11 @@ where
                 .write()
                 .check_and_update(resp.leader_id, resp.term);
             match resp.error {
-                Some(e) => return Err(bincode::deserialize(&e)?),
+                Some(e) => {
+                    // Leader may not be correct leader, resend the shutdown request to the new leader
+                    warn!("shutdown error: {:?}", e);
+                    continue;
+                }
                 None => return Ok(()),
             }
         }
--- a/curp/src/log_entry.rs
+++ b/curp/src/log_entry.rs
@@ -25,7 +25,7 @@ pub(crate) enum EntryData<C> {
     /// `ConfChange` entry
     ConfChange(ProposeId),
     /// `Shutdown` entry
-    Shutdown(ProposeId),
+    Shutdown,
 }
 
 impl<C> LogEntry<C>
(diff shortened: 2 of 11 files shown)


curp/src/rpc/mod.rs
```diff
@@ -489,3 +482,17 @@ impl ConfChange {
         }
     }
 }
+
+impl ShutdownResponse {
+    /// Create a new shutdown response
+    pub(crate) fn new(leader_id: Option<ServerId>, term: u64, error: Option<ProposeError>) -> Self {
+        let error = error.map(|e| PbProposeErrorOuter {
+            propose_error: Some(e.into()),
+        });
```
iGxnon, line 491: Could we implement `From` trait for PbProposeErrorOuter type?

curp/src/server/cmd_worker/conflict_checked_mpmc.rs
```diff
@@ -190,8 +191,8 @@ enum OnceState {
 /// Internally it maintains a dependency graph of conflicting cmds
 
 struct Filter<C: Command, CE> {
-    /// Index from `ProposeId` to `vertex`
-    cmd_vid: HashMap<ProposeId, u64>,
+    /// Index from `LogIndex` to `vertex`
+    cmd_vid: HashMap<LogIndex, u64>,
```
iGxnon, line 195: Will there be truncation and log overwriting behavior in Raft, and will this have an impact?

curp/src/client.rs
```diff
@@ -402,7 +402,10 @@ where
                 .write()
                 .check_and_update(resp.leader_id, resp.term);
             match resp.error {
-                Some(e) => return Err(bincode::deserialize(&e)?),
+                Some(e) => {
+                    warn!("shutdown error: {:?}", e);
+                    continue;
```
iGxnon, line 407: I think we should add a comment explaining why this error can be retried instead of returning it to the user.

curp/src/log_entry.rs
```diff
@@ -64,7 +64,13 @@ where
     pub(super) fn id(&self) -> &ProposeId {
         match self.entry_data {
             EntryData::Command(ref cmd) => cmd.id(),
-            EntryData::ConfChange(ref id) | EntryData::Shutdown(ref id) => id,
+            EntryData::ConfChange(ref id) => id,
```
iGxnon, line 67: Will the id of ConfChange also be removed in subsequent PRs?

curp/src/rpc/mod.rs
themanforfree: Fixed

curp/src/client.rs
themanforfree: Added

curp/src/log_entry.rs
themanforfree: No

curp/src/server/cmd_worker/conflict_checked_mpmc.rs
themanforfree: For leader node, logs will not be overwritten.
For follower node, only uncommitted logs can be overwritten. and  uncommitted logs will not send to mpmc conflict channel

Phoenix500526, line 195: We'd better rename this `cmd_vid` because it implicitly indicates the mapping relationship between a command id and a vertex id. But now this assumption is not valid. How about `log_vid` or `entry_vid`?

curp/src/log_entry.rs
themanforfree: Because `ConfChange` need to store it result to `cmd_board`, and `cmd_board` use `ProposeId` as index

curp/src/server/cmd_worker/conflict_checked_mpmc.rs
themanforfree: Renamed

bsbds (approved):

iGxnon (approved):

PR #459, opened 2023-09-24
Refactor/remove xline client

Phoenix500526:
In the previous releases, for testing purposes, we conducted functional tests on the lock server's RPC using the etcd client. However, starting from v0.4.1, since the xline-client has been preliminarily completed, we can re-implement some of the testing scenarios based on the xline-client.


xline-test-utils/src/bin/validation_lock_client.rs
```diff
@@ -1,7 +1,11 @@
 //! this binary is only used for the validation of lock service
 
 use clap::{Parser, Subcommand};
-use etcd_client::Client;
```
iGxnon, line 4: could we remove etcd_client dependency in Cargo.toml?

Phoenix500526: Removed

xline/src/bin/restore_client.rs
```diff
@@ -1,19 +0,0 @@
-use std::path::PathBuf;
```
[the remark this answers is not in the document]
Phoenix500526: This restore client is not used in xline.

bsbds:
I think we keep the original etcd-client, because if we only use xline-client, the servers in `xline/src/server` will not be tested in integration tests. Maybe add a trait so that we can test using both clients?

Phoenix500526:
> I think we keep the original etcd-client, because if we only use xline-client, the servers in `xline/src/server` will not be tested in integration tests. Maybe add a trait so that we can test using both clients?

The integration test only cares about the correctness of the xline server. Therefore, it doesn't matter what exactly the client is. The compatibility with etcd client can be covered in the validation test.

bsbds (approved):

scripts/quick_start.sh
```diff
@@ -59,7 +59,7 @@ run_container() {
     for ((i = 1; i <= ${size}; i++)); do
         docker run -d -it --rm --name=node${i} --net=xline_net --ip=${SERVERS[$i]} --cap-add=NET_ADMIN --cpu-shares=1024 -m=512M -v ${DIR}:/mnt ${image} bash &
     done
-    docker run -d -it --rm --name=node4 --net=xline_net --ip=${SERVERS[0]} --cap-add=NET_ADMIN --cpu-shares=1024 -m=512M -v ${DIR}:/mnt gcr.io/etcd-development/etcd:v3.5.5 bash &
+    docker run -d -it --rm --name=node4 --net=xline_net --ip=${SERVERS[0]} --cap-add=NET_ADMIN --cpu-shares=1024 -m=512M -v ${DIR}:/mnt ghcr.io/xline-kv/etcdctl:v3.5.9 bash &
```
iGxnon, line 62: Do we have this image available?

Phoenix500526: Yes, FYI: https://github.com/orgs/xline-kv/packages/container/package/etcdctl

Phoenix500526:
We don't need to construct two inner clients in a `BenchClient`. It's easy to solve this problem, so I leave a good first issue(#486 ) to track it.

iGxnon (approved):

PR #448, opened 2023-09-14
Stop removed node and start added node 

themanforfree:
Please briefly answer these questions:
#306 
* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #460, opened 2023-09-25
Client discover cluster

themanforfree:
described in #306 
based on #449 

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -243,6 +243,8 @@ enum UnpackStatus {
     IO,
     /// Internal error type
     Internal,
+    /// Wrong Cluster Version
+    WrongClusterVersion,
 }
 
 /// unpack `tonic::Status` and convert it to `UnpackStatus`
@@ -262,6 +264,7 @@ fn unpack_status(status: &tonic::Status) -> UnpackStatus {
             "internal" => UnpackStatus::Internal,
             "storage" => UnpackStatus::Storage,
             "io" => UnpackStatus::IO,
+            "wrong-cluster-version" => UnpackStatus::WrongClusterVersion,
             unsupported_label => {
                 unreachable!("unsupported status label {unsupported_label}")
             }
@@ -318,12 +321,21 @@ where
         while let Some(resp_result) = rpcs.next().await {
             let resp = match resp_result {
                 Ok(resp) => resp.into_inner(),
+
                 Err(e) => {
-                    if let UnpackStatus::ShuttingDown = unpack_status(&e) {
-                        return Err(ClientError::ShuttingDown);
-                    }
                     warn!("Propose error: {}", e);
-                    continue;
+                    match unpack_status(&e) {
+                        UnpackStatus::ShuttingDown => return Err(ClientError::ShuttingDown),
+                        UnpackStatus::WrongClusterVersion => {
+                            return Err(ClientError::WrongClusterVersion)
+                        }
+                        UnpackStatus::Redirect(..)
+                        | UnpackStatus::Transport
+                        | UnpackStatus::EncodeDecode
+                        | UnpackStatus::Storage
+                        | UnpackStatus::IO
+                        | UnpackStatus::Internal => continue,
+                    }
                 }
             };
             self.state
@@ -388,6 +400,9 @@ where
                     warn!("wait synced rpc error: {e}");
                     match unpack_status(&e) {
                         UnpackStatus::ShuttingDown => return Err(ClientError::ShuttingDown),
+                        UnpackStatus::WrongClusterVersion => {
+                            return Err(ClientError::WrongClusterVersion)
+                        }
                         UnpackStatus::Transport => {
                             // it's quite likely that the leader has crashed, then we should wait for some time and fetch the leader again
                             tokio::time::sleep(retry_timeout.next_retry()).await;
@@ -421,6 +436,30 @@ where
         Err(ClientError::Timeout)
     }
 
+    /// Set the information of current cluster
+    #[allow(unused)]
+    async fn set_cluster(&self, cluster: FetchClusterResponse) {
+        debug!("update client by remote cluster: {cluster:?}");
+        self.state
+            .write()
+            .check_and_update(cluster.leader_id, cluster.term);
+        let member_addrs = cluster
+            .members
+            .into_iter()
+            .map(|m| (m.id, m.addrs))
+            .collect::<HashMap<ServerId, Vec<String>>>();
+        self.connects.clear();
+        // TODO
+        #[allow(clippy::unwrap_used)]
+        for (id, connect) in rpc::connect(member_addrs).await.unwrap() {
+            let _ig = self.connects.insert(id, connect);
+        }
+        self.cluster_version.store(
+            cluster.cluster_version,
+            std::sync::atomic::Ordering::Relaxed,
+        );
+    }
+
     /// The shutdown rpc of curp protocol
     #[instrument(skip_all)]
     pub async fn shutdown(&self) -> Result<(), ClientError<C>> {
@@ -447,6 +486,9 @@ where
                 warn!("shutdown rpc error: {e}");
                 match unpack_status(&e) {
                     UnpackStatus::ShuttingDown => return Err(ClientError::ShuttingDown),
+                    UnpackStatus::WrongClusterVersion => {
+                        return Err(ClientError::WrongClusterVersion)
+                    }
                     UnpackStatus::Redirect(new_leader, term) => {
                         self.state.write().check_and_update(new_leader, term);
                         warn!("shutdown: redirect to new leader {new_leader:?}, term is {term}",);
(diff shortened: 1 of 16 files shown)


curp/src/rpc/mod.rs
```diff
@@ -507,7 +515,17 @@ impl ProposeConfChangeRequest {
     #[inline]
     #[must_use]
     pub fn new(id: String, changes: Vec<ConfChange>) -> Self {
-        Self { id, changes }
+        Self {
+            id,
+            changes,
+            cluster_version: 0,
```
iGxnon, line 521: please add `cluster_version` into the constructor parameter

curp/src/rpc/mod.rs
```diff
@@ -5,13 +5,14 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 pub use self::proto::{
     commandpb::{cmd_result::Result as CmdResultInner, CmdResult, ProposeRequest, ProposeResponse},
+    errorpb::Member,
```
iGxnon, line 8: Why `errorpb`?

curp/src/server/raw_curp/mod.rs
```diff
@@ -781,6 +790,21 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         self.ctx.cluster_info.as_ref()
     }
 
+    /// Get cluster info
+    pub(super) fn cluster_err(&self) -> Cluster {
```
iGxnon, line 794: cluster err?

curp/src/members.rs
```diff
@@ -227,6 +258,18 @@ impl ClusterInfo {
...
 
+    /// Get cluster version
+    #[must_use]
+    #[inline]
+    pub fn cluster_version(&self) -> u64 {
+        self.cluster_version.load(Ordering::Relaxed)
+    }
+
+    /// cluster version increase
+    pub(super) fn cluster_version_inc(&self) -> u64 {
```
iGxnon, line 269: pub(crate)?

curp/src/server/raw_curp/mod.rs
```diff
@@ -1369,6 +1411,7 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
                 (vec![], String::new(), false)
             }
         };
+        let _ig = self.ctx.cluster_info.cluster_version_inc();
```
iGxnon, line 1414: Should fall back the cluster version?

Suppose there are 10 conf change requests sent to a leader. Then the cluster version on this leader is updated to 10. A client fetches the cluster and updates its local cluster version to 10. Then the leader failed, and these conf change logs are reverted by a new elected leader. The above client with the cluster version of 10 will still keep the old cluster members and will not be updated quickly even the cluster members are actually changed.

curp/src/rpc/mod.rs
Phoenix500526: I have the same question.

curp/src/server/curp_node.rs
```diff
@@ -65,6 +65,9 @@ pub(super) enum CurpError {
     /// Internal error
     #[error("internal error, {0}")]
     Internal(String),
+    /// TODO
+    #[error("internal error, {0:?}")]
```
Phoenix500526: Please modify the error output message. It will be confusing when a user sees two errors with the totally same output message.  In addition, todo what?

curp/src/client.rs
```diff
@@ -385,6 +406,29 @@ where
...
+        self.state
+            .write()
+            .check_and_update(cluster.leader_id, cluster.term);
+        let member_addrs = cluster
+            .members
+            .into_iter()
+            .map(|m| (m.id, m.addrs))
+            .collect::<HashMap<ServerId, Vec<String>>>();
+        self.connects.clear();
+        // TODO
```
Phoenix500526: todo what?

curp/src/client.rs
```diff
@@ -443,16 +442,25 @@ where
...
                 .shutdown(
-                    ShutdownRequest::default(),
+                    ShutdownRequest::new(propose_id.clone(), self.cluster_version()),
                     *self.config.wait_synced_timeout(),
                 )
                 .await
             {
                 Ok(resp) => resp.into_inner(),
+                Err(RpcError::StatusError(status)) => {
+                    if matches!(status.code(), tonic::Code::FailedPrecondition) {
```
Phoenix500526, line 457: The `Status::Code` has implemented the `Eq` trait. You can directly use "status.code() == tonic::Code::FailedPrecondition". It looks more concise.

curp/src/client.rs
```diff
@@ -522,6 +530,14 @@ where
                     }
                     resp
                 }
+                Err(RpcError::StatusError(status)) => {
```
Phoenix500526, line 533: Same as above.

curp/src/client.rs
```diff
@@ -813,6 +824,14 @@ where
                 .await
             {
                 Ok(resp) => resp.into_inner(),
+                Err(RpcError::StatusError(status)) => {
+                    if matches!(status.code(), tonic::Code::FailedPrecondition) {
```
Phoenix500526, line 828: Same as above.

curp/src/client.rs
```diff
@@ -864,6 +883,14 @@ where
                 .await
             {
                 Ok(resp) => resp.into_inner(),
+                Err(RpcError::StatusError(status)) => {
```
Phoenix500526, line 886: Same as above

curp/src/server/curp_node.rs
```diff
@@ -65,6 +65,9 @@ pub(super) enum CurpError {
     /// Internal error
     #[error("internal error, {0}")]
     Internal(String),
+    /// Transfer the cluster state
+    #[error("transfer cluster state: {0:?}")]
+    Transfer(Cluster),
```
Phoenix500526, line 70: IMO, it's not a good choice to place such a complex data type in `CurpError` since it goes against the assumptions made in error refactoring - errors should carry information as concisely as possible.   If we really need a complex data structure, like Cluster, we should obtain the relevant information through another RPC after receiving this error.

curp/src/rpc/mod.rs
themanforfree: Fixed

curp/src/rpc/mod.rs
themanforfree: Modified, previously to fix the proto circular dependency issue, but no longer has this issue

curp/src/server/raw_curp/mod.rs
themanforfree: Renamed

curp/src/members.rs
themanforfree: Fixed

curp/src/server/raw_curp/mod.rs
themanforfree: Fixed

curp/src/client.rs
themanforfree: Done

curp/src/client.rs
themanforfree: Same as above.

curp/src/client.rs
themanforfree: Same as above.

curp/src/client.rs
themanforfree: Same as above.

curp/src/members.rs
```diff
@@ -227,6 +258,23 @@ impl ClusterInfo {
...
+
+    /// cluster version increase
+    pub(crate) fn cluster_version_inc(&self) -> u64 {
+        self.cluster_version.fetch_add(1, Ordering::Relaxed)
+    }
+
+    /// cluster version decrease
+    pub(crate) fn cluster_version_dec(&self) -> u64 {
+        self.cluster_version.fetch_sub(1, Ordering::Relaxed)
+    }
```
bsbds, line 276: Will these be used across multiple threads? Maybe these also need stricter ordering.

themanforfree: Fixed

curp/src/server/curp_node.rs
```diff
@@ -840,6 +860,20 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
     pub(super) fn shutdown_listener(&self) -> shutdown::Listener {
         self.curp.shutdown_listener()
     }
+
+    /// Check cluster version and return new cluster
+    fn check_cluster_version(&self, client_cluster_version: u64) -> Result<(), CurpError> {
+        let server_cluster_version = self.curp.cluster().cluster_version();
+        if client_cluster_version < server_cluster_version {
+            let cluster = self.curp.cluster_err();
+            warn!(
```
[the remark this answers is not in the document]
themanforfree: Fixed

curp/src/server/raw_curp/mod.rs
```diff
@@ -337,6 +342,9 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
             Err(e) => return (info, Err(ConfChangeError::new_propose(e.into()))),
         };
         let (addrs, name, is_learner) = self.apply_conf_change(changes);
+        self.ctx
+            .last_conf_change_idx
+            .store(entry.index, Ordering::Relaxed);
```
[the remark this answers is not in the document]

curp/src/server/curp_node.rs
```diff
@@ -528,7 +542,7 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
                     }
                 }
                 _ = remove_event.listen() => {
-                    break false;
+                    is_remove_state = true;
```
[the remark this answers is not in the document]
themanforfree: It is not necessary, we can't remove a node twice, And even if the branch is triggered multiple times, it will not cause any error.

curp/src/server/curp_node.rs
themanforfree: Fixed

curp/src/client.rs
```diff
@@ -635,55 +708,100 @@ where
...
         let cmd_arc = Arc::new(cmd);
+        loop {
+            let res_option = if use_fast_path {
+                self.fast_path(Arc::clone(&cmd_arc)).await
+            } else {
+                self.slow_path(Arc::clone(&cmd_arc)).await
+            };
+            let Some(res) = res_option else {
+                let cluster = self.fetch_cluster(false).await?;
+                self.set_cluster(cluster).await?;
```
bsbds, line 719: I think the `fetch_cluster` here need to make sure that the new cluster version is applied on superquorum of servers, otherwise the fast path will probably still fail after that fetch.

curp/src/client.rs
```diff
@@ -323,12 +347,22 @@ where
...
                     *self.config.wait_synced_timeout(),
                 )
                 .await
             {
                 Ok(resp) => resp.into_inner(),
+                Err(RpcError::StatusError(status)) => {
+                    if is_wrong_cluster_version(&status) {
+                        return Err(CommandProposeError::Propose(
+                            ProposeError::WrongClusterVersion,
+                        ));
```
bsbds, line 360: Why should we return immediately if the cluster version is wrong? I think this early return may cause a long time of unavailability if a server is lag behind.

themanforfree: Why it may cause a long time of unavailability? After returning error here, the outer propose will pull the cluster's configuration and retry the entire propose process

curp/src/client.rs
themanforfree: When the argument passed in fetch cluster is true, it will fetch the latest cluster from leader

curp/src/client.rs
bsbds: I mean even if only one server is lagged behind, this method will return error, but the majority of the cluster may be actually available right now.

themanforfree: The client may be using any cluster version of the configuration, so it cannot tell if most nodes are using the new configuration. And there are no correctness issues with the current implementation, it just has a small probability of slowing down requests. I think the current implementation can be used before there is a safe and fast way

curp/src/server/curp_node.rs
```diff
@@ -847,6 +867,19 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
     pub(super) fn shutdown_listener(&self) -> shutdown::Listener {
         self.curp.shutdown_listener()
     }
+
+    /// Check cluster version and return new cluster
+    fn check_cluster_version(&self, client_cluster_version: u64) -> Result<(), CurpError> {
+        let server_cluster_version = self.curp.cluster().cluster_version();
+        if client_cluster_version != server_cluster_version {
+            debug!(
```
Phoenix500526, line 875: Should it be a warning level?

curp/src/server/curp_node.rs
```diff
@@ -998,4 +1034,22 @@ mod tests {
         assert!(curp.is_leader());
         curp.shutdown_trigger().self_shutdown_and_wait().await;
     }
+
+    fn init_curp_node(
```
Phoenix500526, line 1038: This function is never used. Do we really need it?

curp/src/server/raw_curp/mod.rs
```diff
@@ -792,6 +801,24 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         self.ctx.leader_tx.subscribe()
     }
 
+    /// Check on the leader that the remove node conf change entry has been synced to the removed follower
+    pub(super) fn can_remove_follower_after_hb(&self, follower_id: ServerId) -> bool {
+        if self.ctx.connects.contains_key(&follower_id) {
+            return false;
+        }
+        let match_index = self.lst.get_match_index(follower_id);
+        let last_conf_change_idx = self.ctx.last_conf_change_idx.load(Ordering::Acquire);
```
Phoenix500526, line 810: It seems we can use `Ordering::Relaxed` here.

curp/src/client.rs
```diff
@@ -926,6 +1065,11 @@ fn superquorum(nodes: usize) -> usize {
         .wrapping_add(1)
 }
 
+/// Check whether the status is caused by wrong cluster version
+fn is_wrong_cluster_version(status: &tonic::Status) -> bool {
```
Phoenix500526, line 1069: Do not use the message to tell apart different statues with the same code since the message is unstable. The message may be changed in the future version. If you really need to do so, please add an error label in metadata.

curp/tests/common/curp_group.rs
```diff
@@ -299,6 +299,40 @@ impl CurpGroup {
             .map(|(k, v)| (*k, vec![v.clone()]))
             .collect()
     }
+
+    pub async fn fetch_cluster_info(&self, addrs: &[String]) -> ClusterInfo {
```
Phoenix500526, line 303: These two method seems not use in this pr. Shall we keep them?

curp/src/server/curp_node.rs
themanforfree: FYI: https://github.com/xline-kv/Xline/pull/460#discussion_r1361423702

curp/src/server/curp_node.rs
themanforfree: Removed

curp/tests/common/curp_group.rs
themanforfree: The ignored test use this two function, it will be enable after #448 merged

curp/src/server/raw_curp/mod.rs
themanforfree: FYI: https://github.com/xline-kv/Xline/pull/460#discussion_r1361413078

curp/src/members.rs
Phoenix500526: I don't think so. While atomic operations using relaxed memory ordering do not provide any happens-before relationship, they do guarantee a total modification order of each individual atomic variable. This means that all modifications of the same atomic variable happen in an order that is the same from the perspective of every single thread. Do we really need the guarantee of the happens-before relationship for this cluster_version?

Phoenix500526 (approved):

bsbds (approved):

PR #469, opened 2023-10-08
refactor: move command related data structures

bsbds:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

  Some common data structures used by both `xline` and `xline-client` needed to be move to `xlineapi` to simplify dependency relations.

* what changes does this pull request make?

  This PR moves some data structures from `xline` to `xlineapi` and makes necessary changes such as visibility and module path, it does not modify the original code. 

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #472, opened 2023-10-08
fix: txn command key

bsbds:
Closes: #470 

Based on: #469 


Phoenix500526 (approved):
LGTM

xlineapi/src/command.rs
```diff
@@ -508,6 +509,82 @@ impl PbCodec for Command {
     }
 }
 
+/// Get command keys from a Request for conflict check
+pub trait CommandKeys {
+    /// Key ranges
+    fn keys(&self) -> Vec<KeyRange>;
```
iGxnon, line 515: Is this name conflict with `Command::keys`? `key_ranges` might be a better name.

PR #482, opened 2023-10-12
fix: add linearizable parameter in FetchClusterRequest

iGxnon:

proto changes: https://github.com/xline-kv/curp-proto/pull/14

 Fix linearizable read in FetchClusterRequest

### DIFF
--- a/curp/src/client.rs
+++ b/curp/src/client.rs
@@ -507,11 +507,14 @@ where
     }
 
     /// Send fetch cluster requests to all servers
-    /// Note: The fetched cluster may still be outdated
+    /// Note: The fetched cluster may still be outdated if `linearizable` is false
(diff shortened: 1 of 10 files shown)


curp/src/client.rs
```diff
@@ -542,6 +545,10 @@ where
                         continue;
                     }
                 };
+                if inner.members.is_empty() {
+                    // ignore the empty members response
+                    continue;
+                }
```
bsbds, line 551: The `ok_cnt` is not checked at the end of this method, and it skipped adding because of this continue.

iGxnon: fixed

bsbds (approved):

Phoenix500526 (approved):
LGTM

PR #494, opened 2023-10-24
[Feature]: Support pre vote phase

themanforfree:
Support pre vote phase
Avoid timed out nodes from disrupting a healthy cluster
Add pre vote and add more check when node receive a vote request
no

### DIFF
--- a/curp/src/server/raw_curp/mod.rs
+++ b/curp/src/server/raw_curp/mod.rs
@@ -500,16 +500,11 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
             return Err(st_w.term);
         }
         if term > st_w.term {
-            let timeout = st_w.follower_timeout_ticks;
-            if st_w.leader_id.is_some() && self.ctx.election_tick.load(Ordering::Acquire) < timeout
-            {
-                return Err(st_w.term);
-            }
             self.update_to_term_and_become_follower(&mut st_w, term);
         }
 
         // check self role
-        if st_w.role != Role::Follower {
+        if !matches!(st_w.role, Role::Follower | Role::PreCandidate) {
             return Err(st_w.term);
         }
 
@@ -546,7 +541,7 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         last_log_term: u64,
     ) -> Result<(u64, Vec<PoolEntry<C>>), u64> {
         debug!(
-            "{} received vote: term({}), last_log_index({}), last_log_term({}), id({})",
+            "{} received pre vote: term({}), last_log_index({}), last_log_term({}), id({})",
             self.id(),
             term,
             last_log_index,
@@ -576,7 +571,6 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
 
         // grant the vote
         debug!("{} pre votes for server {}", self.id(), candidate_id);
-        self.reset_election_tick();
         Ok((st_r.term, vec![]))
     }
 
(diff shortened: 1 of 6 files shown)


curp/src/server/raw_curp/mod.rs
```diff
@@ -519,6 +530,51 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+    /// Return `Ok(term, spec_pool)` if the vote is granted
+    /// Return `Err(term)` if the vote is rejected
+    pub(super) fn handle_pre_vote(
+        &self,
+        term: u64,
+        candidate_id: ServerId,
+        last_log_index: LogIndex,
+        last_log_term: u64,
+    ) -> Result<(u64, Vec<PoolEntry<C>>), u64> {
+        debug!(
```
bsbds, line 543: Please also update this debug message.

curp/src/server/raw_curp/mod.rs
```diff
@@ -519,6 +530,51 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
...
+        }
+
+        // check if the candidate's log is up-to-date
+        if !log_r.log_up_to_date(last_log_term, last_log_index) {
+            return Err(st_r.term);
+        }
+
+        // grant the vote
+        debug!("{} pre votes for server {}", self.id(), candidate_id);
+        self.reset_election_tick();
```
bsbds, line 574: If we reset election tick here, will the follower accept the pre-vote but reject the vote?

curp/src/server/raw_curp/mod.rs
```diff
@@ -519,6 +530,51 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
         Ok((st_w.term, self_spec_pool))
     }
 
+    /// Handle `pre_vote`
+    /// Return `Ok(term, spec_pool)` if the vote is granted
+    /// Return `Err(term)` if the vote is rejected
+    pub(super) fn handle_pre_vote(
```
iGxnon, line 536: Could we merge this method into `handle_vote()`? They are very similar.

curp/src/server/raw_curp/mod.rs
```diff
@@ -489,6 +495,11 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
             return Err(st_w.term);
         }
         if term > st_w.term {
+            let timeout = st_w.follower_timeout_ticks;
+            if st_w.leader_id.is_some() && self.ctx.election_tick.load(Ordering::Acquire) < timeout
```
iGxnon, line 499: We already check that the leader is None in `PreVote` phase, as the entire process is `PreVote` -> `Vote`, should we check it again here?

curp/src/server/raw_curp/mod.rs
themanforfree: Updated

curp/src/server/raw_curp/mod.rs
themanforfree: There really shouldn't be a reset here. Removed

curp/src/server/raw_curp/mod.rs
themanforfree: I don't think so. They are somewhat similar, but there are more differences, and it is easier to understand in two methods

curp/src/server/raw_curp/mod.rs
themanforfree: Removed

iGxnon (approved):

bsbds (approved):

Issue #497, opened 2023-11-02
[Bug]: Read State


bsbds:
There are several read-state related bugs found in the jepsen tests.

- [x]  1. Lost cmd ids
  Sometimes the cmd ids are lost during the test, this is caused by the current read state only collect `sp`, which is not sufficient, because the sp only store the current speculative execute cmd, but later conflict command will goes into ucp. The solution is to add ucp cmds to read state response.
- [x] 2. Index barrier not triggers in order
  Currently the `IndexBarrier` uses `last_trigger_index` to track the last execute index. However the after sync in execute concurrently so a command with a larger index may execute before a command with a smaller index. This requires a refactor.
```rust
    /// Trigger all barriers whose index is less than or equal to the given index.
    pub(crate) fn trigger(&self, index: u64) {
        let mut inner_l = self.inner.lock();
        if inner_l.last_trigger_index < index {
            inner_l.last_trigger_index = index;
        }
        let mut split_barriers = inner_l.barriers.split_off(&(index.overflow_add(1)));
        std::mem::swap(&mut inner_l.barriers, &mut split_barriers);
        for (_, barrier) in split_barriers {
            barrier.notify(usize::MAX);
        }
    }
```
- [x] 3. Commit index on leader may newer than uncommited cmds
  The commit index may newer because the after sync in execute concurrently. And a cmd with higher log index may complete before a cmd with lower index. To resolve this problem we should combine wait commit index and wait cmd ids, after both returns could we proceed serializable read on current node.

bsbds:
The fix for the first and third is trivial. I'll provide some details about the refactor for the second issue.

As the after sync stage is execute concurrently, we need to first wait for the lower index to complete, before we trigger the higher index.

I uses a binary heap(min heap), aka priority_queue, to track the index triggered. And we use `next` to store the lowest index that haven't triggered. 
```rust
struct IndexBarrierInner {
    next: u64,
    indices: BinaryHeap<IndexRevision>,
}
```
Each time the after sync triggers a index, we push it into the heap and do the following procedure repeatedly: check the heap top if it's equal to `next`, if they are the same, we pop remove the top element and update `next, other wise we break the loop.

```rust
    pub(crate) fn trigger(&self, index: u64, revision: i64) {
        let mut inner_l = self.inner.lock();
        inner_l.indices.push(IndexRevision::new(index, revision));
        while inner_l
            .indices
            .peek()
            .map_or(false, |i| i.index.eq(&inner_l.next))
        {
[...]

PR #503, opened 2023-11-03
refactor: read state barrier

bsbds:
Fix 1 and 2 in #497

### DIFF
--- a/curp/src/server/cmd_worker/conflict_checked_mpmc.rs
+++ b/curp/src/server/cmd_worker/conflict_checked_mpmc.rs
@@ -483,7 +483,7 @@ impl<C: Command, CE: CommandExecutor<C>> Filter<C, CE> {
                 }
             }
             CEEvent::Reset(snapshot, finish_tx) => {
-                // since a reset is needed, all other vertexes doesn't matter anymore, so delete them all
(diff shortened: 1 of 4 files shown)


iGxnon:
A bit vec queue might be sufficient to solve this gracefully.

Screenshot 2023-11-07 at 14 30 46

Based on your description, I found that the [Tracker](https://github.com/xline-kv/Xline/pull/450/files#diff-d16e0732694070d1f04f3160ccafa6b615b4cfd085837a472c1d3e13fe4ceb34) in #450 may be able to meet this requirement. It has a higher memory compression ratio and better complexity than a binary heap.

bsbds:
> A bit vec queue might be sufficient to solve this gracefully.
> 
> Screenshot 2023-11-07 at 14 30 46
> Based on your description, I found that the [Tracker](https://github.com/xline-kv/Xline/pull/450/files#diff-d16e0732694070d1f04f3160ccafa6b615b4cfd085837a472c1d3e13fe4ceb34) in #450 may be able to meet this requirement. It has a higher memory compression ratio and better complexity than a binary heap.

Yeah I can rebase this on #450

bsbds:
After refactoring this PR is no longer needed

PR #522, opened 2023-11-16
fix: remove wrong message in abort_on_panic macro

themanforfree:

remove wrong message in abort_on_panic macro
remove wrong message in abort_on_panic macro
no

### DIFF
--- a/curp/tests/it/read_state.rs
+++ b/curp/tests/it/read_state.rs
@@ -23,6 +23,7 @@ async fn read_state() {
             TestCommandResult::default(),
         );
     });
+    sleep_millis(10).await;
     let get_client = group.new_client().await;
     let res = get_client
         .fetch_read_state(&TestCommand::new_get(vec![0]))
--- a/test-macros/src/lib.rs
+++ b/test-macros/src/lib.rs
@@ -5,12 +5,11 @@ use syn::{parse_macro_input, ItemFn, Stmt};
 #[proc_macro_attribute]
 pub fn abort_on_panic(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut input: syn::ItemFn = parse_macro_input!(item as ItemFn);
-    let test_case = &input.sig.ident.to_string();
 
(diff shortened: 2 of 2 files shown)


curp/tests/it/read_state.rs
```diff
@@ -23,6 +23,7 @@ async fn read_state() {
             TestCommandResult::default(),
         );
     });
+    sleep_millis(10).await;
```
Phoenix500526, line 26: Why do you sleep here?

test-macros/src/lib.rs
```diff
@@ -5,12 +5,11 @@ use syn::{parse_macro_input, ItemFn, Stmt};
 #[proc_macro_attribute]
 pub fn abort_on_panic(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut input: syn::ItemFn = parse_macro_input!(item as ItemFn);
-    let test_case = &input.sig.ident.to_string();
 
     let panic_hook: Stmt = syn::parse_quote! {
         std::panic::set_hook(Box::new(|info| {
             let stacktrace = std::backtrace::Backtrace::force_capture();
-            println!("`{}` panic! \n@info:\n{}\n@stackTrace:\n{}", #test_case, info, stacktrace);
+            println!("test panic! \n@info:\n{}\n@stackTrace:\n{}", info, stacktrace);
```
Phoenix500526, line 12: Can we parse the test case name from the first stack frame?

curp/tests/it/read_state.rs
themanforfree: In order to ensure that the `put_cmd` is proposed before `fetch_read_state`

test-macros/src/lib.rs
themanforfree: the test case name is irregularity in backtrace, we can't parse the test names unless we force all test names to conform to the specified format

bsbds (approved):

Phoenix500526 (approved):

PR #528, opened 2023-11-17
fix: do not increase cluster_version when new node apply old conf changes

themanforfree:

fix xline_add_node test
do not increase cluster_version when new node apply old conf changes
no


Phoenix500526 (approved):
LGTM

Issue #526, opened 2023-11-17
[Bug]: May not be able to Shutdown after Member Add


themanforfree:
### Description about the bug

If we  shutdown the cluster immediately after Member Add, then the Leader node may not be able to shutdown itself properly. leader will consistently attempt to send append entries to the new node that has been shutdown

### Version

0.4.1 (Default)

### Relevant log output

```shell
...
   0.689296339s DEBUG curp::server::curp_node: log persist task exits
   0.689491384s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.699618528s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.700189292s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.716528023s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.716847751s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.732129655s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.732453603s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.748612679s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.748923826s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.765241846s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.765549214s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.781724370s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
   0.782032928s  WARN curp::server::curp_node: ae to 4860512024351935985 failed, transport error, error trying to connect: tcp connect error: Connection refused (os error 111)
   0.798256406s DEBUG curp::server::curp_node: 6805624495683170404 send append_entries to 4860512024351935985
[...]

PR #535, opened 2023-11-21
fix: fix cluster shutdown after member add

themanforfree:
#526 
fix cluster shutdown after member add
Add a new phase for shutdown
no

### DIFF
--- a/curp/src/server/curp_node.rs
+++ b/curp/src/server/curp_node.rs
@@ -31,7 +31,7 @@ use super::{
 use crate::{
     cmd::{Command, CommandExecutor},
     error::ERROR_LABEL,
-    log_entry::LogEntry,
+    log_entry::{EntryData, LogEntry},
     members::{ClusterInfo, ServerId},
     role_change::RoleChange,
     rpc::{
@@ -45,7 +45,8 @@ use crate::{
     },
     server::{cmd_worker::CEEventTxApi, raw_curp::SyncAction, storage::db::DB},
     snapshot::{Snapshot, SnapshotMeta},
-    ConfChangeType, PublishRequest, PublishResponse,
+    ConfChangeType, PublishRequest, PublishResponse, TriggerShutdownRequest,
+    TriggerShutdownResponse,
 };
 
 /// Curp error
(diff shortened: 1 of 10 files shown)


utils/src/shutdown.rs
```diff
@@ -71,6 +73,13 @@ impl Trigger {
             .store(true, Ordering::Relaxed);
     }
 
+    /// Mark leader notified
+    #[inline]
+    pub fn mark_leader_notified(&self) {
+        debug!("mark leader notified");
```
Phoenix500526, line 79: It seems we should use `info!` rather than `debug!` here since the shutdown message is important.

utils/src/shutdown.rs
```diff
@@ -34,6 +34,8 @@ pub struct TriggerInner {
     mpmc_channel_shutdown: AtomicBool,
     /// State of sync follower daemon.
     sync_follower_daemon_shutdown: AtomicBool,
+    /// Shutdown Applied
+    leader_notified: AtomicBool,
```
Phoenix500526, line 38: The `TriggerInner` doesn't need to be `pub`.

curp/src/server/curp_node.rs
```diff
@@ -702,28 +714,43 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
...
                             }
                         } else {
                             hb_opt = true;
                         }
-                        if is_shutdown_state && is_empty && curp.is_synced() {
+                        // Check Follower shutdown
+                        if is_shutdown_state
+                            && ((curp.is_synced(connect_id) && is_empty)
+                                || (!curp.is_synced(connect_id) && is_commit_shutdown))
+                        {
```
bsbds, line 746: Please add more detailed comment here to describe the shutdown condition

curp/src/server/curp_node.rs
```diff
@@ -702,28 +714,43 @@ impl<C: 'static + Command, RC: RoleChange + 'static> CurpNode<C, RC> {
...
+                            let _ig = tokio::spawn(async move {
```
bsbds, line 747: This is a async task, could the leader has already shutdowned before these task still not get started?

curp/src/server/raw_curp/mod.rs
```diff
@@ -1030,10 +1030,12 @@ impl<C: 'static + Command, RC: RoleChange + 'static> RawCurp<C, RC> {
     }
 
     /// Check if all followers have caught up with the leader
-    pub(super) fn is_synced(&self) -> bool {
+    pub(super) fn is_synced(&self, node_id: ServerId) -> bool {
```
bsbds, line 1033: Please also update the comment of this function

utils/src/shutdown.rs
themanforfree: Updated

utils/src/shutdown.rs
themanforfree: Removed

curp/src/server/curp_node.rs
themanforfree: Added

curp/src/server/raw_curp/mod.rs
themanforfree: Updated

Phoenix500526 (approved):

PR #538, opened 2023-11-22
build: modify the dockerfile and bootup logic

Phoenix500526:
Add a startup_xline.sh script to boot up xline service in docker. Therefore, we can generate the k8s pod template more smoothly.

### DIFF
--- a/doc/quick-start/Dockerfile
+++ b/doc/quick-start/Dockerfile
@@ -5,6 +5,7 @@ WORKDIR /build
 COPY . .
 
 RUN set -eux && \
+  sed -i s@/deb.debian.org/@/mirrors.aliyun.com/@g /etc/apt/sources.list && \
   apt-get update && apt-get install -y git build-essential autoconf autogen libtool clang && \
   git clone --branch v3.21.12 --recurse-submodules https://github.com/protocolbuffers/protobuf && \
   cd protobuf && ./autogen.sh && ./configure && make -j4 && make install && cd .. && ldconfig && \
@@ -19,6 +20,7 @@ RUN set -eux && \
 
 COPY --from=builder /build/target/release/xline /usr/local/bin
 COPY --from=builder /build/target/release/benchmark /usr/local/bin
-COPY --from=builder /build/target/release/lock_client /usr/local/bin
+COPY --from=builder /build/target/release/validation_lock_client /usr/local/bin
+COPY --from=builder /build/scripts/start_xline.sh /usr/local/bin
 
-CMD ["/usr/local/bin/xline"]
+CMD ["/bin/bash", "/usr/local/bin/start_xline.sh"]
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -2,7 +2,9 @@ FROM ubuntu:latest
 
 COPY xline /usr/local/bin
 COPY benchmark /usr/local/bin
+COPY start_xline.sh /usr/local/bin
 
 RUN apt-get update && apt-get install -y iproute2 iputils-ping
 
-CMD ["/usr/local/bin/xline"]
+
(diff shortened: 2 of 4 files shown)


doc/quick-start/Dockerfile
```diff
@@ -5,10 +5,12 @@ WORKDIR /build
 COPY . .
 
 RUN set -eux && \
-  apt-get update && apt-get install -y git build-essential autoconf autogen libtool clang && \
+  sed -i s@/deb.debian.org/@/mirrors.aliyun.com/@g /etc/apt/sources.list && \
```
iGxnon, line 8: I guess it works slower on Github Action runner machines.

doc/quick-start/Dockerfile
```diff
@@ -5,10 +5,12 @@ WORKDIR /build
...
+  apt-get clean && apt-get update && apt-get install -y git build-essential autoconf autogen libtool clang && \
```
iGxnon, line 9: Then we could remove `apt-get clean`

doc/quick-start/Dockerfile
```diff
@@ -5,10 +5,12 @@ WORKDIR /build
...
   git clone --branch v3.21.12 --recurse-submodules https://github.com/protocolbuffers/protobuf && \
   cd protobuf && ./autogen.sh && ./configure && make -j4 && make install && cd .. && ldconfig && \
-  cargo build --release
+  cargo build --release && \
+  cargo build --release --bin validation_lock_client --target x86_64-unknown-linux-gnu
```
iGxnon, line 13: The above command `cargo build --release` should build `validation_lock_client`

scripts/startup_xline.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
iGxnon, line 1: The filename `start_xline.sh` might be better.

scripts/startup_xline.sh
```diff
@@ -0,0 +1,36 @@
...
+
```
iGxnon, line 2: BTW, I think we can read environment variables to build the command for running xline, instead of using a script. That's because the script does not appear to have introduced the configuration for optional arguments such as `--jaeger-online`. After including the optional arguments, any changes to the launch parameters of xline in the future need to be synchronized to this script.
On k8s platform, FYI: https://stackoverflow.com/questions/56687542/how-to-use-environment-variable-in-kubernetes-container-command.
How do you think about it?

scripts/Dockerfile
```diff
@@ -2,7 +2,9 @@ FROM ubuntu:latest
```
iGxnon, line 2: This dockerfile is also be used in releasing images, shall we put in other files?

doc/quick-start/Dockerfile
Phoenix500526: Removed

doc/quick-start/Dockerfile
Phoenix500526: Removed

scripts/startup_xline.sh
Phoenix500526: Done

doc/quick-start/Dockerfile
Phoenix500526: But it works faster for domestic users who need to build the Xline docker image from scratch.

scripts/Dockerfile
Phoenix500526: I don't get the point. What does "other files" mean here?

scripts/startup_xline.sh
Phoenix500526: Make sense. I've left an issue #551  to track this.

scripts/Dockerfile
iGxnon: I got it mixed up, resolved.

PR #555, opened 2023-11-30
CI/build env

iGxnon:

    Add build environment image building scripts

   As above

    No


ci/Dockerfile
```diff
@@ -0,0 +1,67 @@
+FROM ubuntu:22.04 AS build-env
+
+ARG LIB_ROCKS_SYS_VER
+
+RUN apt-get update -yy && \
```
Phoenix500526, line 5: Remove the duplicated 'y'

iGxnon: `-yy` seems to be the abbreviation of `--yes --force-yes`, which the later one could omit any configuration.

Phoenix500526 (approved):
LGTM

bsbds (approved):

PR #557, opened 2023-12-01
[Test]: madsim test for xline

themanforfree:

Add basic madsim test framework for xline


benchmark/src/bench_client.rs
```diff
@@ -127,7 +111,7 @@ impl BenchClient {
     ///
     /// If `XlineClient` or `EtcdClient` failed to send request
     #[inline]
-    #[cfg(test)]
+    #[allow(unused)]
     pub(crate) async fn get(
```
Phoenix500526, line 115: This method only be used in tests. I think `cfg(test)` is better than `allow(unused)`

simulation/tests/it/xline.rs
```diff
@@ -0,0 +1,12 @@
+use curp_test_utils::init_logger;
+use simulation::xline_group::XlineGroup;
+use xline_client::types::kv::PutRequest;
+
+#[madsim::test]
+async fn basic_put() {
```
Phoenix500526, line 6: Can we add more test cases for xline？

themanforfree: This PR just add basic madsim test framework for xline, we can add more tests for this if needed

Phoenix500526: Please add a todo comment here.

Phoenix500526 (dismissed):
LGTM

benchmark/src/bench_client.rs
```diff
@@ -139,12 +124,82 @@ impl BenchClient {
...
             KVClient::Etcd(ref mut etcd_client) => {
                 let response = etcd_client.get(request.key(), None).await?;
-                Ok(response.into())
+                Ok(convert::get_res(response))
             }
         }
     }
 }
 
+/// Convert utils
```
iGxnon, line 133: The `From` trait might be more suitable for converting semantics.

simulation/src/xline_group.rs
```diff
@@ -0,0 +1,129 @@
...
+                let id = cluster_info.self_id();
+
+                let handle = handle
+                    .create_node()
+                    .name(id.to_string())
+                    .ip(format!("192.168.1.{}", i + 1).parse().unwrap())
+                    .init(move || {
+                        let server = XlineServer::new(
+                            cluster_info.clone(),
+                            i == 0,
```
iGxnon, line 44: we might restart the old leader.
after the reboot, it may no longer be the leader.

xline-client/src/lib.rs
```diff
@@ -417,9 +417,18 @@ struct AuthService<S> {
 impl<S> AuthService<S> {
     /// Create a new `AuthService`
     #[inline]
+    #[cfg(not(madsim))]
     fn new(inner: S, token: Option<Arc<HeaderValue>>) -> Self {
         Self { inner, token }
     }
+
+    /// Create a new `AuthService`
```
iGxnon, line 425: /// bypassed

benchmark/src/bench_client.rs
themanforfree: Orphan rules don't allow this, both types come from external dependencies

xline-client/src/lib.rs
themanforfree: I think it's better to be consistent with comment in a non-madsim environment

Phoenix500526 (approved):

PR #558, opened 2023-12-03
Ci/generate rsa key pair

Phoenix500526:
generate a key pair in ci process


.github/workflows/validation.yml
```diff
@@ -31,7 +31,8 @@ jobs:
       - run: |
           cd scripts
           cp ../target/release/{xline,benchmark,validation_lock_client} .
-          cp ../xline-test-utils/{private,public}.pem .
+          openssl genrsa -out ./private.pem
+          openssl rsa -in ./private.pem -pubout -outform PEM -out ./public.pem
```
bsbds, line 35: Maybe using a morden ECC algorithm like X25519 is better.

doc/quick-start/README.md
```diff
@@ -88,7 +88,9 @@ docker build . -t ghcr.io/xline-kv/xline -f doc/quick-start/Dockerfile
 ### Start Xline servers
 
 ```bash
-cp ./xline-test-utils/{private,public}.pem ./scripts
+# Generate RSA key pair for JSON Web Token signature
+openssl genrsa -out ./private.pem
+openssl rsa -in ./private.pem -pubout -outform PEM -out ./public.pem
```
bsbds, line 93: Same as above.

xline-test-utils/src/lib.rs
```diff
@@ -6,6 +6,7 @@ use std::{
 
 use curp::members::{get_cluster_info_from_remote, ClusterInfo};
 use jsonwebtoken::{DecodingKey, EncodingKey};
+use openssl::rsa::Rsa;
```
bsbds, line 9: Same as above

xline/src/storage/auth_store/store.rs
```diff
@@ -1140,6 +1140,7 @@ mod test {
     use std::collections::HashMap;
 
     use merged_range::MergedRange;
+    use openssl::rsa::Rsa;
```
bsbds, line 1143: Same as above

.github/workflows/validation.yml
Phoenix500526: The JWT token is transmitted in plaintext, and the key here is only used for signing to prevent the token from being tampered with.

doc/quick-start/README.md
Phoenix500526: Same as above

xline-test-utils/src/lib.rs
Phoenix500526: Same as above

xline/src/storage/auth_store/store.rs
Phoenix500526: Same as above

themanforfree (approved):
LGTM

PR #563, opened 2023-12-05
feat: CURP WAL codec

bsbds:
Based on: #562

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

  There will be 6 PRs of the CURP WAL implementation, this is the 2nd part.

* what changes does this pull request make?

  This PR implements the WAL codec

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #571, opened 2023-12-11
feat: CURP WAL Pipeline

bsbds:
Based on : #563

  Implement CURP WAL

  This PR implements the WAL pipeline part, the file pipeline is used for pre-allocating files for WAL segment.

### DIFF
--- a/crates/curp/src/server/storage/wal/pipeline.rs
+++ b/crates/curp/src/server/storage/wal/pipeline.rs
@@ -46,7 +46,8 @@ impl FilePipeline {
         let handle = tokio::spawn(async move {
             let mut file_count = 0;
             loop {
-                let file = Self::alloc(&dir_c, file_size, &mut file_count)?;
+                let file = Self::alloc(&dir_c, file_size, file_count)?;
+                file_count += 1;
                 tokio::select! {
                     _ = &mut stop_listener => {
                         break;
@@ -78,15 +79,10 @@ impl FilePipeline {
     }
 
     /// Allocates a a new tempfile
-    fn alloc(
-        dir: impl AsRef<Path>,
-        file_size: u64,
-        file_count: &mut usize,
-    ) -> io::Result<LockedFile> {
-        let fpath = PathBuf::from(dir.as_ref()).join(format!("{}{TEMP_FILE_EXT}", *file_count));
+    fn alloc(dir: impl AsRef<Path>, file_size: u64, file_count: usize) -> io::Result<LockedFile> {
+        let fpath = PathBuf::from(dir.as_ref()).join(format!("{file_count}{TEMP_FILE_EXT}"));
         let mut locked_file = LockedFile::open_rw(fpath)?;
         locked_file.preallocate(file_size)?;
-        *file_count = file_count.overflow_add(1);
         Ok(locked_file)
     }
 
--- a/crates/curp/src/server/storage/wal/pipeline.rs
+++ b/crates/curp/src/server/storage/wal/pipeline.rs
@@ -1,6 +1,10 @@
 use std::{
     io,
     path::{Path, PathBuf},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
     task::Poll,
 };
 
@@ -26,10 +30,8 @@ pub(super) struct FilePipeline {
     file_size: u64,
     /// The file receive stream
     file_stream: RecvStream<'static, LockedFile>,
-    /// The handle of the background file creation task
-    handle: JoinHandle<io::Result<()>>,
     /// Stopped flag
-    stopped: bool,
+    stopped: Arc<AtomicBool>,
 }
 
 impl FilePipeline {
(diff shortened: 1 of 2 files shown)


curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        let (file_tx, file_rx) = flume::bounded(1);
+        let stop_event = Event::new();
+        let mut stop_listener = stop_event.listen();
+        let dir_c = dir.clone();
+
+        let handle = tokio::spawn(async move {
+            let mut file_count = 0;
+            loop {
+                let file = Self::alloc(&dir_c, file_size, &mut file_count)?;
```
themanforfree, line 49: I think it's better to pass `usize` directly here and then increment `file_count` inside the loop

curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,169 @@
...
+                        // The receiver is already dropped, stop this task
+                        if let Err(e) = result {
+                            break;
+                        }
+                    }
+                }
+            }
+            Self::clean_up(&dir_c)?;
+            Ok(())
+        });
```
themanforfree, line 64: this task is cancel safe, we can remove `stop_listener` and direct abort it

curp/src/server/storage/wal/pipeline.rs
bsbds: done

curp/src/server/storage/wal/pipeline.rs
bsbds: done

curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,170 @@
...
+                    }
+                    Ok(Ok(_)) => return Poll::Ready(None),
+                }
+            }
+            return Poll::Ready(None);
+        }
+
+        if self.stopped {
+            return Poll::Ready(None);
+        }
```
Phoenix500526, line 128: We can merge this piece of code. The `drop` method will invoke the `stop` method, aborting the handle and setting the stop flag to true. Therefore, the stop flag must be true when the `is_finished` returns true. If the `is_finished` returns false, the stop flag is still true. We can bank on this stop flag to cancel the stream.

curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,167 @@
...
+            dir,
+            file_size,
+            file_stream: file_rx.into_stream(),
+            stopped,
+        })
+    }
+
+    /// Stops the pipeline
+    pub(super) fn stop(&mut self) {
+        self.stopped.store(true, Ordering::SeqCst);
```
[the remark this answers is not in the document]
bsbds: It seems `Ordering::Relaxed` is sufficient in this case, fixed.

Phoenix500526:
Please add a signed-off message in these commits

curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,167 @@
...
+        let (file_tx, file_rx) = flume::bounded(1);
+        let stop_event = Event::new();
+        let dir_c = dir.clone();
+        let stopped = Arc::new(AtomicBool::new(false));
+        let stopped_c = Arc::clone(&stopped);
+
+        let _ignore = tokio::spawn(async move {
+            let mut file_count = 0;
+            loop {
+                let file = match Self::alloc(&dir_c, file_size, file_count) {
```
iGxnon, line 52: `Self::alloc` seems to be slow (preallocation) and it is blocking (i.e. no `.await`). Could we convert it into async io or spawn this task in blocking threads pool(`tokio::spawn_blocking`)?

iGxnon: This may block the Tokio scheduler every time the log is sealed and generate the next log file.

bsbds: Tokio file does not seems to support preallocation, I refactored this to run in another thread.

crates/curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,183 @@
...
+use tokio_stream::Stream;
+use tracing::error;
+
+use super::util::LockedFile;
+
+/// The temp file extension
+const TEMP_FILE_EXT: &str = ".tmp";
+
+/// The file pipeline, used for pipelining the creation of temp file
+pub(super) struct FilePipeline {
```
Phoenix500526, line 26: Can a user create more than one `FilePipeline`? If so, what will happen?

crates/curp/src/server/storage/wal/pipeline.rs
```diff
@@ -0,0 +1,183 @@
...
+    stopped: Arc<AtomicBool>,
+}
+
+impl FilePipeline {
+    /// Creates a new `FilePipeline`
+    pub(super) fn new(dir: PathBuf, file_size: u64) -> io::Result<Self> {
+        Self::clean_up(&dir)?;
+
+        let (file_tx, file_rx) = flume::bounded(1);
+        let stop_event = Event::new();
```
Phoenix500526, line 43: What's this `stop_event` for?

crates/curp/src/server/storage/wal/pipeline.rs
bsbds: The user don't create a pipeline, it only act as a WAL componet and will only be created once.

crates/curp/src/server/storage/wal/pipeline.rs
bsbds: it serve no use, removed

Phoenix500526 (approved):
LGTM

iGxnon (approved):

PR #581, opened 2023-12-20
Refactor/curp client

iGxnon:

    Introduce the new client architecture.

    As above

    No

### DIFF
--- a/curp/src/client_new/mod.rs
+++ b/curp/src/client_new/mod.rs
@@ -1,3 +1,6 @@
+//! CURP client
+//! The Curp client should be used to access the Curp service cluster, instead of using direct RPC.
+
 /// Unary rpc client
 mod unary;
 
@@ -24,9 +27,12 @@ type ProposeResponse<C: Command> = Result<(C::ER, Option<C::ASR>), C::Error>;
 /// `ClientApi`, a higher wrapper for `ConnectApi`, providing some methods for communicating to
 /// the whole curp cluster. Automatically discovery curp server to update it's quorum.
 #[async_trait]
-pub trait ClientApi<C: Command> {
+pub trait ClientApi {
     /// The client error
     type Error;
+
+    /// The command type
+    type Cmd: Command;
 
     /// Get the local connection when the client is on the server node.
     fn local_connect(&self) -> Option<Arc<dyn ConnectApi>>;
@@ -35,9 +41,9 @@ pub trait ClientApi<C: Command> {
     /// requests (event the requests are commutative).
     async fn propose(
         &self,
-        cmd: &C,
+        cmd: &Self::Cmd,
         use_fast_path: bool,
-    ) -> Result<ProposeResponse<C>, Self::Error>;
+    ) -> Result<ProposeResponse<Self::Cmd>, Self::Error>;
 
     /// Send propose configuration changes to the cluster
     async fn propose_conf_change(
@@ -49,7 +55,7 @@ pub trait ClientApi<C: Command> {
     async fn propose_shutdown(&self) -> Result<(), Self::Error>;
 
     /// Send fetch read state from leader
-    async fn fetch_read_state(&self, cmd: &C) -> Result<ReadState, Self::Error>;
+    async fn fetch_read_state(&self, cmd: &Self::Cmd) -> Result<ReadState, Self::Error>;
 
     /// Send fetch cluster requests to all servers (That's because initially, we didn't
     /// know who the leader is.)
--- a/curp/src/client_new/mod.rs
+++ b/curp/src/client_new/mod.rs
@@ -1,0 +1,168 @@
+/// Unary rpc client
+mod unary;
+
+use std::{collections::HashMap, sync::Arc};
+
+use async_trait::async_trait;
+use curp_external_api::cmd::Command;
+use futures::{stream::FuturesUnordered, StreamExt};
+use utils::config::ClientConfig;
+
+use crate::{
+    members::ServerId,
+    rpc::{
+        connect::ConnectApi, protocol_client::ProtocolClient, ConfChange, FetchClusterRequest,
+        FetchClusterResponse, Member, ReadState,
+    },
+};
+
+/// The response of propose command, deserialized from [`crate::rpc::ProposeResponse`] or
+/// [`crate::rpc::WaitSyncedResponse`].
+#[allow(type_alias_bounds)] // that's not bad
+type ProposeResponse<C: Command> = Result<(C::ER, Option<C::ASR>), C::Error>;
+
+/// `ClientApi`, a higher wrapper for `ConnectApi`, providing some methods for communicating to
+/// the whole curp cluster. Automatically discovery curp server to update it's quorum.
+#[async_trait]
+pub trait ClientApi<C: Command> {
+    /// The client error
+    type Error;
+
+    /// Get the local connection when the client is on the server node.
+    fn local_connect(&self) -> Option<Arc<dyn ConnectApi>>;
+
+    /// Send propose to the whole cluster, `use_fast_path` set to `false` to fallback into ordered
+    /// requests (event the requests are commutative).
+    async fn propose(
+        &self,
+        cmd: &C,
+        use_fast_path: bool,
+    ) -> Result<ProposeResponse<C>, Self::Error>;
+
+    /// Send propose configuration changes to the cluster
+    async fn propose_conf_change(
+        &self,
+        changes: Vec<ConfChange>,
+    ) -> Result<Vec<Member>, Self::Error>;
+
+    /// Send propose to shutdown cluster
+    async fn propose_shutdown(&self) -> Result<(), Self::Error>;
+
+    /// Send fetch read state from leader
+    async fn fetch_read_state(&self, cmd: &C) -> Result<ReadState, Self::Error>;
+
+    /// Send fetch cluster requests to all servers (That's because initially, we didn't
+    /// know who the leader is.)
+    /// Note: The fetched cluster may still be outdated if `linearizable` is false
+    async fn fetch_cluster(&self, linearizable: bool) -> Result<FetchClusterResponse, Self::Error>;
+
+    /// Fetch leader id
+    #[inline]
+    async fn fetch_leader_id(&self, linearizable: bool) -> Result<ServerId, Self::Error> {
+        if linearizable {
+            let resp = self.fetch_cluster(true).await?;
+            return Ok(resp.leader_id.unwrap_or_else(|| {
+                unreachable!("linearizable fetch cluster should return a leader id")
+            }));
+        }
+        let resp = self.fetch_cluster(false).await?;
+        if let Some(id) = resp.leader_id {
+            return Ok(id);
+        }
+        // fallback to linearizable fetch
+        self.fetch_leader_id(true).await
+    }
+}
+
+/// Client builder to build a client
+#[derive(Debug, Clone)]
+pub struct ClientBuilder {
+    /// local server id
+    local_server_id: Option<ServerId>,
+    /// initial cluster version
+    cluster_version: Option<u64>,
+    /// client configuration
+    config: ClientConfig,
+    /// initial all members
+    all_members: Option<HashMap<ServerId, Vec<String>>>,
+}
+
+impl ClientBuilder {
+    /// Create a client builder
+    #[inline]
+    #[must_use]
+    pub fn new(config: ClientConfig) -> Self {
+        Self {
+            local_server_id: None,
+            config,
+            all_members: None,
+            cluster_version: None,
+        }
+    }
+
+    /// Set the local server id
+    #[inline]
+    #[must_use]
+    pub fn local_server_id(&mut self, id: ServerId) -> &mut Self {
+        self.local_server_id = Some(id);
+        self
+    }
+
+    /// Set the initial cluster version
+    #[inline]
+    #[must_use]
+    pub fn cluster_version(&mut self, cluster_version: u64) -> &mut Self {
+        self.cluster_version = Some(cluster_version);
+        self
+    }
+
+    /// Fetch initial all members from some endpoints if you do not know the whole members
+    ///
+    /// # Errors
+    ///
+    /// Return `tonic::Status` for connection failure or some server errors.
+    #[inline]
+    pub async fn fetch_all_members(
+        &mut self,
+        addrs: Vec<String>,
+    ) -> Result<&mut Self, tonic::Status> {
+        let propose_timeout = *self.config.propose_timeout();
+        let mut futs: FuturesUnordered<_> = addrs
+            .into_iter()
+            .map(|mut addr| {
+                if !addr.starts_with("http://") {
+                    addr.insert_str(0, "http://");
+                }
+                async move {
+                    let mut protocol_client = ProtocolClient::connect(addr).await.map_err(|e| {
+                        tonic::Status::cancelled(format!("cannot connect to addr, error: {e}"))
+                    })?;
+                    let mut req = tonic::Request::new(FetchClusterRequest::default());
+                    req.set_timeout(propose_timeout);
+                    let fetch_cluster_res = protocol_client.fetch_cluster(req).await?.into_inner();
+                    Ok::<FetchClusterResponse, tonic::Status>(fetch_cluster_res)
+                }
+            })
+            .collect();
+        let mut err = tonic::Status::invalid_argument("addrs is empty");
+        while let Some(r) = futs.next().await {
+            match r {
+                Ok(r) => {
+                    self.cluster_version = Some(r.cluster_version);
+                    self.all_members = Some(r.into_members_addrs());
+                    return Ok(self);
+                }
+                Err(e) => err = e,
+            }
+        }
+        Err(err)
+    }
+
+    /// Set the initial all members
+    #[inline]
+    #[must_use]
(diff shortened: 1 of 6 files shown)


curp/src/client_new/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+                Err(e) => err = e,
+            }
+        }
+        Err(err)
+    }
+
+    /// Set the initial all members
+    #[inline]
+    #[must_use]
+    pub fn set_all_members(&mut self, all_members: HashMap<ServerId, Vec<String>>) -> &mut Self {
```
Phoenix500526, line 183: Please modify the method name to `all_members` to unify the method naming style with other methods, like `local_server_id`.

curp/src/client_new/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+            return Ok(id);
+        }
+        // fallback to linearizable fetch
+        self.fetch_leader_id(true).await
+    }
+}
+
+/// Client builder to build a client
+#[derive(Debug, Clone)]
+pub struct ClientBuilder {
```
Phoenix500526, line 98: We can use `derive_builder` crate to simply this `ClientBuilder`.

curp/src/rpc/connect.rs
```diff
@@ -468,6 +467,119 @@ impl InnerConnectApi for Connect<InnerProtocolClient<Channel>> {
     }
 }
 
+/// A connect api implementation which bypass kernel to dispatch method directly.
```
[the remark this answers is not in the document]
iGxnon: It uses direct function calls instead of going through the TCP/IP protocol stack (check the loopback address in the linux kernel), which called `bypass kernel`

curp/src/client_new/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Send propose to shutdown cluster
+    async fn propose_shutdown(&self) -> Result<(), Self::Error>;
+
+    /// Send fetch read state from leader
+    async fn fetch_read_state(&self, cmd: &C) -> Result<ReadState, Self::Error>;
+
+    /// Send fetch cluster requests to all servers (That's because initially, we didn't
+    /// know who the leader is.)
+    /// Note: The fetched cluster may still be outdated if `linearizable` is false
+    async fn fetch_cluster(&self, linearizable: bool) -> Result<FetchClusterResponse, Self::Error>;
```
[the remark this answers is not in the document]
iGxnon: Etcd just support that, which them called `serializable read`. And there is a field `linearizable` in `FetchClusterResponse` to enable quick cluster status retrieval (may return the stale data).

curp/src/client_new/mod.rs
iGxnon: It was refactored in #582

curp/src/client_new/mod.rs
iGxnon: It seems not possible. we don't have a `struct Client` to put #[derive(derive_builder::Builder)] on it. We only have the trait `ClientApi`.
Also, there are some async functions that `ClientBuilder` may use which we cannot directly use the function derived from `derive_builder`.

This is how we used the `derive_builder`:

https://github.com/xline-kv/Xline/blob/47bcd9faae6c3d5157657e50783feaddd3b25e83/curp/src/client.rs#L34-L54

I don't think it is a good practice because of many `#[builder(setter(skip))]` on fields. 

BTW, the codes in `utils/src/config.rs` seems more necessary to use `derive_builder` for optimization.

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+        f: impl FnMut(Arc<dyn ConnectApi>) -> F,
+    ) -> (usize, impl Stream<Item = R>) {
+        let connects = self
+            .cluster_state
+            .connects
+            .iter()
+            .map(|connect| Arc::clone(&connect))
+            .collect_vec();
+        // quorum calculated here to keep quorum = stream.len(), otherwise Non-atomic read operation on the `connects` may result in inconsistency.
+        let quorum = connects.len();
```
bsbds, line 228: Why this is called `quorum`, shouldn't it be called something like `cluster_size`?

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+            connects.into_iter().map(f).collect::<FuturesUnordered<F>>(),
+        )
+    }
+
+    /// Get a handle `f` and return the future to apply `f` on the leader.
+    /// NOTICE:
+    /// The leader might be outdate if the local `leader_state` is stale.
+    /// `for_leader` should never be invoked in [`ClientApi::fetch_cluster`]
+    /// `for_leader` might call `fetch_leader_id`, `fetch_leader_id` might call
+    /// `fetch_cluster`, finally result in stack over flow.
```
bsbds, line 240: There's a duplication of "might call".
"over flow" should be overflow

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+#[async_trait]
+impl<C: Command> ClientApi for Unary<C> {
+    /// The error is generated from server
+    type Error = CurpError;
+
+    /// The command type
+    type Cmd = C;
+
+    /// Get the local connection when the client is on the server node.
+    fn local_connect(&self) -> Option<Arc<dyn ConnectApi>> {
```
bsbds, line 387: Do we still needs this `local_connect` if we have the bypassed connect?

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+        // then fetch the whole cluster
+        let (quorum, mut responses) = self.for_each_server(|conn| async move {
+            (
+                conn.id(),
+                conn.fetch_cluster(FetchClusterRequest { linearizable }, timeout)
+                    .await
+                    .map(Response::into_inner),
+            )
+        });
+        let majority_quorum = majority_quorum(quorum);
```
bsbds, line 534: Same as above. Majority and quorum means the same thing I think.

curp/src/client_new/unary.rs
iGxnon: This method find the bypassed connect (if exist) in `connects`. If found, we could use it without taking care of network issues if user enable serializable options.

iGxnon: Currently, it is used within the fetch cluster. It will attempt to fetch from the local server first, and then fetch from the cluster.

curp/src/client_new/unary.rs
iGxnon: I feel like they are the same meaning.

`super quorum` : 3/4 * cluster size -> `super`: 3/4, `quorum`: cluster size

:D

curp/src/client_new/unary.rs
iGxnon: Same as above

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
Phoenix500526: In etcd, there is no FetchClusterRequest; only MemberListRequest exists. The MemberListResponse contains only Members, while FetchClusterResponse additionally includes leader_id and term. For the caller, once it obtains outdated leader_id and term, it will finally do another rpc call, which may lead to lower performance.

curp/src/client_new/unary.rs
Phoenix500526: The reason why we call it super quorum is the super quorum (3/4 * cluster_size) is greater than  the quorum (1/2 * (cluster_size + 1)). It should be cluster_size, not quorum here.

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+
+impl ClusterState {
+    /// Check the cluster version and update the connects if needed.
+    async fn check_and_update(
+        &self,
+        cluster_version: u64,
+        member_addrs: HashMap<ServerId, Vec<String>>,
+    ) -> Result<(), tonic::transport::Error> {
+        // Check and update the cluster version in client side
+        // FIXME: is these ordering options correct?
```
Phoenix500526, line 108: It depends on how we use this version.  If what you need is an atomic variable, I think the Relaxed Ordering is enough.

curp/src/client_new/unary.rs
Phoenix500526: I think they mean the same thing too.

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,620 @@
...
+        if let Some(res) = res {
+            debug!("fetch cluster succeeded, result: {res:?}");
+            if let Err(e) = self
+                .cluster_state
+                .check_and_update(res.cluster_version, res.clone().into_members_addrs())
+                .await
+            {
+                warn!("update to a new cluster state failed, error {e}");
+            } else {
+                // FIXME: The updates of leader and cluster states are not atomic,
```
Phoenix500526, line 585: How come? I don't get the point.

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
iGxnon: > And there is a field linearizable in ~FetchClusterResponse~ to enable quick cluster status retrieval (may return the stale data).

It is `MemberListRequest`, I missed them up. FetchCluster is the only method for the curp client to correct the local cluster state and leader state. Retrying may result in an increase in RT for a particular RPC, but it will not affect the overall RT and QPS.

curp/src/client_new/unary.rs
iGxnon: OK, I'll correct it.

curp/src/client_new/unary.rs
iGxnon: I don't get it. Can Relaxed Ordering here also serve as an optimistic lock?

curp/src/client_new/unary.rs
iGxnon: Assuming that two `FetchClusterRequests` invoke simultaneously, when obtaining the results, we first update the cluster state and then update the leader state; this is not atomic. The final result may be that we obtain the cluster state of the second `FetchClusterResponse` and the leader state of the first `FetchClusterResponse`.

iGxnon: Now it is atomic

curp/src/client_new/unary.rs
iGxnon: Fixed

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,625 @@
...
+                .unwrap_or_else(|| unreachable!("{same} must in old connects"));
+            let addrs = member_addrs
+                .get(&same)
+                .unwrap_or_else(|| unreachable!("{same} must in new member addrs"))
+                .clone();
+            conn.update_addrs(addrs).await?;
+        }
+        // Update the leader state here to ensure that each update of the cluster
+        // state and the leader state are atomic.
+        leader_update();
```
bsbds, line 155: The entire `check_and_update` does not seems to be atomic, maybe we need an `RwLock` for all the states? We rarely need to update the states so I think performance is not critcal here.

bsbds:
There's a tmp commit message in 8c5313f9a6a8c290af5bc8628869eb9660c48b27

curp/src/client_new/unary.rs
iGxnon: L110~117 should form an optimistic lock, and only a specific thread with the highest version can enter the subsequent update process; the others will return Ok(()) early.

```rust
        let update = self.version.fetch_update(
            std::sync::atomic::Ordering::Acquire,
            std::sync::atomic::Ordering::Acquire,
            |cur| (cluster_version > cur).then_some(cluster_version),
        );
        if update.is_err() {
            return Ok(());
        }
```

iGxnon:
> There's a tmp commit message in [8c5313f](https://github.com/xline-kv/Xline/commit/8c5313f9a6a8c290af5bc8628869eb9660c48b27)

Reworded

curp/src/rpc/connect.rs
```diff
@@ -468,6 +467,119 @@ impl InnerConnectApi for Connect<InnerProtocolClient<Channel>> {
...
+    /// Send `FetchReadStateRequest`
+    async fn fetch_read_state(
+        &self,
+        request: FetchReadStateRequest,
+        _timeout: Duration,
+    ) -> Result<tonic::Response<FetchReadStateResponse>, CurpError> {
+        let mut req = tonic::Request::new(request);
+        req.metadata_mut().inject_current();
+        self.server.fetch_read_state(req).await.map_err(Into::into)
+    }
```
themanforfree, line 580: Use macros to reduce repetition

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,625 @@
...
+        let _ig = self
+            .cluster_state
+            .connects
+            .insert(id, Arc::new(BypassedConnect::new(id, server)));
+    }
+
+    /// Give a handle `f` to apply to all servers *concurrently* and return a stream to poll result one by one.
+    fn for_each_server<R, F: Future<Output = R>>(
+        &self,
+        f: impl FnMut(Arc<dyn ConnectApi>) -> F,
```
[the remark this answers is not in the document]
iGxnon: `FnMut` is the supertrait of `Fn`. `std::iter::Iterator::map` requires `FnMut`

curp/src/client_new/unary.rs
iGxnon: > If the current `cluster_version` is 1, and two threads call this method with 2 and 3 respectively, then both 1->2 and 2->3 changes may be successful, and the subsequent update process will be executed concurrently.This doesn't seem to guarantee that only one thread will go to the subsequent update process.

Yes, this issue still exists; programming without locks always tends to confuse me.

curp/src/rpc/connect.rs
iGxnon: This file doesn't seem to be very large. I think it can be ignored.
If there are any changes later on (such as adding metrics), using macros is not suitable.

curp/src/client_new/unary.rs
iGxnon: It is refactored now

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,625 @@
...
+    ) -> (usize, impl Stream<Item = R>) {
+        let connects = self
+            .cluster_state
+            .connects
+            .iter()
+            .map(|connect| Arc::clone(&connect))
+            .collect_vec();
```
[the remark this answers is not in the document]
iGxnon: Done

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,625 @@
...
+        )
+    }
+
+    /// Get a handle `f` and return the future to apply `f` on the leader.
+    /// NOTICE:
+    /// The leader might be outdate if the local `leader_state` is stale.
+    /// `for_leader` should never be invoked in [`ClientApi::fetch_cluster`]
+    /// `for_leader` might call `fetch_leader_id`, `fetch_cluster`, finally
+    /// result in stack overflow.
+    async fn for_leader<R, F: Future<Output = R>>(
```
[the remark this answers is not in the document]
iGxnon: Done

curp/src/client_new/unary.rs
iGxnon: It it refactoed now

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
Phoenix500526: if it doesn’t matter that the retrieved leader_id and term are outdated, you can simply read them directly from the local State without sending a FetchCluster Request. Reason has it that obtaining leader_id and term should always be linearizable, but MemberList is another story. 

IMO, we’d better move out the leader_id and the term from FetchClusterResponse if FetchClusterRequest must maintain linearizable field.  Alternatively, we can remove the linearizable from FetchClusterRequest to enforce that FetchCluster must be linearizable.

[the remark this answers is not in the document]
iGxnon: No, Xline Cluster Server could send a non-linearizable `MemberListRequest`, we should implement it in FetchCluster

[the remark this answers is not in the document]
Phoenix500526: @rogercloud

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,590 @@
...
+        if state.cluster_version >= res.cluster_version {
+            debug!(
+                "ignore old cluster version({}) from server",
+                res.cluster_version
+            );
+            return Ok(());
+        }
+
+        debug!("client cluster version updated to {}", res.cluster_version);
+        state.cluster_version = res.cluster_version;
```
bsbds, line 153: Please fix the cluster version issue addressed in today's meeting, that the cluster version could fall back to an older value but is still valid.

iGxnon: I am modifying the implementation of the cluster version. There are issues with using a non-decreasing cluster version, so I am planning to refactor it into a hash value.

from

```txt
server: 10(old) -> 8 -> 10(new)
client: ... -> 10(old)

server will not reject the cluster version 10(old) from client.
```

to

```txt
server: #sha1 -> #sha2 -> #sha3
client: ... -> #sha1

server will only accept the cluster version #sha3.
```

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
iGxnon: After the reconstruction of ClusterServer, it is changed to full linearizable.

bsbds (approved):

Phoenix500526 (approved):
LGTM

PR #582, opened 2023-12-20
Refactor/curp client tests suits

iGxnon:
Subsequent PR of #584 , add tests

### DIFF
--- a/curp/src/client_new/mod.rs
+++ b/curp/src/client_new/mod.rs
@@ -7,6 +7,10 @@ mod unary;
 /// Retry layer
 mod retry;
 
+/// Tests for client
+#[cfg(test)]
+mod tests;
+
 use std::{collections::HashMap, sync::Arc};
 
 use async_trait::async_trait;
@@ -14,11 +18,16 @@ use curp_external_api::cmd::Command;
 use futures::{stream::FuturesUnordered, StreamExt};
 use utils::config::ClientConfig;
 
+use self::{
+    retry::Retry,
+    unary::{UnaryBuilder, UnaryConfig},
+};
 use crate::{
+    client_new::retry::RetryConfig,
     members::ServerId,
     rpc::{
         connect::ConnectApi, protocol_client::ProtocolClient, ConfChange, FetchClusterRequest,
-        FetchClusterResponse, Member, ReadState,
+        FetchClusterResponse, Member, Protocol, ReadState,
     },
 };
 
@@ -99,35 +108,38 @@ trait LeaderStateUpdate {
 
 /// Client builder to build a client
 #[derive(Debug, Clone)]
-pub struct ClientBuilder {
-    /// local server id
-    local_server_id: Option<ServerId>,
+pub struct ClientBuilder<P: Protocol> {
+    /// local server
+    local_server: Option<(ServerId, P)>,
     /// initial cluster version
     cluster_version: Option<u64>,
+    /// initial cluster members
+    all_members: Option<HashMap<ServerId, Vec<String>>>,
+    /// initial leader state
+    leader_state: Option<(ServerId, u64)>,
     /// client configuration
     config: ClientConfig,
-    /// initial all members
-    all_members: Option<HashMap<ServerId, Vec<String>>>,
 }
 
-impl ClientBuilder {
+impl<P: Protocol> ClientBuilder<P> {
     /// Create a client builder
     #[inline]
     #[must_use]
     pub fn new(config: ClientConfig) -> Self {
         Self {
-            local_server_id: None,
-            config,
-            all_members: None,
+            local_server: None,
             cluster_version: None,
+            all_members: None,
+            leader_state: None,
+            config,
         }
     }
 
-    /// Set the local server id
+    /// Set the local server to bypass `gRPC` request
     #[inline]
     #[must_use]
-    pub fn local_server_id(&mut self, id: ServerId) -> &mut Self {
-        self.local_server_id = Some(id);
+    pub fn bypass(&mut self, id: ServerId, server: P) -> &mut Self {
+        self.local_server = Some((id, server));
         self
     }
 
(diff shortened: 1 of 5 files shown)


curp/src/client_new/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+        );
+        if let Some(version) = self.cluster_version {
+            builder.set_cluster_version(version);
+        }
+        if let Some((id, server)) = self.local_server {
+            builder.set_local_server(id, server);
+        }
+        if let Some((id, term)) = self.leader_state {
+            builder.set_leader_state(id, term);
+        }
```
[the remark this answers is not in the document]
bsbds: Also, the `cluster_version` and `leader_state` is required by the client. Should we return an error if failed to get these infomations during discovery?

curp/src/client_new/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+        CurpError::shutting_down(),
+        CurpError::invalid_config(),
+        CurpError::node_already_exists(),
+        CurpError::node_not_exist(),
+        CurpError::learner_not_catch_up(),
+        CurpError::expired_client_id(),
+        CurpError::wrong_cluster_version(),
+        CurpError::redirect(Some(1), 0),
+    ] {
+        assert!(early_err.return_early());
```
bsbds, line 275: The `CurpError` is non-exhaustive, and one may forget to update these tests after updating the variants. Maybe we need a way to explicitly define these errors?

curp/src/client_new/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+                1 | 2 | 3 => ProposeResponse::new_empty(),
+                4 => return Err(CurpError::key_conflict()),
+                _ => unreachable!("there are only 5 nodes"),
+            };
+            Ok(tonic::Response::new(resp))
+        });
+        conn.expect_wait_synced()
+            .return_once(move |_req, _timeout| {
+                assert!(id == 0, "wait synced should send to leader");
+                std::thread::sleep(Duration::from_millis(100));
```
bsbds, line 448: It seems this will block the current thread and the assert below will always hold, maybe specify the runtime to multi threaded?

curp/src/client_new/tests.rs
```diff
@@ -0,0 +1,645 @@
...
+                1 | 2 => ProposeResponse::new_empty(),
+                3 | 4 => return Err(CurpError::key_conflict()),
+                _ => unreachable!("there are only 5 nodes"),
+            };
+            Ok(tonic::Response::new(resp))
+        });
+        conn.expect_wait_synced()
+            .return_once(move |_req, _timeout| {
+                assert!(id == 0, "wait synced should send to leader");
+                std::thread::sleep(Duration::from_millis(100));
```
bsbds, line 495: Same as above.

curp/src/client_new/tests.rs
iGxnon: No effect, as this is not a block that can be suspended(i.e. no `.await`). There is always only one task running in the entrie test case. 

It seems that we cannot pass an async block into `return_once`

The sleep here implies that we did indeed execute wait sync.

curp/src/client_new/tests.rs
iGxnon: Same as above.

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
iGxnon: > Also, the `cluster_version` and `leader_state` is required by the client

They are not required.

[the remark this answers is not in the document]
iGxnon: > It seems the unary builder could directly accept an `Option<T>` in the set methods.

I think that the semantics of `optional` are already reflected when the user either invokes or does not invoke the `set_xxx`.

curp/src/client_new/tests.rs
iGxnon: Maybe we could defined a `fn values() -> Vec<CurpError>` method in `CurpError` ?

curp/src/client_new/tests.rs
```diff
@@ -0,0 +1,644 @@
...
+                    1 | 4 => FetchClusterResponse {
+                        leader_id: Some(0),
+                        term: 2,
+                        cluster_id: 123,
+                        members: vec![], // linearizable read from follower returns empty members
+                        cluster_version: 1,
+                    },
+                    2 => FetchClusterResponse {
+                        leader_id: None, // imagine this node is a disconnected candidate
+                        term: 23,        // with a high term
```
Phoenix500526, line 133: Xline has implemented the pre-vote phase in release 0.6.0, therefore it's impossible to happen.

curp/src/client_new/tests.rs
bsbds: I think adding a note to the definition of `CurpError` will also do the trick.

curp/src/client_new/unary.rs
```diff
@@ -0,0 +1,698 @@
...
+                    warn!("fetch cluster from {} failed, {:?}", id, e);
+                    if e.return_early() {
+                        return Err(e);
+                    }
+                    err = Some(e);
+                    continue;
+                }
+            };
+            // Ignore the response of a node that doesn't know who the leader is.
+            // some disconnected candidates may continue to increase term, disrupting the process below.
```
themanforfree, line 634: Because of pre-vote, disconnected candidate will not increase term

curp/src/client_new/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+            RetryConfig::new_fixed(
+                *self.config.initial_retry_timeout(),
+                *self.config.retry_count(),
+                false,
+            )
+        };
+        let client = Retry::new(unary, retry_config);
+        Ok(client)
+    }
+}
```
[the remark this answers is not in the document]
iGxnon: We could have a dual insurance mechanism.

curp/src/client_new/unary.rs
iGxnon: We could have a dual insurance mechanism.

curp/src/client_new/tests.rs
iGxnon: added

curp/src/client_new/mod.rs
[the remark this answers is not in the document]
iGxnon: The construction process of `Unary` is suitable to be placed in a separate function. Some of the constructor parameters need to be processed, and it might be better to put them in a separate builder.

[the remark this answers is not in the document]
themanforfree: I still can't understand the benefits of this, can you provide a concrete example?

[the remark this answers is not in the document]
iGxnon: used in https://github.com/xline-kv/Xline/pull/585/files#diff-849b69e9d6b4f2aa4030a8ad213cae06998eef92a40061e98e2bc7050d64b269

Phoenix500526 (approved):
LGTM

themanforfree (approved):
LGTM

PR #584, opened 2023-12-21
Refactor/curp client retry

iGxnon:
Based on #590 

    Add retry layer for curp client

    As above

    No

### DIFF
--- a/curp/src/client_new/retry.rs
+++ b/curp/src/client_new/retry.rs
@@ -49,6 +49,8 @@ struct Backoff {
     config: RetryConfig,
     /// Current delay
     cur_delay: Duration,
+    /// Total RPC count
+    count: usize,
 }
 
 impl RetryConfig {
@@ -79,6 +81,7 @@ impl RetryConfig {
         Backoff {
             config: self.clone(),
             cur_delay: self.delay,
+            count: self.count,
         }
     }
 }
@@ -86,10 +89,10 @@ impl RetryConfig {
 impl Backoff {
     /// Get the next delay duration, None means the end.
     fn next_delay(&mut self) -> Option<Duration> {
-        if self.config.count == 0 {
+        if self.count == 0 {
             return None;
         }
-        self.config.count.sub_assign(1);
+        self.count.sub_assign(1);
         let mut cur = self.cur_delay;
         if let BackoffConfig::Exponential { max_delay } = self.config.backoff {
             self.cur_delay = self
--- a/curp/src/client_new/retry.rs
+++ b/curp/src/client_new/retry.rs
@@ -38,8 +38,6 @@ pub(super) struct RetryConfig {
     delay: Duration,
     /// Retry count
     count: usize,
-    /// Enable jitter to randomize the delay to avoid thundering herd
-    jitter: bool,
 }
 
 /// Backoff tool
@@ -55,24 +53,22 @@ struct Backoff {
 
 impl RetryConfig {
     /// Create a fixed retry config
-    fn new_fixed(delay: Duration, count: usize, jitter: bool) -> Self {
+    fn new_fixed(delay: Duration, count: usize) -> Self {
         assert!(count > 0, "retry count should be larger than 0");
         Self {
             backoff: BackoffConfig::Fixed,
             delay,
             count,
-            jitter,
         }
     }
 
     /// Create a exponential retry config
-    fn new_exponential(delay: Duration, max_delay: Duration, count: usize, jitter: bool) -> Self {
+    fn new_exponential(delay: Duration, max_delay: Duration, count: usize) -> Self {
         assert!(count > 0, "retry count should be larger than 0");
         Self {
             backoff: BackoffConfig::Exponential { max_delay },
             delay,
             count,
-            jitter,
         }
     }
 
@@ -100,13 +96,6 @@ impl Backoff {
                 .checked_mul(2)
                 .unwrap_or(self.cur_delay)
                 .min(max_delay);
-        }
-        #[allow(clippy::float_arithmetic)] // It is always correct.
-        if self.config.jitter {
-            // jitter algorithm will randomly pick a delay between [0.5 * delay, 1.5 * delay)
-            let per: f32 = rand::random();
-            let cur_sec = cur.as_secs_f32() * (0.5 + per);
-            cur = Duration::from_secs_f32(cur_sec);
         }
         Some(cur)
     }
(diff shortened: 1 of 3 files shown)


curp/src/client_new/retry.rs
```diff
@@ -0,0 +1,256 @@
...
+/// Retry config to control the retry policy
+#[derive(Debug, Clone)]
+pub(super) struct RetryConfig {
+    /// Backoff config
+    backoff: BackoffConfig,
+    /// Initial delay
+    delay: Duration,
+    /// Retry count
+    count: usize,
+    /// Enable jitter to randomize the delay to avoid thundering herd
```
Phoenix500526, line 41: Why should we handle thundering herd on the client side? IMO, the thundering herd occurs when a large number of processes or threads waiting for an event are awoken when that event occurs, but only one process will finally be able to handle that event. The tonic server will spawn the relavent task to handle the rpc call from client. These rpc calls needn't wait for the same event.

iGxnon: > Why should we handle thundering herd on the client side?

The CURP client can also be on the server.

I think it is meaningful, at the same time, a large number of retry requests occurring can also impose significant pressure on the entire system. (Perhaps the opportunity for a large number of retries is because the quantity of requests is too large for the system to handle. In this case, jitter can be used to introduce some time variance to these requests during the next retry, allowing the system to break out of this loop.)

iGxnon: > only one process

Here, the `only one process` refers to the entire system.

curp/src/client_new/retry.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// The retry config
+    config: RetryConfig,
+    /// Current delay
+    cur_delay: Duration,
+}
+
+impl RetryConfig {
+    /// Create a fixed retry config
+    fn new_fixed(delay: Duration, count: usize, jitter: bool) -> Self {
+        assert!(count > 0, "retry count should be larger than 0");
```
bsbds, line 57: Maybe we could offer the ability to disable retry by allowing the count to be zero?

curp/src/client_new/retry.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /// Send propose to the whole cluster, `use_fast_path` set to `false` to fallback into ordered
+    /// requests (event the requests are commutative).
+    async fn propose(
+        &self,
+        cmd: &Self::Cmd,
+        use_fast_path: bool,
+    ) -> Result<ProposeResponse<Self::Cmd>, tonic::Status> {
+        self.retry::<_, _>(|client| async move { client.propose(cmd, use_fast_path).await })
```
bsbds, line 205: It seems we could replace the async block with just `client.propose(cmd, use_fast_path)`.

curp/src/client_new/retry.rs
iGxnon: set `count` to `1` just disable retring.

curp/src/client_new/retry.rs
iGxnon: done

curp/src/client_new/retry.rs
```diff
@@ -0,0 +1,255 @@
...
+            return None;
+        }
+        self.config.count.sub_assign(1);
+        let mut cur = self.cur_delay;
+        if let BackoffConfig::Exponential { max_delay } = self.config.backoff {
+            self.cur_delay = self
+                .cur_delay
+                .checked_mul(2)
+                .unwrap_or(self.cur_delay)
+                .min(max_delay);
```
themanforfree, line 99: Can we use `self.config.delay` directly?

curp/src/client_new/retry.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
```
themanforfree, line 100: `config` should be read-only, and we'd better create a writable `count` like `cur_delay`, rather than just decrease `count` in `config`

iGxnon: done

curp/src/client_new/retry.rs
Phoenix500526: But it's not the thundering herd problem. Please modify the comment.

iGxnon: It is thundering herd problem

Phoenix500526: @rogercloud

LingKa28: I have a little question. I think nodes in a cluster should normally be deployed on different servers (I'm not sure if this is correct). Each node will receive requests from different clients. These different clients don't seem to do this because of differences in send times and network latency etc.

iGxnon: removed

curp/src/rpc/mod.rs
```diff
@@ -635,24 +635,36 @@ impl CurpError {
...
+            }
+        }
     }
 }
 
+/// The priority of curp error, indicate which error should be handled in retry layer
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum CurpErrorPriority {
+    /// Low priority
+    Low = 1,
```
Phoenix500526, line 661: Please add a comment to explain what should you do when encountering a low-priority `CurpError`? IMO, a better way is to rename the `Low` to something, like `ReturnImmediately`, a more self-explain name.

curp/src/rpc/mod.rs
```diff
@@ -635,24 +635,36 @@ impl CurpError {
...
+    /// High priority
+    High = 2,
```
Phoenix500526, line 663: Same as above.

curp/src/rpc/mod.rs
iGxnon: Added some comments

Phoenix500526: Both `High` and `ReturnImmediately` will cause the RPC return early, what's the difference between them?

iGxnon: The client receiving a high-priority error does not necessarily mean the request cannot be completed, but receiving a ReturnImmediately error will trigger a direct retry.

Phoenix500526: Ok, I get it. Please add this difference in the comment.

iGxnon: Added

Phoenix500526 (approved):
LGTM

themanforfree (approved):
LGTM

PR #588, opened 2023-12-28
feat: implements transaction api in engine

bsbds:
Based on: #548

  Implements transaction api in engine

  same as above.

  No


bsbds:
Closed as it will be replaced by another PR.

PR #590, opened 2023-12-28
fix: fix cluster version update

iGxnon:

    Solve https://github.com/xline-kv/Xline/pull/581#discussion_r1436447062

### DIFF
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -11,7 +11,7 @@ use std::{
 use dashmap::{mapref::one::Ref, DashMap};
 use futures::{stream::FuturesUnordered, StreamExt};
 use itertools::Itertools;
-use tracing::debug;
+use tracing::{debug, info};
 
 use crate::rpc::{self, FetchClusterRequest, FetchClusterResponse, Member};
 
@@ -297,7 +297,7 @@ impl ClusterInfo {
                 addrs.hash(&mut hasher);
             });
         let ver = hasher.finish();
-        debug!("cluster version updates to {ver}");
+        info!("cluster version updates to {ver}");
         self.cluster_version.store(ver, Ordering::Relaxed);
     }
 
--- a/curp/src/members.rs
+++ b/curp/src/members.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::{hash_map::DefaultHasher, HashMap},
-    hash::Hasher,
+    hash::{Hash, Hasher},
     sync::{
         atomic::{AtomicU64, Ordering},
         Arc,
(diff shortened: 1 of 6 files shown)


curp/src/members.rs
```diff
@@ -285,14 +285,20 @@ impl ClusterInfo {
...
+        self.all_members_addrs()
+            .into_iter()
+            .sorted()
+            .for_each(|(id, mut addrs)| {
+                id.hash(&mut hasher);
+                addrs.sort();
+                addrs.hash(&mut hasher);
+            });
+        let ver = hasher.finish();
+        debug!("cluster version updates to {ver}");
```
Phoenix500526, line 300: Cluster version update event should be info level.

curp/tests/it/server.rs
```diff
@@ -299,7 +299,7 @@ async fn shutdown_rpc_should_shutdown_the_cluster() {
     assert!(matches!(res, Err(ClientError::ShuttingDown)));
 
     let collection = collection_task.await.unwrap();
-    sleep_secs(3).await; // wait for the cluster to shutdown
+    sleep_secs(7).await; // wait for the cluster to shutdown
```
Phoenix500526, line 302: Why should we enlarge the sleep time in this test case. The code covered by this test case doesn't involve the parts you modified in this pr.

curp/src/members.rs
```diff
@@ -285,14 +285,20 @@ impl ClusterInfo {
...
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
+        self.all_members_addrs()
+            .into_iter()
+            .sorted()
+            .for_each(|(id, mut addrs)| {
+                id.hash(&mut hasher);
+                addrs.sort();
+                addrs.hash(&mut hasher);
+            });
+        let ver = hasher.finish();
```
bsbds, line 299: I suggest to use at least 128bit hash to avoid collision. FYI: https://stackoverflow.com/questions/62664761/probability-of-hash-collision

iGxnon: Protobuf does not like u128. This is somewhat similar to the inflight id.

iGxnon: > The birthday paradox specifically says that once you've seen roughly √(2k) items, there's a 50% chance of a collision, where k is the number of distinct possible outputs.
> In the case where the hash function hashes to an n-bit output, this means that you'll need roughly 2^(n/2) hashes before you get a collision.

In our case, 2^(n/2) = 2 ^ 32 = 4,294,967,296, which is a huge number. I think nobody could send configuration changes at this scale.

curp/src/server/raw_curp/tests.rs
```diff
@@ -777,7 +777,6 @@ fn add_node_should_add_new_node_to_curp() {
         old_cluster.all_members(),
         cluster_after_fallback.all_members()
     );
-    assert_eq!(cluster_after_fallback.cluster_version(), 0);
```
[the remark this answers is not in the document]
iGxnon: done

curp/tests/it/server.rs
iGxnon: done

curp/src/members.rs
iGxnon: done

curp/tests/it/server.rs
iGxnon: After replacing it to the new client, I will test this time again.

Phoenix500526: Are you planning to complete this test in this PR, or in another PR? If it's in this PR, please resolve this conversation after the test is completed. If it's in another PR, please open an issue to track this matter.

iGxnon: I've submitted a new issue: https://github.com/xline-kv/Xline/issues/596

bsbds (approved):

Phoenix500526 (approved):
LGTM

PR #601, opened 2024-01-08
Feat/tls

themanforfree:
tls


curp/src/client/mod.rs
```diff
@@ -274,6 +296,8 @@ impl ClientBuilder {
             UnaryConfig::new(
                 *self.config.propose_timeout(),
                 *self.config.wait_synced_timeout(),
+                #[cfg(not(madsim))]
+                None,
```
iGxnon, line 300: Why None?

xline-client/src/lib.rs
```diff
@@ -403,8 +425,19 @@ impl ClientOptions {
...
+    }
+
+    /// Set `tls_config`
+    #[inline]
+    #[must_use]
+    #[cfg(not(madsim))]
+    pub fn with_tls_config(self, tls_config: ClientTlsConfig) -> Self {
+        Self {
+            tls_config: Some(tls_config),
+            ..self
```
iGxnon, line 440: Should `ClientTlsConfig` be a part of `ClientConfig`?

xline/src/server/xline_server.rs
```diff
@@ -71,6 +73,12 @@ pub struct XlineServer {
     compact_cfg: CompactConfig,
     /// Server timeout
     server_timeout: ServerTimeout,
+    /// Client tls config
+    #[cfg(not(madsim))]
+    client_tls_config: Option<ClientTlsConfig>,
```
iGxnon, line 78: Same as above, should it be a client config?

Cargo.toml
```diff
@@ -22,9 +22,9 @@ resolver = "2"
...
-madsim = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim" }
-tokio = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tokio" }
-tonic = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tonic" }
-tonic-build = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tonic-build" }
+# # TODO: We should remove it when the madsim updates its release in the crate.io
+# [patch.crates-io]
+# madsim = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim" }
+# tokio = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tokio" }
+# tonic = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tonic" }
+# tonic-build = { git = "https://github.com/madsim-rs/madsim.git", rev = "5eb4198", package = "madsim-tonic-build" }
```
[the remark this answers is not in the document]
themanforfree: Removed

curp/src/client/mod.rs
themanforfree: Fixed

xline-client/src/lib.rs
themanforfree: Currently, the content in `ClientConfig` is directly obtained from command line arguments or configuration files, while `ClientTlsConfig` needs to read the certificate file and cannot be obtained directly, they can't contain each other for now, but our config logic may be modified when config is refactored in the future

xline/src/server/xline_server.rs
themanforfree: Same as above

curp/src/members.rs
```diff
@@ -107,6 +109,8 @@ impl ClusterInfo {
     }
 
     /// Construct a new `ClusterInfo` from `FetchClusterResponse`
+    /// # Panics
+    /// panic if `cluster.members` does contain `self_addr`
```
Phoenix500526, line 113: doesn't contain `self_addr`?

curp/src/rpc/connect.rs
```diff
@@ -64,15 +66,16 @@ impl FromTonicChannel for InnerProtocolClient<Channel> {
 /// Connect to a server
 async fn connect_to<Client: FromTonicChannel>(
     id: ServerId,
-    mut addrs: Vec<String>,
+    addrs: Vec<String>,
+    #[cfg(not(madsim))] tls_config: Option<ClientTlsConfig>,
```
Phoenix500526, line 70: We can define two `connect_to` methods, one with `#[cfg(not(madsim))]` while the other with `#[cfg(madsim)]`. And define a mock `ClientTlsConfig>` in `simulation/src/lib.rs`. Using conditional compile to pick which one should be use. Therefore, we can remove many `#[cfg(madsim)]` in other codes of this pr.

Phoenix500526 (commented):
Etcd supports authenticate a user via certificate. In this way, we don't need the password any more. But I don't see any relevant change in this pr. Will this feature be implemented in other pr?  FYI: https://www.cnblogs.com/traditional/p/17444921.html

curp/src/members.rs
themanforfree: Fixed

curp/src/rpc/connect.rs
themanforfree: Refactored

iGxnon (dismissed):

utils/src/lib.rs
```diff
@@ -143,6 +143,26 @@
...
+#[allow(missing_copy_implementations)] // Same as real config
+#[non_exhaustive]
+pub struct ClientTlsConfig;
+
+/// Fake `ServerTlsConfig` for `madsim`
+#[cfg(madsim)]
+#[derive(Debug, Clone)]
+#[allow(missing_copy_implementations)] // Same as real config
+#[non_exhaustive]
+pub struct ServerTlsConfig;
```
Phoenix500526, line 164: Is it better to place these two struct in `simulation/src/lib.rs`?

Phoenix500526 (commented):
I think that hardcode the `*.crt`, `*.key` in crate is not a good practice. I recommend to place them in the top directory of Xline, and named it `fixtures`, or something like that.

utils/src/lib.rs
themanforfree: `xline` also depends on these fake configs, and `xline` shouldn't depend on `simulation`

Phoenix500526 (approved):
LGTM

iGxnon (approved):
LGTM

iGxnon (approved):

Phoenix500526 (approved):
LGTM

PR #602, opened 2024-01-09
[Refactor]: test framework

themanforfree:

support start `CurpGroup` and `Cluster` with external configuration
Refactor test framework
no

### DIFF
--- a/curp/tests/it/common/curp_group.rs
+++ b/curp/tests/it/common/curp_group.rs
@@ -22,7 +22,7 @@ use engine::{
     Engine, EngineType, MemorySnapshotAllocator, RocksSnapshotAllocator, Snapshot,
     SnapshotAllocator,
 };
-use futures::future::join_all;
+use futures::{future::join_all, Future};
 use itertools::Itertools;
 use tokio::{
     net::TcpListener,
@@ -33,7 +33,9 @@ use tokio::{
 use tokio_stream::wrappers::TcpListenerStream;
 use tracing::debug;
 use utils::{
-    config::{default_quota, ClientConfig, CurpConfigBuilder, EngineConfig, StorageConfig},
+    config::{
+        default_quota, ClientConfig, CurpConfig, CurpConfigBuilder, EngineConfig, StorageConfig,
+    },
     shutdown::{self, Trigger},
 };
 pub mod commandpb {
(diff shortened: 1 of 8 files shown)


xline-test-utils/src/lib.rs
```diff
@@ -197,6 +213,63 @@ impl Cluster {
         let decoding_key = DecodingKey::from_rsa_pem(public_key).unwrap();
         Some((encoding_key, decoding_key))
     }
+
+    pub fn default_cluster_config() -> ClusterConfig {
```
Phoenix500526, line 217: We can derive Debug trait for `ClusterConfig` rather than implement it manually.

xline-test-utils/src/lib.rs
```diff
@@ -27,19 +30,34 @@ pub struct Cluster {
...
-    /// Cluster size
-    size: usize,
-    /// storage paths
-    paths: HashMap<usize, PathBuf>,
     /// Xline servers
     servers: Vec<Arc<XlineServer>>,
-    /// Cluster quotas
-    pub quotas: HashMap<usize, u64>,
+    /// Server configs
+    configs: HashMap<usize, XlineServerConfig>,
```
Phoenix500526, line 36: Why don't we use `Vec` here?

curp/tests/it/common/curp_group.rs
```diff
@@ -61,141 +64,169 @@ pub struct CurpGroup {
...
+            listener,
+            name,
+            cluster_info,
+            Arc::new(CurpConfig::default()),
+            EngineConfig::default(),
+        )
+        .await
+    }
+
+    pub async fn run_node_with_config(
```
Phoenix500526, line 219: The method `run_node_with_config` is only used in the `run_node` method, therefore it dosen't need to be specified as pub. BTW, it seems we only use default config in our test. Shall we inline this method into `run_node`?

xline-test-utils/src/lib.rs
```diff
@@ -113,17 +126,23 @@ impl Cluster {
     }
 
     pub async fn run_node(&mut self, listener: TcpListener, idx: usize) {
+        let config = Self::default_config();
+        self.run_node_with_config(listener, idx, config).await;
+    }
+
+    pub async fn run_node_with_config(
```
Phoenix500526, line 133: Same as above.

xline-test-utils/src/lib.rs
```diff
@@ -52,54 +70,49 @@ impl Cluster {
...
-            paths: HashMap::new(),
             servers: Vec::new(),
-            quotas: HashMap::new(),
+            configs,
         }
     }
 
-    pub fn set_paths(&mut self, paths: HashMap<usize, PathBuf>) {
-        self.paths = paths;
+    pub fn set_paths(&mut self, _paths: HashMap<usize, PathBuf>) {
```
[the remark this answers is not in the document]
themanforfree: Removed

xline-test-utils/src/lib.rs
```diff
@@ -52,54 +70,49 @@ impl Cluster {
...
+        todo!()
+        // self.paths = paths;
     }
 
     /// Start `Cluster`
     pub async fn start(&mut self) {
-        for i in 0..self.size {
+        for i in 0..self.configs.len() {
```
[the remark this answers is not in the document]
themanforfree: Fixed

xline-test-utils/src/lib.rs
themanforfree: Fixed

xline-test-utils/src/lib.rs
themanforfree: Fixed

curp/tests/it/common/curp_group.rs
themanforfree: Sometimes we need to test configs that are not enabled by default, such as TLS, and we need to pass in config from the outside to test these things

xline-test-utils/src/lib.rs
themanforfree: Same as above

Phoenix500526 (approved):
LGTM

utils/src/config.rs
```diff
@@ -102,6 +102,24 @@ pub struct ClusterConfig {
...
+            )]),
+            is_leader: false,
+            curp_config: CurpConfig::default(),
+            client_config: ClientConfig::default(),
+            server_timeout: ServerTimeout::default(),
+            initial_cluster_state: InitialClusterState::default(),
+        }
+    }
+}
+
```
bsbds, line 122: Please adds a `cfg(test)` for the default if it is used in tests only

xline-test-utils/src/lib.rs
```diff
@@ -197,6 +196,33 @@ impl Cluster {
...
+    }
+
+    pub fn default_rocks_config() -> XlineServerConfig {
+        let path = temp_dir().join(random_id());
+        Self::default_config_with_quota_and_rocks_path(path, default_quota())
+    }
+
+    pub fn default_quota_config(quota: u64) -> XlineServerConfig {
+        let path = temp_dir().join(random_id());
+        Self::default_config_with_quota_and_rocks_path(path, quota)
```
bsbds, line 224: I suggest to use `tempfile` for managing temp dirs.

utils/src/config.rs
themanforfree: Config refactoring in the future should be based on this default value, create a default configuration, and then override it based on command line arguments or configuration files, the default value in ClusterConfig represents the startup parameters of a single-node cluster, and the default single-node cluster should be started when xline is run without passing in any parameters

xline-test-utils/src/lib.rs
themanforfree: It doesn't make much sense to use `tempfile` here, because this temp directory will eventually be deleted on the `Cluster` drop, we don't need to rely on the auto-delete capability provided by the `tempfile` library

bsbds (approved):

PR #613, opened 2024-01-13
refactor: isolate client state

iGxnon:

    Separate the client state from unary client to realize subsequent implementation of shared state between stream client and unary client.

    Isolate client state.

    No.

### DIFF
--- a/curp/src/client/state.rs
+++ b/curp/src/client/state.rs
@@ -4,6 +4,8 @@ use std::{
     sync::Arc,
 };
 
+use event_listener::Event;
+use futures::{stream::FuturesUnordered, Future};
 use tokio::sync::RwLock;
 use tracing::{debug, info};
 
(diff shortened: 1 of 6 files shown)


curp/src/client/state.rs
```diff
@@ -0,0 +1,233 @@
...
+        term: u64,
+    ) -> bool {
+        match self.term.cmp(&term) {
+            Ordering::Less => {
+                // reset term only when the resp has leader id to prevent:
+                // If a server loses contact with its leader, it will update its term for election. Since other servers are all right, the election will not succeed.
+                // But if the client learns about the new term and updates its term to it, it will never get the true leader.
+                if let Some(new_leader_id) = leader_id {
+                    debug!("client term updates to {}", term);
+                    debug!("client leader id updates to {new_leader_id}");
```
Phoenix500526, line 86: We can merge these two lines together, and use `info!` instead.

curp/src/client/state.rs
```diff
@@ -0,0 +1,233 @@
...
+                    self.term = term;
+                    self.leader = Some(new_leader_id);
+                }
+            }
+            Ordering::Equal => {
+                if let Some(new_leader_id) = leader_id {
+                    if self.leader.is_none() {
+                        debug!("client leader id updates to {new_leader_id}");
```
Phoenix500526, line 94: Here should also be info level.

curp/src/client/state.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+        if self.cluster_version == res.cluster_version {
+            debug!(
+                "ignore cluster version({}) from server",
+                res.cluster_version
+            );
+            return Ok(());
+        }
+
+        debug!("client cluster version updated to {}", res.cluster_version);
```
Phoenix500526, line 128: Same as above.

curp/src/client/state.rs
iGxnon: done

Phoenix500526 (approved):
LGTM

curp/src/client/state.rs
```diff
@@ -0,0 +1,322 @@
...
+        connect::{BypassedConnect, ConnectApi},
+        CurpError, FetchClusterResponse, Protocol,
+    },
+};
+
+/// The client state
+#[derive(Debug)]
+pub(super) struct State {
+    /// Mutable state
+    mutable: RwLock<StateMut>,
```
bsbds, line 25: We could use `parking_lot::RwLock` for this.

curp/src/client/state.rs
```diff
@@ -0,0 +1,322 @@
...
+    immutable: StateStatic,
+}
+
+/// Immutable client state, could be cloned
+#[derive(Debug, Clone)]
+struct StateStatic {
+    /// Local server id, should be initialized on startup
+    local_server: Option<ServerId>,
+    /// Notifier of leader update
+    leader_notifier: Arc<Event>,
```
bsbds, line 36: What does this notifier do?

iGxnon: Used in #594

curp/src/client/state.rs
iGxnon: What are the benefits?

iGxnon: It seems not easy to convert it into a synchronous lock. During `State::check_and_update`, there are asynchronous behaviors such as `ConnectApi::update_addrs` and `rpc::connects`. We cannot let the synchronous lock span across `.await`, as this may lead to recursive deadlock.

bsbds (approved):

Phoenix500526 (approved):
LGTM

PR #617, opened 2024-01-16
Feat/dedup lease client (part 2)

iGxnon:

    This PR implements a stream client that runs in the background and continuously updates the client id. Ensure that each RPC is initiated by online clients.
    
    As above

    No


crates/curp/src/client/tests.rs
```diff
@@ -690,3 +695,231 @@ async fn test_retry_propose_return_retry_error() {
...
+    /// Send `ProposeRequest`
+    async fn propose(
+        &self,
+        _request: ProposeRequest,
+        _timeout: Duration,
+    ) -> Result<tonic::Response<ProposeResponse>, CurpError> {
+        unreachable!("please use MockedConnectApi")
+    }
+
+    /// Send `ProposeRequest`
```
Phoenix500526, line 728: Please update the comment

crates/curp/src/client/stream.rs
```diff
@@ -0,0 +1,113 @@
...
+                                "cannot find the leader in connects, wait for  leadership update"
+                            );
+                            self.state.leader_notifier().listen().await;
+                        }
+                        CurpError::ShuttingDown(_) => {
+                            debug!("shutting down stream client background task");
+                            break Err(err);
+                        }
+                        _ => {
+                            debug!("got unexpected error {err:?} when keep heartbeat, retrying...");
```
Phoenix500526, line 94: Should here be `warn!`?

crates/curp/src/client/state.rs
```diff
@@ -29,6 +29,8 @@ pub(super) struct State {
     mutable: RwLock<StateMut>,
     /// Immutable state
     immutable: StateStatic,
+    /// The client id. Remove from `mutable`` because the client ID will be updated in the background.
```
Phoenix500526, line 32: There is a extra backtick after 'mutable'

crates/curp/src/client/tests.rs
iGxnon: done

crates/curp/src/client/stream.rs
```diff
@@ -0,0 +1,113 @@
...
+                }
+            });
+
+            tokio::select! {
+                _ = self.state.leader_notifier().listen() => {
+                    debug!("interrupt keep heartbeat because leadership changed");
+                    // TODO release the heartbeat task unless cancellation safety is ensured, especially in the case of the tonic RPC method.
+                    // TODO how to release?
+                }
+                _ = heartbeat => {
```
bsbds, line 107: The heartbeat error seems get ignored?

crates/curp/src/client/stream.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// Keep heartbeat
+    pub(super) async fn keep_heartbeat(&self) {
+        /// Prevent lock contention when leader crashed or some unknown errors
+        const RETRY_DELAY: Duration = Duration::from_millis(100);
+
+        loop {
+            // is heartbeat task cancellation safety?
+            let heartbeat = self.map_remote_leader::<(), _>(|conn| async move {
```
bsbds, line 70: I think heartbeat is cancel safe as it doesn't depend on external states.

crates/curp/src/client/stream.rs
iGxnon: Yes, It seems that the only possible value (`ShuttingDown` error) does not need to be handled at the moment.

crates/curp/src/client/stream.rs
iGxnon: comment removed

bsbds (approved):

Phoenix500526 (approved):
LGTM

bsbds (approved):

PR #620, opened 2024-01-17
chore: bump rust to 1.71.0

iGxnon:

    Refer to https://github.com/rust-lang/rust/issues/112430.
    When switching branches or writing mockall tests, we may encounter bugs in rustc's incremental compilation. The specific reason is unknown, but updating rustc and mockall versions can resolve it.

   Bump rust to 1.71.0

   Exposing APIs to users will not, but for developers, they need to update their own rustc.

### DIFF
--- a/benchmark/src/runner.rs
+++ b/benchmark/src/runner.rs
@@ -82,6 +82,7 @@ impl Stats {
     /// Calculate the histogram from the latencies.
     #[allow(clippy::indexing_slicing)]
     #[allow(clippy::let_underscore_must_use)] // the writeln! will not fail thus always return an Ok
+    #[allow(clippy::arithmetic_side_effects)] // test only
     pub fn histogram(&self) -> String {
         let size = 10;
         let mut buckets = Vec::with_capacity(size);
@@ -254,7 +255,8 @@ impl CommandRunner {
     #[allow(
         clippy::as_conversions,
         clippy::cast_precision_loss,
-        clippy::float_arithmetic
+        clippy::float_arithmetic,
+        clippy::arithmetic_side_effects
     )]
     async fn collecter(&mut self, mut rx: Receiver<CmdResult>, barrier: Arc<Barrier>) -> Stats {
         let bar_len = match self.args.command {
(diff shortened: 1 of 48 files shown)


workspace-hack/Cargo.toml
```diff
@@ -26,7 +26,8 @@ getrandom = { version = "0.2", default-features = false, features = ["js", "rdra
 hashbrown = { version = "0.14", default-features = false, features = ["raw"] }
 hyper = { version = "0.14", features = ["full"] }
 indexmap = { version = "1", default-features = false, features = ["std"] }
-itertools = { version = "0.10" }
+itertools-93f6ce9d446188ac = { package = "itertools", version = "0.10" }
+itertools-a6292c17cd707f01 = { package = "itertools", version = "0.11" }
```
Phoenix500526, line 30: Can we unify these two itertools versions?

benchmark/src/runner.rs
```diff
@@ -82,6 +82,7 @@ impl Stats {
     /// Calculate the histogram from the latencies.
     #[allow(clippy::indexing_slicing)]
     #[allow(clippy::let_underscore_must_use)] // the writeln! will not fail thus always return an Ok
+    #[allow(clippy::arithmetic_side_effects)] // test only
```
Phoenix500526, line 85: Should the comment here be "benchmark only"?

workspace-hack/Cargo.toml
iGxnon: Probably not, this is generated by Hakari.

iGxnon: 0.10 is what we are using, 0.11 is used by some dependent libraries. Do we need to update the itertools version?
Its latest version is 0.12.

benchmark/src/runner.rs
iGxnon: done

workspace-hack/Cargo.toml
iGxnon: updated to 0.11

bsbds (approved):

bsbds (approved):

PR #625, opened 2024-01-22
[feature]: member persistent

themanforfree:

Persist the member and restore cluster info from the store
Persist the member and restore cluster info from the store
no

### DIFF
--- a/crates/curp/src/server/raw_curp/mod.rs
+++ b/crates/curp/src/server/raw_curp/mod.rs
@@ -47,7 +47,7 @@ use self::{
     log::Log,
     state::{CandidateState, LeaderState, State},
 };
-use super::{cmd_worker::CEEventTxApi, lease_manager::LeaseManagerRef};
+use super::{cmd_worker::CEEventTxApi, lease_manager::LeaseManagerRef, storage::StorageApi, DB};
 use crate::{
     cmd::Command,
     log_entry::{EntryData, LogEntry},
@@ -137,7 +137,8 @@ pub(super) struct RawCurpArgs<C: Command, RC: RoleChange> {
     sync_events: DashMap<ServerId, Arc<Event>>,
     /// Connects of peers
     connects: DashMap<ServerId, InnerConnectApiWrapper>,
-
+    /// curp storage
+    curp_storage: Arc<DB<C>>,
     /// client tls config
     #[builder(default)]
     client_tls_config: Option<ClientTlsConfig>,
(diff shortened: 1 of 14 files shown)


curp/src/server/storage/db.rs
```diff
@@ -34,14 +50,85 @@ impl<C: Command> StorageApi for DB<C> {
...
+        Ok(())
+    }
+
+    #[inline]
+    fn put_cluster_info(&self, cluster_info: &ClusterInfo) -> Result<(), StorageError> {
+        let mut ops = Vec::new();
+        ops.push(WriteOperation::new_put(
+            CF,
+            CLUSTER_ID.to_vec(),
+            bincode::serialize(&cluster_info.cluster_id())?,
```
iGxnon, line 85: use `cluster_info.cluster_id().to_be_bytes()`

curp/src/server/storage/db.rs
```diff
@@ -34,14 +50,85 @@ impl<C: Command> StorageApi for DB<C> {
...
+        ));
+        ops.push(WriteOperation::new_put(
+            CF,
+            MEMBER_ID.to_vec(),
+            bincode::serialize(&cluster_info.self_id())?,
```
iGxnon, line 90: same as above

simulation/src/xline_group.rs
```diff
@@ -126,6 +152,20 @@ impl XlineGroup {
             .unwrap()
             .1
     }
+
+    pub async fn crash(&mut self, id: ServerId) {
+        let handle = madsim::runtime::Handle::current();
+        handle.kill(id.to_string());
+        madsim::time::sleep(Duration::from_secs(2)).await;
```
iGxnon, line 159: Is 2s enough? The time is simulated.

themanforfree: Just copy from curp_group, changed to 10s

curp/src/server/storage/db.rs
themanforfree: Done

crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -1376,7 +1384,8 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
                     .map_lock(|mut cst_l| _ = cst_l.config.remove(node_id));
                 self.lst.remove(node_id);
                 _ = self.ctx.sync_events.remove(&node_id);
-                let _ig = self.ctx.cluster_info.remove(&node_id);
+                let _ig1 = self.ctx.cluster_info.remove(&node_id);
```
bsbds, line 1387: The log persistent and member info persistent is not atomic, could this cause inconsistent between log and the member info?

themanforfree: I didn't think of a possible problem for a while, can you provide an example？

bsbds: I reconsidered about it, I think it's fine as long as the log is properly persistent and replayed because member info is persistent before log persistent. Resolved.

bsbds (approved):

iGxnon (approved):

PR #632, opened 2024-01-23
feat: implement tracker

iGxnon:

   Add tracker.

   Add `tracker.rs`.

    No

### DIFF
--- a/crates/curp/src/tracker.rs
+++ b/crates/curp/src/tracker.rs
@@ -1,0 +1,351 @@
+#![allow(unused)] // TODO remove when used
+
+use std::collections::VecDeque;
+use std::ops::{AddAssign, Sub};
+
+use clippy_utilities::NumericCast;
+
+/// Bits of usize
+const USIZE_BITS: usize = std::mem::size_of::<usize>() * 8;
+
+/// Default bit vec queue capacity, this is the number of inflight requests that a client expects.
+/// It can reduce about `log_2(DEFAULT_BIT_VEC_QUEUE_CAP)` `VecDequeue` memory reallocations.
+const DEFAULT_BIT_VEC_QUEUE_CAP: usize = 1024;
+
+/// A one-direction bit vector queue
+/// It use a ring buffer `VecDeque<usize>` to store bits
+///
+/// Memory Layout:
+///
+/// `010000000101_110111011111_000001000000000`
+///  ^        ^                ^    ^
+///  |        |________________|____|
+///  |________|         len    |____|
+///     head                    tail
+#[derive(Debug, Clone)]
+struct BitVecQueue {
+    /// Bits store
+    store: VecDeque<usize>,
+    /// Head length indicator
+    head: usize,
+    /// Tail length indicator
+    tail: usize,
+}
+
+impl Default for BitVecQueue {
+    #[inline]
+    fn default() -> Self {
+        Self::with_capacity(DEFAULT_BIT_VEC_QUEUE_CAP)
+    }
+}
+
+#[allow(clippy::arithmetic_side_effects, clippy::unwrap_used)] // They are checked
+impl BitVecQueue {
+    /// New with a capacity (in bits)
+    fn with_capacity(cap_bits: usize) -> Self {
+        Self {
+            store: VecDeque::with_capacity(cap_bits / USIZE_BITS),
+            head: 0,
+            tail: 0,
+        }
+    }
+
+    /// Get the length
+    fn len(&self) -> usize {
+        if self.store.is_empty() {
+            return 0;
+        }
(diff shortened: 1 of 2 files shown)


crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+/// A one-direction bit vector queue
+/// It use a ring buffer `VecDeque<usize>` to store bits
+///
+/// Memory Layout:
+///
+/// `010000000101_110111011111_000001000000000`
+///  ^        ^                ^    ^
+///  |        |________________|____|
+///  |________|         len    |____|
+///     head                    tail
```
themanforfree, line 24: tail is a part of len?

crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+    fn front(&self) -> Option<bool> {
+        let front = self.store.front()?;
+        Some(front & (1 << (USIZE_BITS - 1 - self.head)) != 0)
+    }
+
+    /// Peek the front usize in store and take a function to check it,
+    /// Used to batch accelerate check.
+    fn front_by(&self, f: impl Fn(usize) -> bool) -> Option<bool> {
+        self.store.front().copied().map(f)
+    }
```
themanforfree, line 119: I think it would be better to name it `map_front`

crates/curp/src/tracker.rs
iGxnon: Tail indicates how many bits are used in the last usize.

crates/curp/src/tracker.rs
iGxnon: done

crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+            tail: 0,
+        }
+    }
+
+    /// Get the length
+    fn len(&self) -> usize {
+        if self.store.is_empty() {
+            return 0;
+        }
+        (self.store.len() - 1) * USIZE_BITS + self.tail - self.head + 1
```
bsbds, line 58: Is `self.tail` always greater than or equal to `self.head + 1`?

crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    /// Split this bit vec queue to `at`
+    /// e.g.
+    /// 001100 -> `split_at(2)` -> 1100
+    fn split_at(&mut self, at: usize) {
+        let idx = self.head + at;
+        let index = idx / USIZE_BITS;
+        let slot = idx % USIZE_BITS;
+        if index == self.store.len() - 1 && slot > self.tail {
```
bsbds, line 163: Same as above.

crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+        (self.store.len() - 1) * USIZE_BITS + self.tail - self.head + 1
+    }
+
+    /// Get the bit value
+    fn get(&self, idx: usize) -> Option<bool> {
+        let idx = self.head + idx;
+        let index = idx / USIZE_BITS;
+        let slot = idx % USIZE_BITS;
+        if index == self.store.len() - 1 && slot > self.tail {
```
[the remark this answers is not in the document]
iGxnon: When len == 0, it seems could be covered with
```rust
let bits = self.store.get(index)?; // always return None when len(store) = 0
```

and

```rust
let Some(bits) = self.store.get_mut(index) else {
        // always return when len(store) = 0
        return;
};
```

crates/curp/src/tracker.rs
```diff
@@ -0,0 +1,351 @@
...
+        let bits = self.store.get(index)?;
+        Some(bits & (1 << (USIZE_BITS - 1 - slot)) != 0)
+    }
+
+    /// Set the bit value
+    fn set(&mut self, idx: usize, v: bool) {
+        let idx = self.head + idx;
+        let index = idx / USIZE_BITS;
+        let slot = idx % USIZE_BITS;
+        if index == self.store.len() - 1 && slot > self.tail {
```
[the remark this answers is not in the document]
iGxnon: Check above.

crates/curp/src/tracker.rs
iGxnon: No, head could equals to tail, when they are both 0, which means that the bitvec is empty.

crates/curp/src/tracker.rs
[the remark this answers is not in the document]
iGxnon: Add an empty checking.

bsbds (approved):

themanforfree (approved):

PR #633, opened 2024-01-24
chore: upload logs when validation cancelled

themanforfree:


scripts/validation_test.sh
```diff
@@ -1,70 +1,92 @@
...
+    bash ${QUICK_START} stop
+}
+
+trap stop EXIT
+trap stop INT
+trap stop TERM
+
+res=""
+
+function parse_line() {
```
Phoenix500526, line 21: `parse_result` would be better.

themanforfree: Done

Phoenix500526 (dismissed):
LGTM

bsbds (approved):

PR #641, opened 2024-01-30
feat: add scheduled compact revision persistent

themanforfree:

add scheduled compact revision persistent
add scheduled compact revision persistent
no

### DIFF
--- a/crates/xline/src/storage/db.rs
+++ b/crates/xline/src/storage/db.rs
@@ -22,8 +22,10 @@ use crate::{
     storage::Revision,
 };
 
-/// Key of compacted revision
-pub(crate) const COMPACT_REVISION: &str = "compact_revision";
+/// Key of finished compact revision
+pub(crate) const FINISHED_COMPACT_REVISION: &str = "finished_compact_revision";
+/// Key of scheduled compact revision
+pub(crate) const SCHEDULED_COMPACT_REVISION: &str = "scheduled_compact_revision";
 
 /// Database to store revision to kv mapping
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


crates/xline/src/storage/db.rs
```diff
@@ -58,6 +60,32 @@ impl DB {
             engine: Arc::new(engine),
         }))
     }
+
+    /// Get del lease key buffer
+    fn get_del_lease_key_buffer(ops: &[WriteOp]) -> HashMap<i64, Vec<u8>> {
```
Phoenix500526, line 65: Better to inline this function

crates/xline/src/storage/db.rs
```diff
@@ -58,6 +60,32 @@ impl DB {
...
+                    Some((lease_id, lease_id.encode_to_vec()))
+                } else {
+                    None
+                }
+            })
+            .collect::<HashMap<_, _>>()
+    }
+
+    /// get del alarm buffer
+    fn get_del_alarm_buffer(ops: &[WriteOp]) -> Vec<u8> {
```
Phoenix500526, line 78: Same as above

themanforfree: Added

crates/xline/src/storage/db.rs
themanforfree: Added

Phoenix500526 (approved):
LGTM

crates/xline/src/storage/db.rs
```diff
@@ -50,6 +52,34 @@ impl DB {
             engine: Arc::new(engine),
         }))
     }
+
+    /// Get del lease key buffer
+    #[inline]
```
iGxnon, line 57: Private function maybe always inlined by llvm.
> Second, it usually isn’t necessary to apply #[inline] to private functions — within a crate, the compiler generally makes good inline decisions. There’s [a joke](https://twitter.com/ManishEarth/status/936084757212946432) that LLVM’s heuristic for when the function should be inlined is “yes”.
> —— https://matklad.github.io/2021/07/09/inline-in-rust.html

iGxnon (approved):

PR #646, opened 2024-02-04
fix: return error if put kv with nonexistent lease id

themanforfree:

return error if put kv with nonexistent lease id
return error if put kv with nonexistent lease id
no

### DIFF
--- a/crates/xline/src/storage/kv_store.rs
+++ b/crates/xline/src/storage/kv_store.rs
@@ -598,6 +598,9 @@ where
             header: Some(self.header_gen.gen_header()),
             ..Default::default()
         };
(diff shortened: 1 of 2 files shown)


crates/simulation/tests/it/xline.rs
```diff
@@ -13,6 +13,8 @@ async fn basic_put() {
     let group = XlineGroup::new(3).await;
     let client = group.client().await;
     let res = client.put(PutRequest::new("key", "value")).await;
+    println!("{:?}", res);
```
Phoenix500526, line 16: Please remove this print statement.

themanforfree: Removed

Phoenix500526 (approved):
LGTM

bsbds (approved):

PR #645, opened 2024-02-02
feat: add auth wrapper and split ports

themanforfree:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
add auth wrapper
* what changes does this pull request make?
add auth wrapper
* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)
no

PR #648, opened 2024-02-06
[feat]: move token authentication to the network layer

themanforfree:
based on #645 

move token authentication to the network layer
move token authentication to the network layer
no

### DIFF
--- a/crates/curp/src/client/mod.rs
+++ b/crates/curp/src/client/mod.rs
@@ -64,6 +64,7 @@ pub trait ClientApi {
     async fn propose(
         &self,
         cmd: &Self::Cmd,
+        token: Option<&String>, // TODO: Allow external custom interceptors, do not pass token in parameters
         use_fast_path: bool,
     ) -> Result<ProposeResponse<Self::Cmd>, Self::Error>;
 
@@ -126,6 +127,7 @@ trait RepeatableClientApi: ClientApi {
         &self,
         propose_id: ProposeId,
         cmd: &Self::Cmd,
+        token: Option<&String>,
         use_fast_path: bool,
     ) -> Result<ProposeResponse<Self::Cmd>, Self::Error>;
 
--- a/crates/curp/tests/it/server.rs
+++ b/crates/curp/tests/it/server.rs
@@ -27,11 +27,12 @@ async fn basic_propose() {
     init_logger();
 
     let group = CurpGroup::new(3).await;
+
     let client = group.new_client().await;
 
     assert_eq!(
         client
-            .propose(&TestCommand::new_put(vec![0], 0), true)
+            .propose(&TestCommand::new_put(vec![0], 0), None, true)
             .await
             .unwrap()
             .unwrap()
@@ -40,7 +41,7 @@ async fn basic_propose() {
     );
     assert_eq!(
         client
-            .propose(&TestCommand::new_get(vec![0]), true)
+            .propose(&TestCommand::new_get(vec![0]), None, true)
             .await
             .unwrap()
             .unwrap()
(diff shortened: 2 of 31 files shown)


crates/curp/tests/it/server.rs
```diff
@@ -27,11 +27,12 @@ async fn basic_propose() {
     init_logger();
 
     let group = CurpGroup::new(3).await;
+    sleep_secs(1).await;
```
Phoenix500526, line 30: Why do you sleep one second here?

crates/curp/src/client/mod.rs
```diff
@@ -64,6 +64,7 @@ pub trait ClientApi {
     async fn propose(
         &self,
         cmd: &Self::Cmd,
+        token: Option<&String>, // TODO: Allow external custom interceptors, do not pass token in parameters
```
Phoenix500526, line 67: Please submit an issue to track this TODO

crates/xline/src/server/auth_server.rs
```diff
@@ -55,11 +65,13 @@ impl AuthServer {
     where
         T: Into<RequestWrapper>,
     {
-        let token = get_token(request.metadata());
-        let wrapper = RequestWithToken::new_with_token(request.into_inner().into(), token);
-        let cmd = command_from_request_wrapper(wrapper);
-
-        let res = self.client.propose(&cmd, use_fast_path).await??;
+        let auth_info = match get_token(request.metadata()) {
+            Some(token) => Some(self.auth_store.verify(&token)?),
```
iGxnon, line 69: If we access the storage layer here, can we ensure that permission verification is always valid?

crates/xline/src/server/auth_wrapper.rs
```diff
@@ -43,12 +49,21 @@ where
...
         &self,
-        request: tonic::Request<ProposeRequest>,
+        mut request: tonic::Request<ProposeRequest>,
     ) -> Result<tonic::Response<ProposeResponse>, tonic::Status> {
         debug!(
             "AuthWrapper received propose request: {}",
             request.get_ref().propose_id()
         );
+        if let Some(token) = get_token(request.metadata()) {
+            let auth_info = self.auth_store.verify(&token)?;
```
iGxnon, line 59: Same as above

crates/xline/src/storage/auth_store/store.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused)] // TODO: remove this when refactoring is done
```
iGxnon, line 1: Remove?

crates/xline/src/server/auth_server.rs
themanforfree: This is just authentication, the command will be authorized when it enters the state machine, and if the auth info is invalid, it will be intercepted during authorization

crates/xline/src/server/auth_wrapper.rs
themanforfree: Same as above

crates/xline/src/storage/auth_store/store.rs
themanforfree: Removed

crates/curp/src/client/mod.rs
themanforfree: #657

crates/xline/src/server/auth_server.rs
iGxnon: If it should be invalid (e.g. expired when checked on the leader), but is recognized as valid on a machine with incorrect system time, should the JWT validation be separated from checking permissions?

themanforfree: Now it's separated, it's just JWT validation here, the permission check will be done in the `prepare` method of the state machine

iGxnon (approved):

Phoenix500526 (approved):
LGTM

PR #650, opened 2024-02-06
fix: fix timeout when append entries

iGxnon:
Signed-off-by: iGxnon <igxnon@gmail.com>

Fix the lock order

No

### DIFF
--- a/crates/curp/src/server/raw_curp/mod.rs
+++ b/crates/curp/src/server/raw_curp/mod.rs
@@ -101,6 +101,8 @@ pub struct RawCurp<C: Command, RC: RoleChange> {
     cst: Mutex<CandidateState<C>>,
     /// Curp logs
     log: RwLock<Log<C>>,
+    /// Contexts of fallback log entries
+    fallback_contexts: DashMap<LogIndex, FallbackContext<C>>,
     /// Relevant context
     ctx: Context<C, RC>,
     /// Task manager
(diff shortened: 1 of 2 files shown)


crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -101,6 +101,8 @@ pub struct RawCurp<C: Command, RC: RoleChange> {
     cst: Mutex<CandidateState<C>>,
     /// Curp logs
     log: RwLock<Log<C>>,
+    /// Contexts of fallback log entries
+    fallback_contexts: DashMap<LogIndex, FallbackContext<C>>,
```
bsbds, line 105: Please use a `RwLock<HashMap>` for this, as we don't need to concurrently modify the map.

crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -747,6 +755,15 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
             let mut log_w = RwLockUpgradableReadGuard::upgrade(log_r);
             if last_sent_index > log_w.commit_index {
                 log_w.commit_to(last_sent_index);
+                self.fallback_contexts.retain(|&idx, c| {
```
bsbds, line 758: Please move this to a seperate function.

bsbds:
Please add more details about why the lock order could cause a timeout.

Phoenix500526:
Close this PR since it doesn't provide enough information in the description. Additionally, there are no cases to demonstrate that this PR solves any problems.

PR #654, opened 2024-02-18
ci: add bench ci for pr

k7Xm9pQr:
Signed-off-by: iGxnon <igxnon@gmail.com>

Example: https://github.com/iGxnon/Xline/pull/2

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

    Add benchmark CI for each PR which labeled `run-benchmark`

* what changes does this pull request make?

    As above

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

    No

PR #658, opened 2024-02-22
fix:  fix xline member add bug

Phoenix500526:

1. close issue #661 

1. enrich the `cluster_validation` test case to cover the whole member add process.
3. sort the `self_peer_urls` to match the `member.peer_urls` and add client URLs 
4. add a single node cluster mode flag to specify a special case —— scale out from one node cluster to a two nodes clsuter.


crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -1659,12 +1659,20 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
             return false;
         }
 
-        let replicated_cnt = self
-            .lst
-            .iter()
-            .filter(|f| !f.is_learner && f.match_index >= i)
-            .count();
-        replicated_cnt + 1 >= quorum(self.ctx.cluster_info.voters_len())
+        let voter_len = self.ctx.cluster_info.voters_len();
```
bsbds, line 1662: It seems the `lst` and `ctx.cluster_info` both stored the voter infomations? I suggest to use `lst` on leader for consistency.

crates/curp/src/members.rs
```diff
@@ -138,26 +144,31 @@ impl ClusterInfo {
...
     ) -> Self {
         let mut member_id = 0;
-        let members = cluster
+        let sorted_self_addr = self_peer_urls.iter().cloned().sorted().collect::<Vec<_>>();
+        let members: DashMap<u64, Member> = cluster
             .members
             .into_iter()
             .map(|mut member| {
-                if member.peer_urls() == self_peer_urls {
+                if sorted_self_addr == member.peer_urls() {
```
rogercloud, line 156: If self_addr is sorted, does the `peer_urls`'s ordering matter while being compared?

scripts/common.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+SERVERS=("172.20.0.2" "172.20.0.3" "172.20.0.4" "172.20.0.5" "172.20.0.7")
+MEMBERS="node1=${SERVERS[1]}:2380,${SERVERS[1]}:2381,node2=${SERVERS[2]}:2380,${SERVERS[2]}:2381,node3=${SERVERS[3]}:2380,${SERVERS[3]}:2381"
+NEWMEMBERS="${MEMBERS},node4=${SERVERS[4]}:2380,${SERVERS[4]}:2381"
+
+XLINE_IMAGE="ghcr.io/xline-kv/xline:latest"
+ETCDCTL_IMAGE="ghcr.io/xline-kv/etcdctl:v3.5.9"
+
+function common::run_container() {
```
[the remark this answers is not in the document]
Phoenix500526: We use this naming convention in our xline operator E2E tests. We learn that from risingwave. FYI: https://github.com/risingwavelabs/risingwave-operator/blob/main/test/e2e/e2e.sh

scripts/validation_test.sh
```diff
@@ -37,6 +47,15 @@ function check() {
     fi
 }
 
+function check_without() {
```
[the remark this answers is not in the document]
Phoenix500526: I have already renamed it as `check_negative`.  The expression `check_negative X` means that the X should not appear in the result.

crates/curp/src/members.rs
Phoenix500526: Done

scripts/validation_test.sh
```diff
@@ -18,6 +19,15 @@ trap stop TERM
 
 res=""
 
+function run_new_member() {
+     docker_id=$(docker ps -qf "name=node4")
+    if [ -n "$docker_id" ]; then
+        docker stop $docker_id
```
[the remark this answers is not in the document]
Phoenix500526: In PR #654, we delete log files when the `stop_all` method is invoked. I think it's not a good practice for debugging. If the validation test failed, it will remove the log file too.

crates/curp/src/server/raw_curp/mod.rs
Phoenix500526: The single node mode is removed in that we don't fix this issue in 0.7.0.

rogercloud (approved):

rogercloud (dismissed):

crates/curp/src/members.rs
```diff
@@ -138,16 +138,26 @@ impl ClusterInfo {
     pub fn from_cluster(
         cluster: FetchClusterResponse,
         self_peer_urls: &[String],
+        self_client_urls: &[String],
         self_name: &str,
     ) -> Self {
         let mut member_id = 0;
+        let sorted_self_addr = self_peer_urls.iter().cloned().sorted().collect::<Vec<_>>();
```
bsbds, line 145: We can compare two iter directly without collecting them into Vec.

Phoenix500526: Done

bsbds (approved):

rogercloud (approved):

Issue #515, opened 2023-11-10
[Refactor] : Try Append Entries


themanforfree:
Currently, our `try_append_entries` method needs to Iterate through all new entries and do something for every entries, such as truncate, this is redundant and unnecessary.
The following is a better way
1. Find the index of the first entry that needs to be truncated.
2. Record entries that need to be fallback in the truncated entries.
3. truncate entries
4. Push the remaining entries and record the conf change entries

```rust
pub(super) fn try_append_entries(
	&mut self,
	entries: Vec<LogEntry<C>>,
	prev_log_index: LogIndex,
	prev_log_term: u64,
) -> Result<(ConfChangeEntries<C>, FallbackIndexes), Vec<LogEntry<C>>> {
	let mut conf_changes = vec![];
	let mut need_fallback_indexes = HashSet::new();
	// check if entries can be appended
	if self.get(prev_log_index).map_or_else(
		|| (self.base_index, self.base_term) != (prev_log_index, prev_log_term),
		|entry| entry.term != prev_log_term,
	) {
		return Err(entries);
	}
	// append log entries, will erase inconsistencies
	let mut li = prev_log_index;
	for entry in entries {
		let entry = Arc::new(entry);
		li += 1;
		let pi = self.li_to_pi(li);
		if self
			.entries
			.get(pi)
			.map_or(false, |old_entry| old_entry.term == entry.term)
		{
			continue;
		}
		for e in self.entries.range(pi..) {
			if matches!(e.entry_data, EntryData::ConfChange(_)) {
				let _ig = need_fallback_indexes.insert(e.index);
			}
		}
		if matches!(entry.entry_data, EntryData::ConfChange(_)) {
			conf_changes.push(Arc::clone(&entry));
		}
		self.entries.truncate(pi);
		#[allow(clippy::expect_used)] // It's safe to expect here.
		self.entries
			.push_back(Arc::clone(&entry))
			.expect("log entry {entry:?} cannot be serialized");
		self.send_persist(entry);
	}
	Ok((conf_changes, need_fallback_indexes))
}
```

caicancai:
> Currently, our method needs to Iterate through all new entries and do something for every entries, such as truncate, this is redundant and unnecessary. The following is a better way`try_append_entries`
> 
> 1. Find the index of the first entry that needs to be truncated.
> 2. Record entries that need to be fallback in the truncated entries.
> 3. truncate entries
> 4. Push the remaining entries and record the conf change entries
> 
> ```rust
[...]

PR #663, opened 2024-02-28
refactor: improve append entries

GG2002:

Optimize the behavior of log append entries. Ref to #515
The behavior of log append entries was redudant and this pr has made optimization according to following steps:
1. Find the index of the first entry that needs to be truncated.
2. Record entries that need to be fallback in the truncated entries.
3. truncate entries
4. Push the remaining entries and record the conf change entries
No.


Phoenix500526 (approved):
LGTM

iGxnon (approved):
LGTM

Issue #664, opened 2024-02-28
[Bug]: The xlinectl won't renew the lease of the lock key


Phoenix500526:
### Description about the bug

The `xlinectl` won't renew the lease of the lock key during the lock usage.

How do we reproduce the issue?
1. start up an xline cluster via quick_start.sh
2. execute a watch command to watch the lock key 
```bash
$ docker exec client /mnt/xlinectl --endpoints=http://172.20.0.3:2379,http://172.20.0.4:2379,http://172.20.0.5:2379 watch --prefix mutex
```
3. execute the lock command on another terminal  
```bash
$ docker exec client /mnt/xlinectl --endpoints=http://172.20.0.3:2379,http://172.20.0.4:2379,http://172.20.0.5:2379 lock mutex`
```
7. wait the lock lease expires.  

The result was pasted down below:
![image](https://github.com/xline-kv/Xline/assets/9337239/75f8f06e-77cc-4e02-8daa-3d1493f6099b)

### Version

0.6.1 (Default)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #666, opened 2024-02-29
fix: add keep alive logic for xlinectl lock command

Phoenix500526:
Closes: #664

fix issue #664 

1. remove the lease client from LockClient
2. add keep alive logic for xlinectl lock cmd

### DIFF
--- a/crates/xline-client/src/clients/lock.rs
+++ b/crates/xline-client/src/clients/lock.rs
@@ -19,11 +19,9 @@ use xlineapi::{
 };
 
 use crate::{
-    clients::{lease::LeaseClient, watch::WatchClient},
+    clients::watch::WatchClient,
     error::{Result, XlineClientError},
-    lease_gen::LeaseIdGenerator,
     types::{
-        lease::LeaseGrantRequest,
         lock::{LockRequest, UnlockRequest},
         watch::WatchRequest,
     },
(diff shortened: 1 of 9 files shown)


crates/xline-client/src/clients/lock.rs
```diff
@@ -84,6 +73,10 @@ impl LockClient {
     ///
     /// This function will return an error if the inner CURP client encountered a propose failure
     ///
+    /// # Panics
+    ///
+    /// Panic if the given `LockRequest.inner.lease` less than or equal to 0
+    ///
```
bsbds, line 79: Could it be better to return an error than panic?

crates/xlinectl/src/command/lock.rs
```diff
@@ -23,20 +26,32 @@ pub(crate) fn build_request(matches: &ArgMatches) -> LockRequest {
...
+    let lock_resp = client.lock_client().lock(req).await?;
 
-    let resp = client.lock_client().lock(req).await?;
+    lock_resp.print();
 
-    resp.print();
+    let (keeper, stream) = client
+        .lease_client()
+        .keep_alive(LeaseKeepAliveRequest::new(lock_lease_id))
+        .await?;
```
bsbds, line 42: We could  wrap these code in to a new "session" type in `xline-client`, instead of letting the user to keep alive the lock manually.
FYI: https://github.com/etcd-io/etcd/blob/main/client/v3/concurrency/session.go

Phoenix500526: I have the same idea. This should be a refactor, I'll leave an issue to track it. #684

crates/xline-client/src/clients/lock.rs
Phoenix500526: We don't need to panic or return an error when issue #684  is closed.

PR #656, opened 2024-02-20
feat: implement interval tree

bsbds:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

  Implements a interval tree, which is useful for dynamic interval update&query.

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #667, opened 2024-02-29
feat: Add bytes affine

bsbds:
Based on: #656 

  Adds a new type `BytesAffine`, which defines the order of the keys. 
  FYI: 
- https://etcd.io/docs/v3.3/learning/api/#key-ranges
- https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/pkg/adt/interval_tree.go#L919-L947

### DIFF
--- a/crates/xlineapi/src/interval.rs
+++ b/crates/xlineapi/src/interval.rs
@@ -1,0 +1,76 @@
+use std::cmp;
+
+use utils::interval_map::Interval;
+
+use crate::command::KeyRange;
+
+impl From<KeyRange> for Interval<BytesAffine> {
+    fn from(range: KeyRange) -> Self {
+        let start = range.range_start().to_vec();
+        let end = match range.range_end() {
+            &[] => {
+                let mut end = start.clone();
+                end.push(0);
+                BytesAffine::Bytes(end)
+            }
+            &[0] => BytesAffine::Unbounded,
+            bytes => BytesAffine::Bytes(bytes.to_vec()),
+        };
+        Interval::new(BytesAffine::Bytes(start), end)
+    }
+}
+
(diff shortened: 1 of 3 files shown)


crates/xlineapi/src/interval.rs
```diff
@@ -0,0 +1,76 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bytes_affine_cmp_is_ok() {
+        assert_eq!(BytesAffine::new_key("abc"), BytesAffine::new_key("abc"));
+        assert!(BytesAffine::new_key("a") < BytesAffine::new_key("b"));
+        assert!(BytesAffine::new_key("abcd") < BytesAffine::new_key("b"));
+        assert!(BytesAffine::new_key("abcd") < BytesAffine::new_unbounded());
+        assert!(BytesAffine::new_key("123") < BytesAffine::new_unbounded());
```
Phoenix500526, line 73: This line is duplicated. It doesn't provide more information with the previous one.

bsbds: removed

crates/xlineapi/src/interval.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bytes_affine_cmp_is_ok() {
```
Phoenix500526, line 68: Please cover the `KeyRange` from trait in this test case.

Phoenix500526 (approved):
LGTM，please squash these commits into one.

rogercloud (approved):

PR #673, opened 2024-03-05
refactor: engine refactor

bsbds:
Depends-On: #703 

Implements `StorageOps` for Engine

### DIFF
--- a/crates/engine/src/api/operation.rs
+++ b/crates/engine/src/api/operation.rs
@@ -1,6 +1,7 @@
 use crate::EngineError;
 
 /// Storage operations
+/// TODO: refactor this trait, require `&mut self` for write operations
 pub trait StorageOps {
     /// Write an op to the transaction
     ///
@@ -42,7 +43,7 @@ pub trait StorageOps {
 /// Write operation
 #[allow(clippy::module_name_repetitions)]
 #[non_exhaustive]
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum WriteOperation<'a> {
     /// `Put` operation
     Put {
--- a/crates/engine/src/memory_engine/mod.rs
+++ b/crates/engine/src/memory_engine/mod.rs
@@ -11,16 +11,17 @@ use std::{
 
 use bytes::{Bytes, BytesMut};
 use clippy_utilities::NumericCast;
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockWriteGuard};
 use tokio::io::AsyncWriteExt;
 use tokio_util::io::read_buf;
 
-pub(super) use self::transaction::MemoryTransaction;
 use crate::{
     api::{engine_api::StorageEngine, snapshot_api::SnapshotApi},
     error::EngineError,
-    WriteOperation,
+    StorageOps, WriteOperation,
 };
+
+pub(super) use self::transaction::MemoryTransaction;
 
 /// A helper type to store the key-value pairs for the `MemoryEngine`
 type MemoryTable = HashMap<Vec<u8>, Vec<u8>>;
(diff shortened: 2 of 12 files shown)


crates/engine/src/memory_engine/mod.rs
```diff
@@ -53,6 +54,45 @@ impl MemoryEngine {
...
+                    .ok_or_else(|| EngineError::TableNotFound(table.to_owned()))?;
+                table.retain(|key, _value| {
+                    let key_slice = key.as_slice();
+                    match key_slice.cmp(from) {
+                        Ordering::Less => true,
+                        Ordering::Equal => false,
+                        Ordering::Greater => match key_slice.cmp(to) {
+                            Ordering::Less => false,
+                            Ordering::Equal | Ordering::Greater => true,
+                        },
```
Phoenix500526, line 89: nit. Better use `matches!(key_slice.cmp(to), Ordering::Equal | Ordering::Greater)` to replace.

crates/engine/src/mock_rocksdb_engine.rs
```diff
@@ -9,12 +8,12 @@ use bytes::{Bytes, BytesMut};
...
     error::EngineError,
-    memory_engine::{MemoryEngine, MemorySnapshot},
-    TransactionApi, WriteOperation,
+    memory_engine::{MemoryEngine, MemorySnapshot, MemoryTransaction},
+    StorageOps, TransactionApi, WriteOperation,
 };
 
 /// Mock `RocksDB` Storage Engine
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
Phoenix500526, line 16: Same as above. Why shoule a `RocksEngine` be cloneable?

crates/engine/src/mock_rocksdb_engine.rs
```diff
@@ -206,20 +212,48 @@ impl SnapshotApi for RocksSnapshot {
     }
 }
 
-/// A transaction of the `RocksEngine`
-#[derive(Copy, Clone, Debug, Default)]
+/// Mock `RocksTransaction`
```
Phoenix500526, line 215: Shall we rename this `RocksTransaction` as `MockRocksTransaction`? Is it better?

crates/engine/src/api/operation.rs
```diff
@@ -42,7 +43,7 @@ pub trait StorageOps {
 /// Write operation
 #[allow(clippy::module_name_repetitions)]
 #[non_exhaustive]
-#[derive(Debug)]
```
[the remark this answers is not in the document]
bsbds: The current implementation will write the operation in a loop and needs to be cloned every time. The loop will be removed during the following refactor.

crates/engine/src/mock_rocksdb_engine.rs
bsbds: It's a mock implementation, and it's a wrapper around a `MemoryEngine`, which is clonable.

crates/engine/src/mock_rocksdb_engine.rs
bsbds: It's meant to replace the normal `RocksTransaction` if the madsim compile flag is set, they should share the same name.

crates/engine/src/memory_engine/mod.rs
bsbds: done

crates/engine/src/api/operation.rs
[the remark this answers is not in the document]
Phoenix500526: Shall we leave something, like an issue or comment, to track?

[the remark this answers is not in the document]
bsbds: There's a todo comment to track this: https://github.com/xline-kv/Xline/blob/159d74cf02cc63c611788dca2f2b47e8a9f5cbe6/crates/engine/src/rocksdb_engine/mod.rs#L153

Phoenix500526 (dismissed):
LGTM, please squash these commits into one.

Phoenix500526 (approved):
LGTM

Issue #677, opened 2024-03-07
[Bug]: Xline will loss event when using watch feature


Phoenix500526:
### Description about the bug

When I ported Xline to Kubernetes, I encountered timeouts during kubeadm init. By comparing the behavior of etcd and Xline, I discovered that Xline's watch feature may experience event loss under certain conditions. The root cause of the event loss is documented in issue #491.

### Version

0.6.1 (Default)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #403, opened 2023-07-28
[WIP] Add e2e ci for kubernetes with kind.

liangyuanpeng:
### what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

Make sure xline can work with kubernetes, so add e2e ci for kubernetes.

### what changes does this pull request make?

 add e2e ci for kubernetes.

---
Now, it's sometimes fail, sometimes success.

successed: https://github.com/liangyuanpeng/Xline/actions/runs/5688707901
failed:  https://github.com/liangyuanpeng/Xline/actions/runs/5688679640/job/15418920092

PR #678, opened 2024-03-07
fix: xline watcher will loss event

Phoenix500526:
Closes: #677

Add a HashSet to track all notified revision. 

### DIFF
--- a/crates/xline/src/storage/kvwatcher.rs
+++ b/crates/xline/src/storage/kvwatcher.rs
@@ -8,15 +8,13 @@ use std::{
     time::Duration,
 };
 
-use clippy_utilities::OverflowArithmetic;
 use itertools::Itertools;
-use log::warn;
 use parking_lot::RwLock;
 use tokio::{
     sync::mpsc::{self, error::TrySendError},
     time::sleep,
 };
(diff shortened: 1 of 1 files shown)


crates/xline/src/storage/kvwatcher.rs
```diff
@@ -136,26 +138,29 @@ impl Watcher {
...
         };
 
         match self.event_tx.try_send(watch_event) {
             Ok(_) => {
-                debug!(watch_id, revision, "response sent successfully");
-                self.start_rev = revision.overflow_add(1);
+                assert!(
+                    self.notified_set.insert(revision),
+                    "should not notify the same revision twice!!"
+                );
```
bsbds, line 154: The revision is checked in L143, so the assert is not needed here.

Phoenix500526: Done

bsbds (approved):

liangyuanpeng:
Have tested this PR with https://github.com/xline-kv/Xline/pull/403 (single kubernetes cluster and single xline ) and always success.

rogercloud (approved):

PR #669, opened 2024-03-04
Feat: Instrument await using await-tree

Harsh1s:
Signed-off-by: Harsh Raj <thisishraj@gmail.com>

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
-> FIxes #580 
-> Currently, some simple information is logged to display the execution path of the command. This uses more detailed information to show this process (i.e. use await-tree to indicate the duration of each await.).

* what changes does this pull request make?
-> Makes use of `instrument_await` within functions annotated with the #[instrument] macro to add more detailed async tracing.

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)
-> No

PR #682, opened 2024-03-08
feat: Instrument await using await-tree

Harsh1s:
- what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)
-> FIxes #669 
-> Currently, some simple information is logged to display the execution path of the command. This uses more detailed information to show this process (i.e. use await-tree to indicate the duration of each await.).

- what changes does this pull request make?
-> Makes use of instrument_await within functions annotated with the #[instrument] macro to add more detailed async tracing.

- are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)
-> No

### DIFF
--- a/crates/curp/Cargo.toml
+++ b/crates/curp/Cargo.toml
@@ -14,6 +14,7 @@ version = "0.1.0"
 [dependencies]
 async-stream = "0.3.4"
 async-trait = "0.1.53"
+await-tree = "0.1.2"
 bincode = "1.3.3"
 bytes = "1.4.0"
 clippy-utilities = "0.2.0"
--- a/crates/xline/Cargo.toml
+++ b/crates/xline/Cargo.toml
@@ -15,6 +15,7 @@ categories = ["KV"]
 anyhow = "1.0.57"
 async-stream = "0.3.5"
 async-trait = "0.1.53"
+await-tree = "0.1.2"
 axum = "0.6.20"
 bytes = "1.4.0"
 clap = { version = "4", features = ["derive"] }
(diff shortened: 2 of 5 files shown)


Phoenix500526 (commented):
Will this modification affect performance in any way? I suggest adding a switch here so that we can enable it as needed.

iGxnon:
Can you provide an example after running it?

GFX9 (approved):

Phoenix500526:
Hi, @Harsh1s ! This pr has been stalled for 3 weeks. Would you like to update it? 😄

Harsh1s:
> Hi, @Harsh1s ! This pr has been stalled for 3 weeks. Would you like to update it? 😄

Oh I'm really sorry about the stalling, it's been a busy past month at my uni, a lot of tests and project deadlines. My end semester exams are going on currently too. I'll try to update it within a week for sure. Sorry again!

crates/curp/Cargo.toml
```diff
@@ -14,6 +14,7 @@ version = "0.1.0"
 [dependencies]
 async-stream = "0.3.4"
 async-trait = "0.1.80"
+await-tree = "0.1.2"
```
rogercloud, line 17: The latest version is "0.2.1", would you please update it?

crates/xline/Cargo.toml
```diff
@@ -15,6 +15,7 @@ categories = ["KV"]
 anyhow = "1.0.82"
 async-stream = "0.3.5"
 async-trait = "0.1.80"
+await-tree = "0.1.2"
```
rogercloud, line 18: The same as above.

Issue #516, opened 2023-11-10
[Refactor]: Xline logs


bsbds:
Currently there are serveral issues in the log of Xline:

1. Log size is huge
During a recent Jepsen test with Xline `debug` level enabled, Xline printed the key-values of every command to logs. This resulted in large log files, as the values can be several tens of kilobytes in size. To address this issue, the logs should be refactored to either use the `tracing` level or disable the printing of key-values. Note that this issue may also be present in snapshot related logs.

2. Logs are confusing
The current logs are confusing as they lack essential information required for effective debugging. The logs should include key information, such as propose ID and revisions, to enable easy tracking of command execution.

Phoenix500526:
Refer to https://github.com/xline-kv/Xline/pull/683

rohansx:
@bsbds would like to take on this task

Phoenix500526:
Hi, @rohansx , would you like to take on this issue?

rohansx:
@Phoenix500526 Yeah sure i can give it a try

Phoenix500526:
Hi，@rohansx, I'm glad you're interested in this issue. However, I believe there are some ambiguous points in the issue description, such as how to define the so-called "essential information" in point 2. This may require a deep understanding of the entire project and experience in debugging. IMO, starting with the first point would be relatively easier as a starting point.  @bsbds can you give some information for the point two?

PR #683, opened 2024-03-08
Refactor/better debug

Phoenix500526:
Currently, some gRPC requests and responses consist of bytes and are not human-readable. I implemented the Display trait for these requests and responses so they can be output in a more developer-friendly format. The reason for implementing Display instead of Debug is that Prost automatically derives Debug for all requests and responses. This pr can reduce the log content size which mentioned in issue #516 .

Implement Display trait for some requests and responses

no

### DIFF
--- a/crates/xline/src/server/auth_server.rs
+++ b/crates/xline/src/server/auth_server.rs
@@ -142,8 +142,8 @@ where
         &self,
         mut request: tonic::Request<AuthUserAddRequest>,
     ) -> Result<tonic::Response<AuthUserAddResponse>, tonic::Status> {
-        debug!("Receive AuthUserAddRequest {:?}", request);
         let user_add_req = request.get_mut();
+        debug!("Receive AuthUserAddRequest {}", user_add_req);
         user_add_req.validation()?;
         let hashed_password = Self::hash_password(user_add_req.password.as_bytes());
         user_add_req.hashed_password = hashed_password;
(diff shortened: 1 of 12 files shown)


bsbds (approved):

crates/xline/src/server/auth_server.rs
```diff
@@ -142,8 +142,8 @@ where
         &self,
         mut request: tonic::Request<AuthUserAddRequest>,
     ) -> Result<tonic::Response<AuthUserAddResponse>, tonic::Status> {
-        debug!("Receive AuthUserAddRequest {:?}", request);
         let user_add_req = request.get_mut();
+        debug!("Receive AuthUserAddRequest {}", user_add_req);
```
rogercloud, line 146: It's in the `debug` tracing call. Why do you use display instead of debugging?

rogercloud: Try to override this trait https://doc.rust-lang.org/std/fmt/trait.Debug.html

Phoenix500526: The prost will automatically derive the debug trait for all the requests and responses it generates.  So far, we cannot switch off this behavior for the prost crate. FYI: https://github.com/tokio-rs/prost/issues/334. So I have to override the Display trait for it.

bsbds (approved):

crates/xline/src/storage/kvwatcher.rs
```diff
@@ -552,6 +552,19 @@ pub(crate) struct WatchEvent {
     compacted: bool,
 }
 
+impl std::fmt::Debug for WatchEvent {
+    #[allow(clippy::arithmetic_side_effects)]
```
rogercloud, line 556: If it's affected by the macro, please fix it in the macro.

Phoenix500526: Done

rogercloud (approved):

bsbds (approved):

PR #696, opened 2024-03-16
Add github action to run e2e test with kubernetes cluster.

liangyuanpeng:

Add github action to run e2e test with kubernetes cluster.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -109,3 +109,95 @@ jobs:
       - uses: actions/checkout@v4
       - name: Check Spelling
         uses: crate-ci/typos@master
+
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Configure sccache
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+
+      - name: Build xline image
+        run: |
+          docker run -q --rm -v $(pwd):/xline \
+            -e SCCACHE_GHA_ENABLED=on \
+            -e ACTIONS_CACHE_URL=${ACTIONS_CACHE_URL} \
+            -e ACTIONS_RUNTIME_TOKEN=${ACTIONS_RUNTIME_TOKEN} \
+            ghcr.io/xline-kv/build-env:latest \
+            cargo build --release --bin xline --bin benchmark --bin validation_lock_client
+
+          sudo apt-get install -y --force-yes expect
+          cd scripts
+          cp ../target/release/{xline,benchmark,validation_lock_client} .
+          ldd ./xline
+          ldd ./benchmark
+          cp ../fixtures/{private,public}.pem .
+          docker build . -t ghcr.io/xline-kv/xline:master
+
+          cd ..
+          pwd
+          mkdir -p _output
+          docker save -o _output/xline.tar ghcr.io/xline-kv/xline:master
+          ls _output
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: xline
+          path: _output/xline.tar
+
+  e2e:
+    name: e2e_kube
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s:
+          - v1.27.3
+          - v1.28.7
+          - v1.29.2
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v3
+        with:
+          name: xline
+
+      - name: build kind node image with xline
+        run: |
+          docker build . -f ci/artifacts/kind-node-xline.Dockerfile -t ghcr.io/xline-kv/kind-node-xline:${{ matrix.k8s }}-master --build-arg K8S_VERSION=${{ matrix.k8s }}
+
+      - name: Create kubernetes cluster with kind
+        env:
+          K8S_VERSION: ${{ matrix.k8s }}
+        run: |
+          ./ci/scripts/kind.sh
+
+      - name: Get cluster status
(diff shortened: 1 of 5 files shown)


.github/workflows/pull_request.yml
```diff
@@ -109,3 +109,36 @@ jobs:
...
+      - name: Export logs
+        if: always()
+        run: |
+          mkdir -p _artifacts
+          docker ps
+          kind export logs  --loglevel=debug ./_artifacts/logs
+          ls _artifacts
+
+      - name: Upload logs
+        if: always()
```
liangyuanpeng, line 146: > https://github.com/xline-kv/Xline/pull/403#discussion_r1525719154
> I think uploading detailed logs when deployment failed is a better choice.

@Phoenix500526 

I think this would still be of value, at least proving that the log collection functionality works, but if you still prefer to have it only work on failure, I'd be happy to update it.

ci/artifacts/xline-pod.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  creationTimestamp: null
+  name: xline
+  namespace: kube-system
+spec:
+  containers:
+    - name: xline
+      image: ghcr.io/liangyuanpeng/xline:ffa824f
```
liangyuanpeng, line 10: Should i follow the doc https://github.com/xline-kv/Xline/tree/master/doc/quick-start#build-image to build xline image here?

Maybe there are other CIs that have tested building images, but I didn't find any.

cc @bsbds @Phoenix500526

ci/artifacts/xline-pod.yaml
```diff
@@ -0,0 +1,17 @@
...
+      imagePullPolicy: IfNotPresent
+      command:
+        - /bin/sh
+        - -c
+        - xline --name node1 --members node1=kind-control-plane:2379 --data-dir /tmp/xline --storage-engine rocksdb --client-listen-urls=http://kind-control-plane:2379 --peer-listen-urls=http://kind-control-plane:2380,http://kind-control-plane:2381 --client-advertise-urls=http://kind-control-plane:2379 --peer-advertise-urls=http://kind-control-plane:2380,http://kind-control-plane:2381
```
liangyuanpeng, line 15: This PR will focus on single xline nodes and can be improved to support multiple xline nodes in the future.

ci/artifacts/xline-pod.yaml
liangyuanpeng: https://github.com/xline-kv/Xline/blob/75e51f4fcbb23b75d3b0f31ea69ed40684694f29/.github/workflows/validation.yml#L42

It looks like the image is built here.

Phoenix500526: Yes, it is. We do the same thing in our benchmark:
https://github.com/xline-kv/Xline/blob/75e51f4fcbb23b75d3b0f31ea69ed40684694f29/.github/workflows/benchmark.yml#L36C1-L41C30

.github/workflows/pull_request.yml
Phoenix500526: I prefer to have it on failure. In general, we only check the logs if a task fails. So why upload logs when it succeeds? Additionally, the success of log uploads has no bearing on Xline itself and doesn't require repeated verification.

.github/workflows/pull_request.yml
```diff
@@ -109,3 +109,42 @@ jobs:
...
+          - v1.29.2
+    steps:
+      - uses: actions/checkout@v4
+      - name: Create kubernetes cluster with kind
+        env:
+          K8S_VERSION: ${{ matrix.k8s }}
+        run: |
+          ./ci/scripts/kind.sh
+
+      - name: Get cluster status
```
Phoenix500526, line 131: Please remove the trailing space on this line.

.github/workflows/pull_request.yml
liangyuanpeng: Updated.

.github/workflows/pull_request.yml
```diff
@@ -109,3 +109,94 @@ jobs:
...
+          cd .. 
+          pwd
+          mkdir -p _output
+          docker save -o _output/xline.tar ghcr.io/xline-kv/xline:master
+          ls _output
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: xline
+          path: _output/xline.tar
```
liangyuanpeng, line 154: Obviously this can be refactored into a reusable workflow, But it will not be processed in this PR.

ci/artifacts/xline-pod.yaml
liangyuanpeng: Thanks for your quickly confirmation.

.github/workflows/pull_request.yml
Phoenix500526: Could you please to initiate an issue for that?

liangyuanpeng: Sure, created https://github.com/xline-kv/Xline/issues/700 to trace it.

Phoenix500526 (approved):
LGTM

.github/workflows/pull_request.yml
```diff
@@ -109,3 +109,95 @@ jobs:
...
+          docker run -q --rm -v $(pwd):/xline \
+            -e SCCACHE_GHA_ENABLED=on \
+            -e ACTIONS_CACHE_URL=${ACTIONS_CACHE_URL} \
+            -e ACTIONS_RUNTIME_TOKEN=${ACTIONS_RUNTIME_TOKEN} \
+            ghcr.io/xline-kv/build-env:latest \
+            cargo build --release --bin xline --bin benchmark --bin validation_lock_client
+
+          sudo apt-get install -y --force-yes expect
+          cd scripts
+          cp ../target/release/{xline,benchmark,validation_lock_client} .
```
bsbds, line 139: It seems the `validation_lock_client` does not need to be built.

liangyuanpeng: As the https://github.com/xline-kv/Xline/pull/696#discussion_r1527402223 say, just leave this to https://github.com/xline-kv/Xline/issues/700 ,  WDYT?

liangyuanpeng:
cc @iGxnon for review,you may be interested in this.

bsbds (approved):

PR #726, opened 2024-03-23
chore: ignore hyper dependency update

Phoenix500526:
Prevent automatic upgrade of hyper from 0.14.28 to 1.0.0, as tonic depends on deprecated data structures, such as `AddrStream` and `AddrIncoming`.
FYI: https://github.com/hyperium/hyper/issues/2850

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,8 @@ updates:
   labels:
     - "dependencies"
   open-pull-requests-limit: 10
+  ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,3 +16,7 @@ updates:
   labels:
     - "dependencies"
   open-pull-requests-limit: 10
+  ignore:
```
rogercloud, line 19: Please add a comment containing `FIXME` to explain the the reason, and when to remove the ignorance.

Phoenix500526: Done

PR #745, opened 2024-03-30
chore: update xline pod of ci,add probe,volume and security config.

liangyuanpeng:

The content is basically copied from etcd pod yaml,happy to update it if we have better config.


ci/artifacts/xline-pod.yaml
```diff
@@ -6,12 +6,52 @@ metadata:
...
+        port: 9100
+        scheme: HTTP
+      initialDelaySeconds: 10
+      periodSeconds: 10
+      timeoutSeconds: 15
+    startupProbe:
+      failureThreshold: 24
+      httpGet:
+        host: 127.0.0.1
+        path: /metrics
```
liangyuanpeng, line 26: This probably requires a better one, but I haven't found one. Remind me if i missed something,Thanks.

liangyuanpeng:
cc @Phoenix500526 @bsbds 
PTAL,Thanks.

ci/artifacts/xline-pod.yaml
liangyuanpeng: Basically, this requires the application to have `/readyz` and `/livez` interfaces. etcd is also implement the grpc for it https://github.com/etcd-io/etcd/commit/293f087fb987421348dd9a4173087bb5f08cb850#diff-ab6fb0684315e16355f6ebe0f4b3cf860b9b2ff5a0fe1b4e4308a680b19f1b0c

liangyuanpeng:
kindly ping :)

PR #771, opened 2024-04-09
ci: update .mergeify bot

Phoenix500526:
update modify bot to track e2e-kube status

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -155,7 +155,7 @@ jobs:
           path: _output/xline.tar
 
   e2e:
-    name: e2e_kube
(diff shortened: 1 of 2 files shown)


liangyuanpeng:
@Phoenix500526  this is LGTM and the reason of close it is this test is not required?

Phoenix500526:
> @Phoenix500526 this is LGTM and the reason of close it is this test is not required?

It's necessary, but we may not do it this way. Currently, the mergify bot requires explicit status checks; otherwise, it will ignore those that have failed but not checked statuses, potentially merging the pr into the master branch. Therefore a better way is to check all statues unless explicitly skipped in the .mergify.yml file.

Issue #774, opened 2024-04-10
[Bug]: The test case curp::it server::shutdown_rpc_should_shutdown_the_cluster failed


Phoenix500526:
### Description about the bug

The test case `curp::it server::shutdown_rpc_should_shutdown_the_cluster` failed sometimes. The output log has been pasted below.

FYI: https://github.com/xline-kv/Xline/actions/runs/8626722031/job/23645355082?pr=772

### Version

0.6.1 (Default)

### Relevant log output

```shell
SIGABRT [   8.488s] curp::it server::shutdown_rpc_should_shutdown_the_cluster

--- STDOUT:              curp::it server::shutdown_rpc_should_shutdown_the_cluster ---

running 1 test
   0.049279817s DEBUG curp::server::curp_node: 8280384654968381684 to 13740163161196736920 sync follower task start
   0.049290450s DEBUG curp::server::curp_node: 8280384654968381684 to 11072999871023124112 sync follower task start
   0.090303489s DEBUG curp::server::raw_curp: 11072999871023124112 becomes the leader
   0.090505576s DEBUG curp::server::curp_node: 11072999871023124112 to 8280384654968381684 sync follower task start
   0.090521933s DEBUG curp::server::curp_node: 11072999871023124112 to 13740163161196736920 sync follower task start
   0.093805270s DEBUG curp_append_entries: curp::server::raw_curp: 8280384654968381684 updates to term 1 and becomes a follower
   0.143433580s  WARN curp::server::curp_node: ae to 13740163161196736920 failed, Internal("status: Cancelled, message: \"Timeout expired\", details: [], metadata: MetadataMap { headers: {} }")
   0.159462727s DEBUG curp::server::curp_node: 13740163161196736920 to 8280384654968381684 sync follower task start
   0.159521547s DEBUG curp::server::curp_node: 13740163161196736920 to 11072999871023124112 sync follower task start
   0.393044405s DEBUG curp_append_entries: curp::server::raw_curp: 13740163161196736920 updates to term 1 and becomes a follower
   0.465063805s DEBUG curp::rpc::connect: client request a client id
   0.466749142s DEBUG curp_propose: curp::server::raw_curp: 13740163161196736920 gets proposal for cmd(0#8342557537241818130)
   0.466812601s DEBUG curp_propose: curp::server::spec_pool: insert cmd(0#8342557537241818130) into spec pool
[...]

PR #776, opened 2024-04-12
fix: add txn retry logic

Phoenix500526:
Refs: #774
Closes: #774

FYI: #774 

Add a retry logic in `write_batch`.

There is nothing to change breakingly, but we need to do a performance regression test for this modification.

### DIFF
--- a/crates/curp/tests/it/server.rs
+++ b/crates/curp/tests/it/server.rs
@@ -297,7 +297,7 @@ async fn shutdown_rpc_should_shutdown_the_cluster() {
     ));
 
     let collection = collection_task.await.unwrap();
-    sleep_secs(7).await; // wait for the cluster to shutdown
+    sleep_secs(1).await; // wait for the cluster to shutdown
     assert!(group.is_finished());
 
     let group = CurpGroup::new_rocks(3, tmp_path).await;
--- a/crates/engine/src/rocksdb_engine/mod.rs
+++ b/crates/engine/src/rocksdb_engine/mod.rs
@@ -14,11 +14,12 @@ use std::{
 use bytes::{Buf, Bytes, BytesMut};
 use clippy_utilities::{NumericCast, OverflowArithmetic};
 use rocksdb::{
-    Direction, Error as RocksError, IteratorMode, OptimisticTransactionDB, Options, SstFileWriter,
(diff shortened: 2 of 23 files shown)


crates/curp/tests/it/server.rs
```diff
@@ -297,7 +297,7 @@ async fn shutdown_rpc_should_shutdown_the_cluster() {
     ));
 
     let collection = collection_task.await.unwrap();
-    sleep_secs(7).await; // wait for the cluster to shutdown
+    sleep_secs(1).await; // wait for the cluster to shutdown
```
bsbds, line 300: Why change this sleep duration to 1s?

crates/engine/src/rocksdb_engine/mod.rs
```diff
@@ -200,51 +201,71 @@ impl StorageEngine for RocksEngine {
...
-                        .collect::<Result<Vec<_>, _>>()?;
-                    for (key, _) in kvs {
-                        transaction.delete_cf(&cf, key)?;
+                Err(err) if matches!(err.kind(), RocksErrorKind::Busy | RocksErrorKind::TryAgain) => {
+                    if retry_count > max_retry_count {
+                        warn!("Oops, txn commit retry count reach the max_retry_count: {max_retry_count}");
+                        return Err(EngineError::UnderlyingError(err.to_string()));
                     }
+                    warn!("Rocksdb txn commit failed, retrying after {retry_interval}ms");
+                    std::thread::sleep(std::time::Duration::from_millis(retry_interval));
```
bsbds, line 263: Using a thread::sleep will block a tokio worker thread, I suggest to retry immediately as the conflict situation is rare.

crates/curp/tests/it/server.rs
Phoenix500526: Because it's a random bug, I must run it several times to reproduce.  The 7 sec is too long. Do we need to sleep for such a long duration?

crates/engine/src/rocksdb_engine/mod.rs
Phoenix500526: Retrying immediately when a conflict occurs does nothing to help; instead, it aggravates the conflict.

bsbds (approved):

crates/curp/tests/it/server.rs
rogercloud: Is there another way to tell whether group is finished or not?

rogercloud (approved):

PR #785, opened 2024-04-19
Pr/stale bot

Phoenix500526:
add a workflow to handle stale issues and prs

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,23 @@
+name: Mark stale pull requests
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v9
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        start-date: '2024-04-19T00:00:00Z' # ISO 8601 or RFC 2822
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 14 days.'
+        stale-pr-message: 'This PR is stale because it has been open for 30 days with no activity. Remove stale label or comment or this will be closed in 14 days'
+        close-issue-message: 'This issue was closed because it has been stalled for 14 days with no activity.'
+        close-pr-message: 'This PR was closed because it has been stalled for 14 days with no activity.'
+        stale-pr-label: 'stale'
+        exempt-issue-labels: 'good-first-issue'
+        exempt-all-milestones: true
+        days-before-stale: 30
+        days-before-close: 14
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -21,3 +21,11 @@ jobs:
         exempt-all-milestones: true
         days-before-stale: 30
         days-before-close: 14
+
+  unassign:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Auto unassign stale assignments
+      uses: JasonEtco/slash-assign-action@v1
+      with:
+        required_label: 'good-first-issue'
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,21 +1,23 @@
 pull_request_rules:
   - name: Automatic merge on approval
     conditions:
-      - check-success="Tests Normal"
-      - check-success="Tests Madsim"
-      - check-success="Commit Message Validation"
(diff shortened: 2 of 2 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+        days-before-close: 14
+
+  unassign:
+    if: github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Auto unassign stale assignments
+      uses: JasonEtco/slash-assign-action@v1
+      with:
+        required_label: 'good-first-issue'
```
liangyuanpeng, line 32: It's mean every day will unassign for `good-first-issue`? and not `good-first-issue` & `stale`? could you like to explain here?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+        close-issue-message: 'This issue was closed because it has been stalled for 14 days with no activity.'
+        close-pr-message: 'This PR was closed because it has been stalled for 14 days with no activity.'
+        stale-pr-label: 'stale'
+        exempt-issue-labels: 'good-first-issue'
+        exempt-all-milestones: true
+        days-before-stale: 30
+        days-before-close: 14
+
+  unassign:
+    if: github.event_name == 'schedule'
```
liangyuanpeng, line 26: this workflow is only trigger by schedule, so this is unnecessary?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,32 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        start-date: '2024-04-19T00:00:00Z' # ISO 8601 or RFC 2822
+        stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 14 days.'
+        stale-pr-message: 'This PR is stale because it has been open for 30 days with no activity. Remove stale label or comment or this will be closed in 14 days'
+        close-issue-message: 'This issue was closed because it has been stalled for 14 days with no activity.'
+        close-pr-message: 'This PR was closed because it has been stalled for 14 days with no activity.'
+        stale-pr-label: 'stale'
+        exempt-issue-labels: 'good-first-issue'
+        exempt-all-milestones: true
```
liangyuanpeng, line 21: It's an inactive issue but it could still be in milestones, right?

.mergify.yml
```diff
@@ -64,13 +70,15 @@ pull_request_rules:
 
   - name: comment when a pull request cannot be merged
     conditions:
+      - -draft
```
liangyuanpeng, line 73: cool, so there will be no duplicate comments, right?

.github/workflows/stale.yml
Phoenix500526: Yes, because we usually use issues in milestones to track our release plan. This means that these issues will closed when the relevant version is released. These issues should not be counted as stale.

.mergify.yml
Phoenix500526: That means it won't approve any WIP pr.

.github/workflows/stale.yml
Phoenix500526: It means that this workflow will only search those issues with good-first-issue labels.

.github/workflows/stale.yml
Phoenix500526: Done

Phoenix500526:
@Mergifyio rebase

.github/workflows/stale.yml
liangyuanpeng: Yes, so it's  with `JasonEtco/slash-assign-action` for stable assign,  different with `actions/stale`. found the answer from https://github.com/JasonEtco/slash-assign-action?tab=readme-ov-file#days_until_unassign

.mergify.yml
liangyuanpeng: want to know  it will fix the duplicate comment,right, see  
image

Phoenix500526: But they are two different comments.

liangyuanpeng: I just send the wrong screen (:

it's 
02b8f342f9ff48a1e63a6418ff66d03


after talking with @Phoenix500526 , it will fix it.

liangyuanpeng (dismissed):

bsbds (approved):

liangyuanpeng (approved):

PR #786, opened 2024-04-19
refactor: remove sleep_secs method when waiting the cluster to shutdown

Phoenix500526:
Using a notify logic to instead the sleep logic in some integration test cases

### DIFF
--- a/crates/curp/tests/it/common/curp_group.rs
+++ b/crates/curp/tests/it/common/curp_group.rs
@@ -29,6 +29,7 @@ use tokio::{
     runtime::{Handle, Runtime},
     sync::{mpsc, watch},
     task::{block_in_place, JoinHandle},
+    time::timeout,
 };
 use tokio_stream::wrappers::TcpListenerStream;
 use tonic::transport::{Certificate, Channel, ClientTlsConfig, Endpoint, ServerTlsConfig};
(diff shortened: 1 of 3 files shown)


crates/curp/tests/it/server.rs
```diff
@@ -297,8 +297,7 @@ async fn shutdown_rpc_should_shutdown_the_cluster() {
     ));
 
     let collection = collection_task.await.unwrap();
-    sleep_secs(1).await; // wait for the cluster to shutdown
-    assert!(group.is_finished());
+    group.wait_for_group_shutdown(Duration::from_secs(7)).await;
```
rogercloud, line 300: As all the timeout is the same, shall we provide a default value for it or just wrap the value in the default function?

crates/curp/tests/it/common/curp_group.rs
```diff
@@ -332,6 +333,49 @@ impl CurpGroup {
...
+        .await
+        .expect("wait for group to shutdown timeout");
+        assert!(self.is_finished(), "The group is not finished yet");
+    }
+
+    async fn wait_for_targets_shutdown(targets: impl Iterator<Item = &CurpNode>) {
+        let final_tasks: [TaskName; 3] = [
+            TaskName::WatchTask,
+            TaskName::ConfChange,
+            TaskName::LogPersist,
```
rogercloud, line 365: These are bottom tasks, which are waited by all the other tasks. I recommend to put them into a group so that we won't miss any of them even we refactor the task dependency topo later.

crates/curp/tests/it/server.rs
Phoenix500526: Done

crates/curp/tests/it/common/curp_group.rs
Phoenix500526: Done

crates/utils/src/task_manager/tasks.rs
```diff
@@ -6,6 +6,11 @@
 //                    \        /      |      \       /
 //                   WATCH_TASK  CONF_CHANGE  LOG_PERSIST
 
+// In integration tests, we use the bottom tasks, like `WatchTask`, `ConfChange`, and `LogPersist`,
```
rogercloud, line 9: Add a "NOTE:" prefix to highlight this comment.

rogercloud (approved):

crates/utils/src/task_manager/tasks.rs
Phoenix500526: Done

rogercloud (approved):

bsbds (approved):

Issue #823, opened 2024-05-16
[Feature]: Add exec command args for xlinectl lock command


Phoenix500526:
### Description about the feature

The etcdctl provides [the exec-command arg](https://github.com/etcd-io/etcd/tree/main/etcdctl#lock-options-lockname-command-arg1-arg2-) for the lock command.  Therefore,  when you want to execute some commands after acquiring a lock, you can use one-line command, like `./etcdctl lock mylock ./etcdctl put foo bar`, instead of several shell commands, like `./etcdctl lock mylock; ./etcdctl put foo bar`. I think this way is more user-friendly. 

So far, the xlinectl doesn't support the exec command args for the lock command. I think it's not hard to implement. I leave an issue here to track. Please feel free to take this issue if you are interested in it. And please let me know if you need any help or more information.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bhavik-goplani:
Working on it. Need to incorporate new changes to the lock command

PR #846, opened 2024-06-10
feat: add exec command args for xlinectl lock command

bhavik-goplani:

https://github.com/xline-kv/Xline/issues/823

Added exec command support to xlinectl lock command. Updated the README with new examples for lock.

No

Additional comments:
I'm not sure how to test the exec command args. I thought of two solutions:
1. Extend LockRequest or Test Case Representation: Add a way to represent the exec commands that should be executed after the lock is acquired. This could be a new field in the LockRequest struct or a separate structure linked to the test case.
2. Modify the Test Case: Adjust the test case in macros.rs to include the exec commands as part of the expected outcome.

### DIFF
--- a/crates/xlinectl/README.md
+++ b/crates/xlinectl/README.md
@@ -652,14 +652,25 @@ Acquire a lock, which will return a unique key that exists so long as the lock i
 #### Usage
 
 ```bash
-lock <lockname>
+lock <lockname> [command arg1 arg2 ...]
 ```
 
 #### Examples
 
 ```bash
-# Hold a lock named foo until `SIGINT` is received
+# Hold a lock named `foo` until `SIGINT` is received
 ./xlinectl lock foo
+foo/key
+
+# Hold a lock named `mylock` until `SIGINT` is received and print lock acquired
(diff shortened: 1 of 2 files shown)


CrystalAnalyst:
Sorry for the late response. I really appreciate your interest about Xline and the contribution you've made.
I've reviewd your code logic, it's natural and clear. 
However, **recently we did a major change about this `xlinectl lock` command**, you can refer to the #820.
In essence, the lock usage is changed from : ` client.lock_client().lock(req).await?;`
to : `let mut xutex = Xutex::new().await?;
let xutex_guard = xutex.lock_unsafe().await?` 
So you may update your repo and make corresponding change, especially the : 
`pub(crate) async fn execute(client: &mut Client, matches: &ArgMatches) -> Result<()> `.
And for the test, we have a **validation test** in the: `Xline/scripts/validation_test.sh`, you can add your test case here.
If you have any questions, feel free to ask.

bhavik-goplani:
@CrystalAnalyst Thanks for the feedback. I'll close this pull request and create a new PR after pulling the recent changes.

PR #852, opened 2024-06-14
refactor: remove keys from command

bsbds:
The keys are not necessary in serialization, we could generate keys only when being used.

### DIFF
--- a/crates/curp-external-api/src/cmd.rs
+++ b/crates/curp-external-api/src/cmd.rs
@@ -44,7 +44,7 @@ pub trait Command: pri::Serializable + ConflictCheck + PbCodec {
     type ASR: pri::Serializable + PbCodec;
 
     /// Get keys of the command
-    fn keys(&self) -> &[Self::K];
+    fn keys(&self) -> Vec<Self::K>;
 
     /// Returns `true` if the command is read-only
     fn is_read_only(&self) -> bool;
--- a/crates/xline/src/server/lease_server.rs
+++ b/crates/xline/src/server/lease_server.rs
@@ -16,7 +16,7 @@ use utils::{
     task_manager::{tasks::TaskName, Listener, TaskManager},
 };
 use xlineapi::{
-    command::{Command, CommandResponse, CurpClient, KeyRange, SyncResponse},
+    command::{Command, CommandResponse, CurpClient, SyncResponse},
     execute_error::ExecuteError,
 };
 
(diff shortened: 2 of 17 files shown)


crates/xline/src/storage/lease_store/mod.rs
```diff
@@ -124,6 +124,8 @@ where
     }
 
     /// Get keys attached to a lease
+    /// FIXME: use this in conflict pools
```
Phoenix500526, line 127: Please leave an issue to track

crates/xline/src/server/lease_server.rs
```diff
@@ -132,18 +132,8 @@ where
...
-                    .get_keys(req.id)
-                    .into_iter()
-                    .map(|k| KeyRange::new(k, ""))
-                    .collect()
-            } else {
-                vec![]
-            }
-        };
-        let cmd = Command::new_with_auth_info(keys, request, auth_info);
+        // FIXME: get the keys in the conflict pools
```
Phoenix500526, line 135: Shall we leave an issue to track it?

crates/curp-external-api/src/cmd.rs
```diff
@@ -44,7 +44,7 @@ pub trait Command: pri::Serializable + ConflictCheck + PbCodec {
     type ASR: pri::Serializable + PbCodec;
 
     /// Get keys of the command
-    fn keys(&self) -> &[Self::K];
+    fn keys(&self) -> Vec<Self::K>;
```
Phoenix500526, line 47: Why do you return a `Vec` instead of `&[]`?

bsbds: The underlying method returned an owned type of the keys
```rust
impl RequestWrapper {
    /// Get keys of the request
    pub fn keys(&self) -> Vec<KeyRange> {
        match *self {
            RequestWrapper::RangeRequest(ref req) => req.keys(),
            RequestWrapper::PutRequest(ref req) => req.keys(),
            RequestWrapper::DeleteRangeRequest(ref req) => req.keys(),
            RequestWrapper::TxnRequest(ref req) => req.keys(),
            _ => vec![],
        }
    }
...
}

File: crates/xline/src/server/lease_server.rs (hunk 2 above)
Comment (bsbds) [reply crates/xline/src/server/lease_server.rs:135]: I'll submit a PR to fix this.

File: crates/curp-external-api/src/cmd.rs (hunk 3 above)
Comment (Phoenix500526) [reply crates/curp-external-api/src/cmd.rs:47]: Shall we implement a `key_ref` method to avoid unnecessary copying?

File: crates/xline/src/server/lease_server.rs (hunk 2 above)
Comment (bsbds) [reply crates/xline/src/server/lease_server.rs:135]: fixed in #861

File: crates/curp-external-api/src/cmd.rs (hunk 3 above)
Comment (bsbds) [reply crates/curp-external-api/src/cmd.rs:47]: Currently the `KeyRange` requires copying the keys of the command. Implementing a `key_ref` may involves significant change to the `KeyRange` itself.
```rust
impl KeyRange {
    /// New `KeyRange`
    #[inline]
    pub fn new(start: impl Into<Vec<u8>>, end: impl Into<Vec<u8>>) -> Self {
        let key_vec = start.into();
        let range_end_vec = end.into();
        ...
    }
```
I'll leave a TODO to refactor this in the future.

Phoenix500526 (approved):
LGTM, please squash these commits into one.

CrystalAnalyst (approved):
LGTM

PR #806, opened 2024-05-09
refactor: remove DB generic

bsbds:
Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

  The DB generic is redundant as no other type will ever implement `StorageApi` in Xline. This change aims to enhance readability of the codebase.

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #853, opened 2024-06-14
refactor: xline storage api

bsbds:
Depends-On: #673 #806 

Adds an `XlineStorageOps` trait as a replacement to `StorageApi` which is removed in #806

### DIFF
--- a/crates/xline/src/storage/db.rs
+++ b/crates/xline/src/storage/db.rs
@@ -1,6 +1,10 @@
+#![allow(clippy::multiple_inherent_impl)]
+
 use std::{collections::HashMap, path::Path, sync::Arc};
 
-use engine::{Engine, EngineType, Snapshot, StorageEngine, StorageOps, WriteOperation};
+use engine::{
+    Engine, EngineType, Snapshot, StorageEngine, StorageOps, Transaction, WriteOperation,
+};
 use prost::Message;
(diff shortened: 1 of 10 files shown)


crates/xline/src/storage/db.rs
```diff
@@ -56,76 +59,46 @@ impl DB {
...
+        keys: &[impl AsRef<[u8]>],
+    ) -> Result<Vec<Option<Vec<u8>>>, engine::EngineError> {
+        self.engine.get_multi(table, keys)
+    }
+}
+
+impl DB {
+    /// Creates a transaction
+    #[allow(unused)] // TODO: use this in the following refactor
+    pub(crate) fn transaction(&self) -> Transaction {
```
CrystalAnalyst, line 100: Is this a local transaction provided by rocksdb? How will we use it (where could we use it) in the following refactor ?

bsbds: Yes, it will be used in speculative execution to provide a temporary execution environment. In speculative execution, the results are not committed to the state machine, so we need to discard the transaction once we obtain the result.

Phoenix500526 (dismissed):
LGTM. Please squash these commits and rebase this pr

Phoenix500526 (approved):
LGTM

Issue #782, opened 2024-04-17
[Bug]: The test case curp::it read_state::read_state sometimes failed


Phoenix500526:
### Description about the bug

As the title description, the test case `curp::it read_state::read_state` fails sometimes. The log has been paste down below.

### Version

0.6.1 (Default)

### Relevant log output

```shell
PASS [   7.556s] curp server::raw_curp::tests::follower_will_not_start_election_when_heartbeats_are_received
     SIGABRT [   1.047s] curp::it read_state::read_state

--- STDOUT:              curp::it read_state::read_state ---

running 1 test
   0.020246518s DEBUG curp::server::curp_node: 2368797937378912452 to 10032832549340216369 sync follower task start
   0.020247149s DEBUG curp::server::curp_node: 2368797937378912452 to 3383425158938590665 sync follower task start
   0.020811636s DEBUG curp::server::raw_curp: 10032832549340216369 becomes the leader
   0.020982688s DEBUG curp::server::curp_node: 10032832549340216369 to 2368797937378912452 sync follower task start
   0.021027802s DEBUG curp::server::curp_node: 10032832549340216369 to 3383425158938590665 sync follower task start
   0.021459411s DEBUG curp::server::curp_node: 3383425158938590665 to 10032832549340216369 sync follower task start
   0.021489868s DEBUG curp::server::curp_node: 3383425158938590665 to 2368797937378912452 sync follower task start
   0.041409297s DEBUG curp_append_entries: curp::server::raw_curp: 3383425158938590665 updates to term 1 and becomes a follower
   0.041408692s DEBUG curp_append_entries: curp::server::raw_curp: 2368797937378912452 updates to term 1 and becomes a follower
   0.361394941s DEBUG curp_propose: curp::server::raw_curp: 10032832549340216369 gets proposal for cmd(0#8502775129477426034)
   0.361697215s DEBUG curp_propose: curp::server::raw_curp: 3383425158938590665 gets proposal for cmd(0#8502775129477426034)
   0.362214577s DEBUG curp::rpc::connect: client request a client id
   0.362988157s DEBUG curp_propose: curp::server::raw_curp: 2368797937378912452 gets proposal for cmd(0#8502775129477426034)
   0.363452834s DEBUG curp_wait_synced: curp::server::curp_node: 10032832549340216369 get wait synced request for cmd(0#8502775129477426034)
[...]

PR #851, opened 2024-06-14
feat: move id barrier to utils

bsbds:
Make `IdBarrier` reusable

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

Issue #501, opened 2023-11-02
[Bug]: Read different values in execute during read state


bsbds:
## Description
In read state, after the wait for cmd ids or log index are completed, we directly read from the current storage. However, I find that if there's multiple reads for a same key in a txn request, it might read different values.

The root cause of this is that duing serializable read there might be other commands being executed on the same server. And one txn may read the storage multiple times. Thus will lead to inconsistency in a single txn.

## Solution
When we do read operations in a txn request, we specify a `read_revision` for reads. When we execute a range request, we use the same read_revision to read the result. Thus this can gaurentee that we can read the same value for the same key.

## Details

To obtain the `read_revision`. We need let the barriers return the revision of that command we waited, and pass it to `execute` in `do_serializable`.

Phoenix500526:
Closed in #857

PR #857, opened 2024-06-17
refactor: remove read state

bsbds:
Depends-On: #851 #852 

Closes: #782, #501, #497, as we don't rely on this mechanism any more

Removes the current read state mechanism, and sets all proxied etcd commands to use slow path.

The read state requires the leader to query it's `UncommittedPool`. However, for every query, we must lock it in another tokio worker thread. Lock in multiple threads could reduce general performance, I think it's best to remove the current implementation and reimplement it in the future.

All etcd commands need to use slow path because we need 2RTTs to acquire a revision.


Phoenix500526 (commented):
Please leave an issue to track this problem: When a network partition occurs, the leader in the smaller partition may not be aware that a new leader has already been elected in the cluster. If the original leader continues to serve read-only requests to clients during this time, it could result in stale reads.

bsbds:
> Please leave an issue to track this problem: when a network partition occurs, the leader in the smaller partition may not be aware that a new leader has already been elected in the cluster. If the original leader continues to serve read-only requests to clients during this time, it could result in stale reads.

Tracked in https://github.com/xline-kv/Xline/issues/870

Phoenix500526 (approved):

PR #808, opened 2024-05-09
refactor: xline command execution

bsbds:
Depends-On: #687 #807 #673 #853 

This PR refactors xline kv storage and xline command executor.

Some tests will not be fixed in this PR; they will be fixed in #821.

Please briefly answer these questions:

* what problem are you trying to solve? (or if there's no problem, what's the motivation for this change?)

* what changes does this pull request make?

* are there any non-obvious implications of these changes? (does it break compatibility with previous versions, etc)

PR #858, opened 2024-06-17
refactor: curp storage api

bsbds:
Depends-On: #808


crates/curp/src/server/storage/db.rs
```diff
@@ -198,20 +206,23 @@ mod tests {
         let storage_cfg = EngineConfig::RocksDB(db_dir.clone());
         {
             let s = DB::<TestCommand>::open(&storage_cfg)?;
-            s.flush_voted_for(1, 222).await?;
-            s.flush_voted_for(3, 111).await?;
+            let (voted_for, entries) = s.recover()?;
```
rogercloud, line 209: If revover is necessary for the opened DB, why not included it into the `open` function?

bsbds: The db is only recovered in CurpNode, however, the xline server opens the db before pass it to `CurpNode`. If we merge open and recover, it would require `voted_for` and `entries` be passed around.

crates/curp-external-api/src/cmd.rs
```diff
@@ -26,7 +26,7 @@ impl<T> pri::Serializable for T where T: pri::ThreadSafe + Clone + Serialize + D
 
 /// Command to execute on the server side
 #[async_trait]
-pub trait Command: pri::Serializable + ConflictCheck + PbCodec {
+pub trait Command: pri::Serializable + ConflictCheck + PbCodec + Unpin {
```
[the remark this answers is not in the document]
bsbds: It is used in deprecated async version.  Removed.

crates/curp/src/server/storage/db.rs
rogercloud: So we'd better to comment somewhere to tell the api users that these two APIs should be used together.

bsbds: Add in 88fc6465ce3df65a371ff0a40e226b5af910eb39

rogercloud (approved):

Issue #840, opened 2024-06-07
[Bug]: test_unary_propose_fast_path_fallback_slow_path will randomly failed 


Phoenix500526:
### Description about the bug

As the title described, the test case `test_unary_propose_fast_path_fallback_slow_path` will randomly failed. 

### Version

0.6.1 (Default)

### Relevant log output

```shell
--- STDOUT:              curp client::tests::test_unary_propose_fast_path_fallback_slow_path ---

running 1 test
2024-06-07T07:51:50.826655Z  WARN test_unary_propose_fast_path_fallback_slow_path: curp::client::unary: propose cmd(0#16819838916010368377) to server(4) error: KeyConflict(())
2024-06-07T07:51:50.826753Z  WARN test_unary_propose_fast_path_fallback_slow_path: curp::client::unary: propose cmd(0#16819838916010368377) to server(3) error: KeyConflict(())
2024-06-07T07:51:50.927001Z DEBUG test_unary_propose_fast_path_fallback_slow_path: curp::client::unary: slow round for cmd(0#16819838916010368377) succeed
test client::tests::test_unary_propose_fast_path_fallback_slow_path ... FAILED

failures:

failures:
    client::tests::test_unary_propose_fast_path_fallback_slow_path

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.11s


--- STDERR:              curp client::tests::test_unary_propose_fast_path_fallback_slow_path ---
thread 'client::tests::test_unary_propose_fast_path_fallback_slow_path' panicked at crates/curp/src/client/tests.rs:583:5:
assertion `left == right` failed
  left: 3
 right: 5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

   Canceling due to test failure: 3 tests still running
        PASS [   0.023s] curp client::tests::test_unary_slow_round_fetch_leader_first
        PASS [   0.123s] curp client::tests::test_unary_propose_slow_path_works
        PASS [   1.910s] benchmark bench_client::test::test_new_etcd_client
        PASS [   2.092s] benchmark bench_client::test::test_new_xline_client
------------
     Summary [   2.101s] 25/345 tests run: 24 passed, 1 failed, 0 skipped
        FAIL [   0.125s] curp client::tests::test_unary_propose_fast_path_fallback_slow_path
error: test run failed
[...]

PR #859, opened 2024-06-18
fix: remove counter to keep aligned with early return in fast path

CrystalAnalyst:

   Update testing logic to keep aligned with early return in fast path.
   **Since we can return early in fast path when the number of conlicts is so large that we cannot reach SuperQuorum.**
   So we **DO NOT NEED to force** all the 5 servers to handle the propose. 
   related issue : #840 

   Remove the counter, and the rest part works fine (Successfully fallback to slow path) .


Phoenix500526 (approved):
LGTM

bsbds (approved):

Issue #824, opened 2024-05-16
[Feature]: Add exec command args for xlinectl watch command 


Phoenix500526:
### Description about the feature

The etcdctl provides [the exec-command arg](https://github.com/etcd-io/etcd/tree/main/etcdctl#watch-options-key-or-prefix-range_end----exec-command-arg1-arg2-) for the watch command. Therefore, when you want to execute some commands after triggered by watch command, you can use one-line command, like `./etcdctl watch foo -- sh -c "env | grep ETCD_WATCH_"`. I think this way is more user-friendly.

So far, the xlinectl doesn't support the exec command args for the watch command. I think it's not hard to implement. I leave an issue here to track. Please feel free to take this issue if you are interested in it. And please let me know if you need any help or more information.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

panwarharsh15:
Hey, i don't understand your issue , can you explain  what problem are you facing?

Phoenix500526:
Hi, @panwarharsh15 ! Here is an etcdctl watch command example:
```bash
./etcdctl watch foo -- sh -c "env | grep ETCD_WATCH_"

# PUT
# foo
# bar
# ETCD_WATCH_REVISION=11
# ETCD_WATCH_KEY="foo"
# ETCD_WATCH_EVENT_TYPE="PUT"
# ETCD_WATCH_VALUE="bar"
```
We can split this command `./etcdctl watch foo -- sh -c "env | grep ETCD_WATCH_"` by -- into two parts: one is `./etcdctl watch foo`, and the other part is a shell command: `env | grep ETCD_WATCH_"`. It indicates that execute the watch command  `./etcdctl watch foo` and perform the shell command `env | grep ETCD_WATCH_` when the watch command is triggered.

And currently, the watch sub-command in `xlinectl` doesn't support this way. We can only perform `xlinectl watch foo` without `-- sh -c "env | grep ETCD_WATCH_"`.

PR #862, opened 2024-06-24
feat: add exec command support for xlinectl watch

CrystalAnalyst:

  To solve issue: #824 

  A : change the `xlinectl watch` command.

### DIFF
--- a/crates/xlinectl/src/command/watch.rs
+++ b/crates/xlinectl/src/command/watch.rs
@@ -1,17 +1,21 @@
 use std::io;
+use std::{collections::HashMap, ffi::OsString};
 
 use anyhow::{anyhow, Result};
 use clap::{arg, value_parser, ArgMatches, Command};
+use std::process::Command as StdCommand;
 use xline_client::{
     error::XlineClientError,
     types::watch::{WatchRequest, Watcher},
     Client,
 };
 use xlineapi::command::Command as XlineCommand;
+use xlineapi::WatchResponse;
 
 use crate::utils::printer::Printer;
 
-/// Definition of `watch` command
+/// Definition of `watch` command :
+/// `WATCH [options] [key or prefix] [range_end] [--] [exec-command arg1 arg2 ...]`
 pub(crate) fn command() -> Command {
     Command::new("watch")
         .about("Watches events stream on keys or prefixes")
@@ -25,6 +29,12 @@ pub(crate) fn command() -> Command {
         .arg(arg!(--pre_kv "Get the previous key-value pair before the event happens"))
         .arg(arg!(--progress_notify "Get periodic watch progress notification from server"))
         .arg(arg!(--interactive "Interactive mode"))
+        .arg(
+            arg!(<exec_command> "command to execute after -- ")
+                .num_args(1..)
+                .required(false)
+                .last(true),
+        )
 }
 
 /// a function that builds a watch request with existing fields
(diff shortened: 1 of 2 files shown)


crates/xlinectl/src/command/watch.rs
```diff
@@ -67,23 +78,112 @@ pub(crate) async fn execute(client: &mut Client, matches: &ArgMatches) -> Result
...
+    if command.len() > 1 {
+        let _ = cmd.args(&command[1..]);
+    }
+    // executing the command
+    let output = cmd.output()?;
+    if !output.status.success() {
+        eprintln!("This command is not supported yet!");
+    }
+    println!("{}", String::from_utf8_lossy(&output.stdout));
+    eprintln!("{}", String::from_utf8_lossy(&output.stderr));
```
Phoenix500526, line 183: Why do you print the output message twice?

crates/xlinectl/src/command/watch.rs
```diff
@@ -67,23 +78,112 @@ pub(crate) async fn execute(client: &mut Client, matches: &ArgMatches) -> Result
...
+    }
+
+    // collecting the args
+    if command.len() > 1 {
+        let _ = cmd.args(&command[1..]);
+    }
+    // executing the command
+    let output = cmd.output()?;
+    if !output.status.success() {
+        eprintln!("This command is not supported yet!");
```
Phoenix500526, line 180: Why do you print this message? I think directly print the `output` is better.

crates/xlinectl/src/command/watch.rs
```diff
@@ -67,23 +78,112 @@ pub(crate) async fn execute(client: &mut Client, matches: &ArgMatches) -> Result
...
+
+        execute_inner(command_to_execute, envs)?;
+    }
+
+    Ok(())
+}
+
+/// Actual executor responsible for the user command,
+/// command format: [exec-command arg1 arg2 ...]
+#[allow(clippy::indexing_slicing)]
```
Phoenix500526, line 164: Please add a comment to specify the reason.

Phoenix500526 (commented):
Please add e2e test in our validation test.

crates/xlinectl/src/command/watch.rs
```diff
@@ -1,17 +1,22 @@
+#![allow(clippy::let_underscore_untyped)]
```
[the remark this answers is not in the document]
CrystalAnalyst: To beautify the code, otherwise I have to write something like : `let _ : &mut Command = cmd.env(key, value)`, especially this `: &mut Command` type annotation. I think it's unnecessary and not elegant, so I add this lint after checking its safety.

crates/xlinectl/src/command/watch.rs
CrystalAnalyst: My original intention is like this : one for normal case (stdout), and one for error case (stderr). Now I have changed it into a better one.

crates/xlinectl/src/command/watch.rs
CrystalAnalyst: Your advice is perferred and I adjusted my code.

crates/xlinectl/src/command/watch.rs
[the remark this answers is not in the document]
Phoenix500526: What I mean is, if you put #![allow(clippy::let_underscore_untyped)] at the very beginning of the file, it means that this lint is allowed for all the code in the entire file.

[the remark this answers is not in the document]
CrystalAnalyst: All right! I'll go and place it at right place.

Phoenix500526 (dismissed):
LGTM

scripts/validation_test.sh
```diff
@@ -137,9 +137,9 @@ EOF
 watch_validation() {
     log::info "watch validation test running..."
 
-    command="${ETCDCTL} watch watch_key"
+    command="${ETCDCTL} watch watch_key -- echo watch event received"
```
liangyuanpeng, line 140: Is it better to add a new test rather than modify the existing one? like `watch_with_command_validation` This means that the original watch test still needs to be retained.

Phoenix500526 (approved):
LGTM

Issue #819, opened 2024-05-14
[Refactor]: Remove request wrappers in xline-client


Phoenix500526:
We implement some request wrappers in the xline-client crate, like 
```Rust
/// Xline/crates/xline-client/src/types/lock.rs
/// Request for `Lock`
#[derive(Debug, PartialEq)]
pub struct LockRequest {
    /// The inner request
    pub(crate) inner: xlineapi::LockRequest,
    /// The ttl of the lease that attached to the lock
    pub(crate) ttl: i64,
}

impl LockRequest {
    /// Creates a new `LockRequest`
    #[inline]
    #[must_use]
    pub fn new(name: impl Into<Vec<u8>>) -> Self {
        ...
    }

    /// Set lease.
    #[inline]
    #[must_use]
    pub const fn with_lease(mut self, lease: i64) -> Self {
        self.inner.lease = lease;
        self
    }

    /// Set session TTL.
    /// Will be ignored when lease id is set
    #[inline]
    #[must_use]
    pub const fn with_ttl(mut self, ttl: i64) -> Self {
        self.ttl = ttl;
        self
    }
}
```
Therefore, when we want to perform a lock operation by xline-client, we should write some code like this:
```Rust
#[tokio::main]
async fn main() -> Result<()> {
     ...
     let lock_request = LockRequest::new("lock-test").with_ttl(50);
     // acquire a lock session
     let session = lock_client
         .lock(lock_request)
         .await?;
     ....
     Ok(())
}
```
This way is not very user-friendly. I prefer to use `lock_client.lock("lock-test", 50)` rather than `lock_client.lock(LockRequest::new("lock-test").with_ttl(50))`.

PR #882, opened 2024-07-11
refactor(client)!: Remove request wrappers - KvClient::put

lxl66566:

    - one of the series of #819 refactor, to remove request wrappers in xline-client and make user interface easier to use.

    - change function in `clients/kv.rs` from `fn put(&self, request: PutRequest)` to `put(&self, key: impl Into<Vec<u8>>, value: impl Into<Vec<u8>>, option: Option<PutOptions>)`
    - Renames `types::PutRequest` to `types::PutOptions`, made it an option for put function. The option will be combined with key and value to a `xlineapi::PutRequest` inside the `KvClient.put()` function.
        -  If it gets a None, construct a default PutOptions instead.
    - TxnOp::put function has been changed as well.

    - yes, it's a breaking change and changes all `put()` function thoughtout the crate.

### DIFF
--- a/crates/xline-client/examples/kv.rs
+++ b/crates/xline-client/examples/kv.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use xline_client::{
     types::kv::{
-        CompactionRequest, Compare, CompareResult, DeleteRangeRequest, PutRequest, RangeRequest,
+        CompactionRequest, Compare, CompareResult, DeleteRangeRequest, PutOptions, RangeRequest,
         TxnOp, TxnRequest,
     },
     Client, ClientOptions,
@@ -17,8 +17,8 @@ async fn main() -> Result<()> {
         .kv_client();
 
     // put
-    client.put(PutRequest::new("key1", "value1")).await?;
-    client.put(PutRequest::new("key2", "value2")).await?;
+    client.put("key1", "value1", None).await?;
+    client.put("key2", "value2", None).await?;
 
     // range
     let resp = client.range(RangeRequest::new("key1")).await?;
@@ -49,7 +49,9 @@ async fn main() -> Result<()> {
         .when(&[Compare::value("key2", CompareResult::Equal, "value2")][..])
         .and_then(
             &[TxnOp::put(
-                PutRequest::new("key2", "value3").with_prev_kv(true),
+                "key2",
+                "value3",
+                Some(PutOptions::default().with_prev_kv(true)),
             )][..],
         )
         .or_else(&[TxnOp::range(RangeRequest::new("key2"))][..]);
--- a/crates/xline-client/examples/lock.rs
+++ b/crates/xline-client/examples/lock.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use xline_client::{
     clients::Xutex,
-    types::kv::{Compare, CompareResult, PutRequest, TxnOp},
+    types::kv::{Compare, CompareResult, PutOptions, TxnOp},
     Client, ClientOptions,
 };
 
@@ -23,7 +23,9 @@ async fn main() -> Result<()> {
         .txn_check_locked_key()
         .when([Compare::value("key2", CompareResult::Equal, "value2")])
         .and_then([TxnOp::put(
-            PutRequest::new("key2", "value3").with_prev_kv(true),
+            "key2",
+            "value3",
+            Some(PutOptions::default().with_prev_kv(true)),
         )])
         .or_else(&[]);
 
--- a/crates/xline-client/src/clients/kv.rs
+++ b/crates/xline-client/src/clients/kv.rs
@@ -8,7 +8,7 @@ use xlineapi::{
 
 use crate::{
     error::Result,
-    types::kv::{CompactionRequest, DeleteRangeRequest, PutRequest, RangeRequest, TxnRequest},
+    types::kv::{CompactionRequest, DeleteRangeRequest, PutOptions, RangeRequest, TxnRequest},
     AuthService, CurpClient,
 };
 
@@ -65,7 +65,7 @@ impl KvClient {
     /// # Examples
     ///
     /// ```no_run
-    /// use xline_client::{types::kv::PutRequest, Client, ClientOptions};
+    /// use xline_client::{types::kv::PutOptions, Client, ClientOptions};
     /// use anyhow::Result;
     ///
     /// #[tokio::main]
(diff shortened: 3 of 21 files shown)


crates/xline-client/examples/kv.rs
```diff
@@ -17,8 +17,8 @@ async fn main() -> Result<()> {
         .kv_client();
 
     // put
-    client.put(PutRequest::new("key1", "value1")).await?;
-    client.put(PutRequest::new("key2", "value2")).await?;
+    client.put("key1", "value1", None).await?;
+    client.put("key2", "value2", None).await?;
```
bsbds, line 21: Can we implement a builder pattern that supports:
```rust
client.put("key1", "value1").await?;
client.put("key1", "value1").with_prev_key().await?;
```
Then we don't need to explicitly write a `PutOptions`.

lxl66566: That sounds good! But how can we implement like that?
Suppose the `put` function returns a struct `PutRequest`, do we need 
1. `impl Future for PutRequest` to make it awaitable?
2. `impl PutRequestBuilder` with an `async fn execute()`, and call it as `client.put("key1", "value1").with_prev_key().execute().await?;`
3. [using traits](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=62c2dff987602f3a6f1b6b6cefaf4616), but needs to record a `prev_key` state in Client member.

bsbds: I think implement a future could be the most convenient. Maybe something like: 
```rust
struct PutFut {
    inner: PutRequest,
}

impl Future for PutFut {
    // send the command
}

impl PutFut {
    fn with_prev_key(mut self) -> Self {
        self.inner.prev_key = true;
        self
    }
}

impl Client {
    fn put(&self, key: Vec<u8>, value: Vec<u8>) -> PutFut {
        ...
    }
}


```

lxl66566: There're some questions:
1. Since we have to send command in Future impl, we need to pass `curp_client` and `token` to `PutFut`
2. We need to construct command in `poll()`. `poll()` will be called multiple times, the command needs to be stored somewhere. But `Self` has a mut ref (in `poll()` parameter), so we cannot store it in `PutFut`. ([ex code here](https://gist.github.com/lxl66566/2d39e1d3efc8715cb77ce33dafa8d301), `cmd` does not live long enough).
3. I don't know how to modify the `TxnOp::put` impl below

bsbds: For 2 you can move the command into the future
```rust
            let cmd = Command::new(RequestWrapper::from(self.inner.take().unwrap()));
            let client = Arc::clone(self.curp_client);
            let token = self.token.cloned();
            let fut = async move { client.propose(&cmd, token.as_ref(), true).await };
            self.fut = Some(Box::pin(fut));
 ```

Comment (bsbds) [reply crates/xline-client/examples/kv.rs:21]: For 3, maybe:
```rust
    client.txn().when(...).and_then(|c| {
        c.put("key0", "value0").with_prev_key();
        c.put("key1", "value1");
    }).await?;
```

bsbds: If you find it too complex to implement, you can always opt for the original approach.

crates/xline-client/examples/kv.rs
```diff
@@ -12,13 +11,13 @@ async fn main() -> Result<()> {
...
-    let client = Client::connect(curp_members, ClientOptions::default())
+    let mut client = Client::connect(curp_members, ClientOptions::default())
         .await?
         .kv_client();
 
     // put
-    client.put(PutRequest::new("key1", "value1")).await?;
-    client.put(PutRequest::new("key2", "value2")).await?;
+    client.put("key1", "value1").await?;
+    client.put("key2", "value2").await?;
```
bsbds, line 20: Please add example of using the builder.

crates/xline-client/examples/lock.rs
```diff
@@ -13,21 +13,20 @@ async fn main() -> Result<()> {
...
-        .and_then([TxnOp::put(
-            PutRequest::new("key2", "value3").with_prev_kv(true),
-        )])
-        .or_else(&[]);
+    let lock_check = xutex_guard.txn_check_locked_key();
 
-    let _resp = kv_client.txn(txn_req).await?;
+    let _resp = kv_client
+        .when(lock_check)
+        .when(Compare::value("key2", CompareResult::Equal, "value2"))
```
bsbds, line 26: I suggest to merge multiple `when`, use a vector of `Compare` instead.

crates/xline-client/src/clients/kv.rs
```diff
@@ -25,6 +27,8 @@ pub struct KvClient {
     kv_client: xlineapi::KvClient<Channel>,
     /// The auth token
     token: Option<String>,
+    /// TxnRequest
+    txn: TxnRequest,
```
bsbds, line 31: Do not store state in the client itself; create a new builder type instead.

crates/xline-client/src/clients/kv.rs
```diff
@@ -173,75 +189,117 @@ impl KvClient {
...
+    #[inline]
+    pub fn and_then<F, O>(&mut self, operation: F) -> &mut Self
+    where
+        F: FnOnce(&mut Self) -> O,
+        O: Into<TxnOp>,
+    {
+        let temp = operation(self).into();
+        self.txn.and_then(temp);
+        self
+    }
```
bsbds, line 264: I believe it should be able to accept multiple operations because it will be hard to use if the user needs to perform multiple operations.

crates/xline-client/src/types/kv.rs
```diff
@@ -1,29 +1,68 @@
...
+
+/// The type of the future inside [`PutFut`].
+// type RequestFutureType<'a> = Option<BoxFuture<'a, ResponseType>>;
+type RequestFutureType = Option<Pin<Box<dyn Future<Output = ResponseType> + Send>>>;
+
+/// Future for `Put`, make it awaitable.
+///
+/// Before first poll, fut will be [`Option::None`].
+/// Once It's been polled, fut will be [`Option::Some`] and inner will be [`Option::None`].
+#[pin_project]
```
bsbds, line 33: This macro does not seem to be used.

crates/xline-client/src/types/kv.rs
```diff
@@ -1,29 +1,68 @@
...
+pub struct PutFut {
```
bsbds, line 34: I think it's better to add a lifetime parameter to this struct. Otherwise it may outlive the original client.

bsbds (approved):

PR #884, opened 2024-07-12
refactor(client)!: Remove request wrappers - AuthClient::user_add

lxl66566:

    - one of the series of #819 refactor, to remove request wrappers in xline-client and make user interface easier to use.

    - change `fn user_add(&self, mut request: AuthUserAddRequest)` to `user_add(&self, name: impl Into<String>,  password: impl AsRef<str>, options: Option<xlineapi::UserAddOptions>)`
    - remove `types::AuthUserAddRequest`

    - yes, it's a breaking change and may break code uses old version of `user_add()`.

### DIFF
--- a/crates/xline-client/src/clients/auth.rs
+++ b/crates/xline-client/src/clients/auth.rs
@@ -15,7 +15,7 @@ use crate::{
     error::{Result, XlineClientError},
     types::auth::{
         AuthRoleAddRequest, AuthRoleDeleteRequest, AuthRoleGetRequest,
-        AuthRoleGrantPermissionRequest, AuthRoleRevokePermissionRequest, AuthUserAddRequest,
+        AuthRoleGrantPermissionRequest, AuthRoleRevokePermissionRequest,
         AuthUserChangePasswordRequest, AuthUserDeleteRequest, AuthUserGetRequest,
         AuthUserGrantRoleRequest, AuthUserRevokeRoleRequest,
     },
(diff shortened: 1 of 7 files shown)


crates/xline-client/src/clients/auth.rs
```diff
@@ -222,33 +222,44 @@ impl AuthClient {
...
     ///     Ok(())
     /// }
     ///```
     #[inline]
-    pub async fn user_add(&self, mut request: AuthUserAddRequest) -> Result<AuthUserAddResponse> {
-        if request.inner.name.is_empty() {
+    pub async fn user_add(
+        &self,
+        name: impl Into<String>,
+        password: impl AsRef<str>,
```
bsbds, line 233: Should the password be made optional and put into options?

lxl66566: if i use `password: Option<impl AsRef<str>>`, then the `user_add` function should pass a `None::<&str>`, because *type annotations needed*. That's a bit complex.
I prefer to use empty str, because in `xlinectl/src/command/user/add.rs`, user must input a password without `--no_password` given. So we'll always get a String.
If you think Option is important to indicate **no** password, maybe we should use `Option<String>`?

bsbds: That makes sense, please add a comment describing that an empty string means no password.

bsbds (approved):

PR #889, opened 2024-07-14
refactor(client)!: refactor AuthClient::user_change_password

lxl66566:
Depends-On: #882 

    - change `user_change_password(&self,mut request: AuthUserChangePasswordRequest)` to `user_change_password(&self,name: impl Into<String>,password: impl AsRef<str>)`
    - remove `types::AuthUserChangePasswordRequest`
 
    - yes


lxl66566:
And there's another question: do we need to remove the `string password = 2;` in `rpc.proto`, since the password in api should always be empty string, and there's no need to construct it from function parameter.

LucienY01 (approved):

lxl66566:
rebased

LucienY01 (approved):

PR #898, opened 2024-07-18
refactor(client)!: Remove request wrappers - AuthClient::role_grant_permission

lxl66566:
Depends-On: #882

### DIFF
--- a/crates/xline-client/examples/auth_role.rs
+++ b/crates/xline-client/examples/auth_role.rs
@@ -1,8 +1,6 @@
 use anyhow::Result;
 use xline_client::{
-    types::auth::{
-        AuthRoleGrantPermissionRequest, AuthRoleRevokePermissionRequest, Permission, PermissionType,
-    },
+    types::auth::{AuthRoleRevokePermissionRequest, PermissionType},
     Client, ClientOptions,
 };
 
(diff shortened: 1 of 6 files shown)


crates/xline-client/examples/auth_role.rs
```diff
@@ -23,16 +22,10 @@ async fn main() -> Result<()> {
 
     // grant permissions to roles
     client
-        .role_grant_permission(AuthRoleGrantPermissionRequest::new(
-            "role1",
-            Permission::new(PermissionType::Read, "key1"),
-        ))
+        .role_grant_permission("role1", Permission::new(PermissionType::Read, "key1"))
```
bsbds, line 25: Please also refactor the `Permission` type. Adds some methods like `Permission::read`, `Permission::write` and `Permission::readwrite`

lxl66566: Ok. The Permission type also has functions like `with_prefix`, `with_from_key`, which needs to use `range_end.rs` introduced by #906. I will fix it based on that.
There's also a pr (#899) has the same problem.

lxl66566: I flattened the `role_grant_permission` api into:
```
pub async fn role_grant_permission(
        &self,
        name: impl Into<String>,
        perm_type: PermissionType,
        perm_key: impl Into<Vec<u8>>,
        range_option: Option<RangeOption>,
    )
```
how about this?

bsbds (approved):

PR #904, opened 2024-07-19
fix(ci): add doc test

lxl66566:

    - ci does not test doc

    - add a doc test
        - note that [nextest does not support doc test](https://nexte.st/docs/running/#fn:doctest). [ref](https://github.com/nextest-rs/nextest/issues/16)

    - no

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -24,13 +24,13 @@ jobs:
               name: "Normal",
               args: "",
               rustflags: "",
-              test: "llvm-cov nextest --all-features --workspace --codecov --output-path codecov.info",
+              test: "llvm-cov nextest --all-features --workspace --codecov --output-path codecov.info && cargo test --doc",
             }
           - {
               name: "Madsim",
               args: "--package=simulation",
               rustflags: "--cfg madsim",
-              test: "nextest run --package=simulation",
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -6,6 +6,7 @@ on:
       - master
   pull_request:
     types: [opened, synchronize, reopened]
+  workflow_dispatch:
```
bsbds, line 9: Why do you need to add a workflow_dispatch condition?

.github/workflows/pull_request.yml
```diff
@@ -24,7 +25,7 @@ jobs:
               name: "Normal",
               args: "",
               rustflags: "",
-              test: "llvm-cov nextest --all-features --workspace --codecov --output-path codecov.info",
+              test: "llvm-cov nextest --all-features --workspace --codecov --output-path codecov.info && cargo test --doc",
             }
           - {
               name: "Madsim",
```
bsbds, line 31: Please also add doc tests to madsim cfg.

.github/workflows/pull_request.yml
lxl66566: That's not needed. removed.

bsbds (approved):

PR #905, opened 2024-07-19
refactor(client)!: Remove request wrappers - LeaseClient::*

lxl66566:


crates/xline-client/src/clients/lease.rs
```diff
@@ -88,19 +88,22 @@ impl LeaseClient {
...
     /// ```
     #[inline]
-    pub async fn grant(&self, mut request: LeaseGrantRequest) -> Result<LeaseGrantResponse> {
-        if request.inner.id == 0 {
-            request.inner.id = self.id_gen.next();
+    pub async fn grant(&self, ttl: i64, id: Option<i64>) -> Result<LeaseGrantResponse> {
+        let mut id = id.unwrap_or_default();
+        if id == 0 {
+            id = self.id_gen.next();
         }
```
LucienY01, line 102: We can use `unwrap_or_else` here.

crates/xlinectl/src/command/lease/grant.rs
```diff
@@ -11,15 +11,15 @@ pub(super) fn command() -> Command {
...
-pub(super) fn build_request(matches: &ArgMatches) -> LeaseGrantRequest {
+pub(super) fn build_request(matches: &ArgMatches) -> i64 {
     let ttl = matches.get_one::<i64>("ttl").expect("required");
-    LeaseGrantRequest::new(*ttl)
+    *ttl
 }
 
 /// Execute the command
 pub(super) async fn execute(client: &mut Client, matches: &ArgMatches) -> Result<()> {
     let request = build_request(matches);
```
LucienY01, line 21: Please rename `request` `ttl`.

crates/xline-client/src/clients/lease.rs
lxl66566: if id = `Some(0)`, it needs to be generated as well. `unwrap_or_else` may not deal with that.

LucienY01: You are right. I didn't take this case into account.

LucienY01 (approved):

PR #917, opened 2024-07-25
refactor: remove cmd worker

bsbds:
Depends-On: #916 

This PR mainly removes the old cmd worker. We will replaced it with a new execution model instead.


crates/utils/src/task_manager/tasks.rs
```diff
@@ -54,10 +50,7 @@ enum_with_iter! {
 }
 
 /// All edges of task graph, the first item in each pair must be shut down before the second item
-pub const ALL_EDGES: [(TaskName, TaskName); 9] = [
-    (TaskName::ConflictCheckedMpmc, TaskName::CmdWorker),
-    (TaskName::CmdWorker, TaskName::CompactBg),
-    (TaskName::CmdWorker, TaskName::KvUpdates),
+pub const ALL_EDGES: [(TaskName, TaskName); 6] = [
```
Phoenix500526, line 53: How should we track this task if we remove the `CompactBg` from the `ALL_EDGES`? Is it possible that this task will fall off the task manager's radar?

bsbds: It could be canceled safely when Xline exits, because it doesn't have an dependent tasks.

crates/utils/src/task_manager/tasks.rs
```diff
@@ -1,10 +1,8 @@
...
-//       CMD_WORKER            LEASE_KEEP_ALIVE
-//         /     \                    |
-//  COMPACT_BG  KV_UPDATES      TONIC_SERVER       ELECTION
-//                    \        /      |      \       /
-//                   WATCH_TASK  CONF_CHANGE  LOG_PERSIST
+//                LEASE_KEEP_ALIVE
+//                       |
+// KV_UPDATES      TONIC_SERVER       ELECTION
+//       \        /      |      \       /
+//      WATCH_TASK  CONF_CHANGE  LOG_PERSIST
```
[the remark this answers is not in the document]
bsbds: done

Phoenix500526 (approved):
LGTM

PR #922, opened 2024-07-29
refactor(task_manager): introduce atomic_enum to remove magic numbers

lxl66566:

    - I found the `State` in task_manager is stored as `Arc<AtomicU8>` in `Listener` and `TaskManager`. And the code performs some `AtomicU8 -> State` transform, but there's no  `State -> AtomicU8` code, instead it uses magic numbers directly in the code.

    - Introduce [`atomic_enum`](https://crates.io/crates/atomic_enum) to make the enum `State` Atomic.

   - no

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -227,6 +227,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic_enum"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


LucienY01:
It seems that `atomic_enum` stores values in `AtomicUsize`. Would it be a bit wasteful in terms of space?

lxl66566:
> It seems that `atomic_enum` stores values in `AtomicUsize`. Would it be a bit wasteful in terms of space?

you are right, this is an unnecessary overhead. I will close this pr until there are better ways to improve it.

PR #930, opened 2024-07-31
chore: upgrade axum to 0.7

LucienY01:
Upgrade axum to 0.7, tonic to 0.12, hyper to 1.0,
opentelemetry to 0.24, opentelemetry_sdk to 0.24.1, opentelemetry_otlp to 0.17.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,7 @@ ignored = ["prost", "workspace-hack"]
 
 [patch.crates-io]
 # This branch update the tonic version for madsim. We should switch to the original etcd-client crate when new version release.
-madsim = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tonic = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tonic-build = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tokio = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-
+madsim = { git = "https://github.com/LucienY01/madsim.git", branch = "bz/tonic-0-12" }
+madsim-tonic = { git = "https://github.com/LucienY01/madsim.git", branch = "bz/tonic-0-12" }
+madsim-tonic-build = { git = "https://github.com/LucienY01/madsim.git", branch = "bz/tonic-0-12" }
+madsim-tokio = { git = "https://github.com/LucienY01/madsim.git", branch = "bz/tonic-0-12" }
--- a/crates/xline/src/storage/auth_store/store.rs
+++ b/crates/xline/src/storage/auth_store/store.rs
@@ -193,13 +193,13 @@ impl AuthStore {
     ) -> Result<CommandResponse, ExecuteError> {
         #[allow(clippy::wildcard_enum_match_arm)]
         let res = match *request {
-            RequestWrapper::AuthEnableRequest(ref req) => {
+            RequestWrapper::AuthEnableRequest(req) => {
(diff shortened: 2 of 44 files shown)


crates/xline/src/storage/auth_store/store.rs
```diff
@@ -193,13 +193,13 @@ impl AuthStore {
...
             }
-            RequestWrapper::AuthDisableRequest(ref req) => {
+            RequestWrapper::AuthDisableRequest(req) => {
                 Ok(self.handle_auth_disable_request(req).into())
             }
-            RequestWrapper::AuthStatusRequest(ref req) => {
+            RequestWrapper::AuthStatusRequest(req) => {
                 Ok(self.handle_auth_status_request(req).into())
             }
             RequestWrapper::AuthUserAddRequest(ref req) => {
```
Phoenix500526, line 205: Why do some requests use `req` while others use `ref req`?

LucienY01: Clippy indicated that it's more efficient to pass value than reference for `AuthStatusRequest`. The same reason applies to other similar changes.

Cargo.toml
```diff
@@ -20,10 +20,3 @@ resolver = "2"
 
 [workspace.metadata.cargo-machete]
 ignored = ["prost", "workspace-hack"]
-
-[patch.crates-io]
-# This branch update the tonic version for madsim. We should switch to the original etcd-client crate when new version release.
-madsim = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tonic = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tonic-build = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
-madsim-tokio = { git = "https://github.com/Phoenix500526/madsim.git", branch = "update-tonic" }
```
bsbds, line 29: Should we maintain a fork that tracks the upstream? It seems we constantly have to wait for a release in madsim.

crates/xline/src/utils/metrics.rs
```diff
@@ -49,17 +51,16 @@ pub fn init_metrics(config: &MetricsConfig) -> anyhow::Result<()> {
...
         .unwrap_or_else(|_| {
             unreachable!("local address 0.0.0.0:{} should be parsed", config.port())
         });
     info!("metrics server start on {addr:?}");
     let app = axum::Router::new().route(config.path(), axum::routing::any(metrics));
     let _ig = tokio::spawn(async move {
-        axum::Server::bind(&addr)
-            .serve(app.into_make_service())
-            .await
+        let listener = real_tokio::net::TcpListener::bind(addr).await?;
```
bsbds, line 62: Why do we need the original tokio crate?

Cargo.toml
LucienY01: OK.

crates/xline/src/utils/metrics.rs
LucienY01: `axum::serve()` doesn't accept a madsim-tokio TcpListener. The original codes `axum::Server::bind(&addr).serve(app.into_make_service())` create an original `tokio::net::TcpListener` internally, so I used the original tokio crate to achieve the same effect.

bsbds (changes requested):
Please signed off all the commits.

LucienY01:
Sign off all commits by rebase. No other changes.

Cargo.toml
bsbds: I think we should maintain the fork under xline-kv organization rather than personal account. We could address this in another PR instead of this one.

LucienY01:
> hello, could you squash your commits into one? @LucienY01

OK, it's done.

bsbds (approved):

PR #932, opened 2024-08-02
refactor(xlineapi): improve is_conflict

### LINKED ISSUE #379: improve the maintainability of `is_conflict` method
Currently, the is_conflict method in xine/src/server/command.rs is challenging to maintain, primarily due to the following reasons:

```Rust
impl ConflictCheck for Command {
    #[inline]
    fn is_conflict(&self, other: &Self) -> bool {
        if self.id == other.id {
            return true;
        }
        let this_req = &self.request.request;
        let other_req = &other.request.request;
        // auth read request will not conflict with any request except the auth write request
        if (this_req.is_auth_read_request() && other_req.is_auth_read_request())
            || (this_req.is_kv_request() && other_req.is_auth_read_request())
            || (this_req.is_auth_read_request() && other_req.is_kv_request())
        {
            return false;
        }
        // any two requests that don't meet the above conditions will conflict with each other
        // because the auth write request will make all previous token invalid
        if (this_req.is_auth_request()) || (other_req.is_auth_request()) {
            return true;
        }

        // Lease leases request is conflict with Lease grant and revoke requests
        if (this_req.is_lease_read_request() && other_req.is_lease_write_request())
            || (this_req.is_lease_write_request() && other_req.is_lease_read_request())
        {
            return true;
        }

        if this_req.is_compaction_request() && other_req.is_compaction_request() {
            return true;
        }

        if (this_req.is_txn_request() && other_req.is_compaction_request())
            || (this_req.is_compaction_request() && other_req.is_txn_request())
        {
            match (this_req, other_req) {
                (
                    &RequestWrapper::CompactionRequest(ref com_req),
                    &RequestWrapper::TxnRequest(ref txn_req),
                )
                | (
                    &RequestWrapper::TxnRequest(ref txn_req),
                    &RequestWrapper::CompactionRequest(ref com_req),
                ) => {
                    let target_revision = com_req.revision;
                    return txn_req.is_conflict_with_rev(target_revision)
                }
                _ => unreachable!("The request must be either a transaction or a compaction request! \nthis_req = {this_req:?} \nother_req = {other_req:?}")
            }
        }

        let this_lease_ids = get_lease_ids(this_req);
        let other_lease_ids = get_lease_ids(other_req);
        let lease_conflict = !this_lease_ids.is_disjoint(&other_lease_ids);
        let key_conflict = self
            .keys()
            .iter()
            .cartesian_product(other.keys().iter())
            .any(|(k1, k2)| k1.is_conflicted(k2));
        lease_conflict || key_conflict
    }
}
```

1. Some obviously conflicting requests require multiple conditional checks to determine the result. For example, two `CompactionRequest` are clearly conflicting, but they must pass the first three if statements before the control flow hit the right condition.
2. With the increase in requests, the possibilities of matches grow rapidly. For instance, with 10 different requests, there are 100 possible combinations of (this_req, other_req). Adding one more request increases the total combinations to (10 + 1)^2 = 121. Therefore, we need an intuitive declarative approach to describe whether different combinations result in conflicts. One possible solution is to use macros for abstraction.

lxl66566:

    - close #379

    - use macros to easily add conflict rules
        - There are two types of conflict check rules: one is based on it's type and readable/writable status, and another type is based on the infomation in the request, like `lease()`. The PR is mainly to solve the first type of check rules.
    - use a global conflict rules cache to cache conflict result based on request type checking

### DIFF
--- a/crates/xlineapi/src/classifier.rs
+++ b/crates/xlineapi/src/classifier.rs
@@ -1,0 +1,137 @@
+use super::RequestWrapper;
+
+/// Backend store of request
+#[allow(missing_docs)]
+#[derive(Debug, PartialEq, Eq)]
+pub enum RequestBackend {
+    Kv,
+    Auth,
+    Lease,
+    Alarm,
+}
+
+impl From<&RequestWrapper> for RequestBackend {
+    fn from(value: &RequestWrapper) -> Self {
+        match *value {
+            RequestWrapper::PutRequest(_)
+            | RequestWrapper::RangeRequest(_)
+            | RequestWrapper::DeleteRangeRequest(_)
+            | RequestWrapper::TxnRequest(_)
+            | RequestWrapper::CompactionRequest(_) => RequestBackend::Kv,
+            RequestWrapper::AuthEnableRequest(_)
+            | RequestWrapper::AuthDisableRequest(_)
+            | RequestWrapper::AuthStatusRequest(_)
+            | RequestWrapper::AuthRoleAddRequest(_)
+            | RequestWrapper::AuthRoleDeleteRequest(_)
+            | RequestWrapper::AuthRoleGetRequest(_)
+            | RequestWrapper::AuthRoleGrantPermissionRequest(_)
+            | RequestWrapper::AuthRoleListRequest(_)
+            | RequestWrapper::AuthRoleRevokePermissionRequest(_)
+            | RequestWrapper::AuthUserAddRequest(_)
+            | RequestWrapper::AuthUserChangePasswordRequest(_)
+            | RequestWrapper::AuthUserDeleteRequest(_)
+            | RequestWrapper::AuthUserGetRequest(_)
+            | RequestWrapper::AuthUserGrantRoleRequest(_)
+            | RequestWrapper::AuthUserListRequest(_)
+            | RequestWrapper::AuthUserRevokeRoleRequest(_)
+            | RequestWrapper::AuthenticateRequest(_) => RequestBackend::Auth,
+            RequestWrapper::LeaseGrantRequest(_)
+            | RequestWrapper::LeaseRevokeRequest(_)
+            | RequestWrapper::LeaseLeasesRequest(_) => RequestBackend::Lease,
+            RequestWrapper::AlarmRequest(_) => RequestBackend::Alarm,
+        }
+    }
+}
+
+/// Type of request. This is the extending of [`RequestBackend`].
+#[allow(missing_docs)]
+#[derive(Debug, PartialEq, Eq)]
+pub enum RequestType {
+    Put,
+    Compaction,
+    Txn,
+    Range,
+    Auth,
+    Lease,
+    Alarm,
+}
+
+impl From<&RequestWrapper> for RequestType {
+    fn from(value: &RequestWrapper) -> Self {
+        match *value {
+            RequestWrapper::PutRequest(_) => RequestType::Put,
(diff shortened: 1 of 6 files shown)


crates/xlineapi/src/classifier.rs
```diff
@@ -0,0 +1,171 @@
...
+            RequestWrapper::AuthUserListRequest(_) => 19,
+            RequestWrapper::AuthUserRevokeRoleRequest(_) => 20,
+            RequestWrapper::AuthenticateRequest(_) => 21,
+            RequestWrapper::LeaseGrantRequest(_) => 22,
+            RequestWrapper::LeaseRevokeRequest(_) => 23,
+            RequestWrapper::LeaseLeasesRequest(_) => 24,
+            RequestWrapper::AlarmRequest(_) => 25,
+        }
+    }
+}
```
bsbds, line 35: Derive `#[repr(u8)]` in `build.rs`?

crates/xlineapi/src/classifier.rs
```diff
@@ -0,0 +1,171 @@
...
+#[allow(missing_docs)]
+#[derive(Debug, PartialEq, Eq)]
+pub enum RequestBackend {
+    Kv,
+    Auth,
+    Lease,
+    Alarm,
+}
+
+impl From<&RequestWrapper> for RequestBackend {
```
bsbds, line 47: I believe it would be more effective to implement a method directly within the `RequestWrapper`. The `RequestBackend` serves more as an attribute rather than something that requires conversion.

crates/xlineapi/src/command.rs
```diff
@@ -222,67 +230,136 @@ pub struct Command {
...
+        let t_req = &self.request;
+        let o_req = &other.request;
+        let mut cache_key: (u8, u8) = (t_req.into(), o_req.into());
+        if cache_key.0 > cache_key.1 {
+            cache_key = (cache_key.1, cache_key.0);
+        }
+        if let Some(res) = CONFLICT_RULES_CACHE
+            .read()
+            .ok()
+            .and_then(|c| c.get(&cache_key).cloned())
```
bsbds, line 313: Why do you need a cache? I believe the matches are already efficient. Or maybe use a thread local cache instead.

crates/xlineapi/src/classifier.rs
lxl66566: How to add it? I tried `.field_attribute("Command.request_wrapper", ...)` but it seems not work.

crates/xlineapi/src/classifier.rs
lxl66566: fixed

crates/xlineapi/src/command.rs
lxl66566: thread_local used

crates/xlineapi/src/classifier.rs
bsbds: `.type_attribute("Command.request_wrapper", "#[repr(u8)]")`

lxl66566: This does not work as well?

lxl66566: OK, the `#[repr(u8)]` added successfully, but I still cannot transfer from `&RequestWrapper` to  `u8`?
see https://doc.rust-lang.org/reference/items/enumerations.html#casting, this `RequestWrapper` has field in its variant, so to cast it to u8 is unsafe.

bsbds: LGTM. Keep the current implementation.

bsbds (approved):

PR #935, opened 2024-08-05
fix(ci): free disk space to avoid CI runs out of space problem

lxl66566:

    - The CI failed a lot of time without error message, and it's not problem of code. When retry, the ci may pass, or fail as same.
    - related ci: https://github.com/xline-kv/Xline/actions/runs/10209311482/job/28247138131?pr=880
    - after one failure, i captured the output of github: 
图片

    - ~~introduce [free-disk-space](https://github.com/marketplace/actions/free-disk-space-ubuntu)~~
    - mount some big dirs to image, and delete them manually.
    - use sccache action instead of set up variables manually.
    
    - no

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -17,12 +17,8 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Configure sccache
-        uses: actions/github-script@v7
-        with:
-          script: |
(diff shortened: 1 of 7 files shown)


lxl66566:
The Free Disk Space step will not run in ubuntu host, but in docker contanier, which is not the expected behavior.

bsbds (approved):

PR #938, opened 2024-08-05
refactor: after sync implementation

bsbds:
This PR introduces the following major changes:
- Switch after sync to sync version (previously async)
- Rewrite error returning mechanism during after sync stage

During the after sync execution (in `crates/xline/src/server/command.rs`), we use batch persistent, and this will introduce two types of errors
- a execution error for a single command
- a batch persistent error

Thus, we cannot return immediately after encounter and error, but need to record the execution result for each command. A `ASResultStates` type is introduced to resolve this issue.

### DIFF
--- a/crates/xline/src/server/command.rs
+++ b/crates/xline/src/server/command.rs
@@ -359,7 +359,7 @@ impl<'a> ASResults<'a> {
     where
         F: Fn(&AfterSyncCmd<'_, Command>) -> Result<(), ExecuteError>,
     {
-        self.map_results(|(cmd, result_opt)| {
+        self.for_each_none_result(|(cmd, result_opt)| {
             if let Err(e) = op(cmd) {
                 let _ignore = result_opt.replace(Err(e));
             }
@@ -372,14 +372,14 @@ impl<'a> ASResults<'a> {
     where
         F: Fn(&AfterSyncCmd<'_, Command>) -> AfterSyncResult,
     {
-        self.map_results(|(cmd, result_opt)| {
+        self.for_each_none_result(|(cmd, result_opt)| {
             let _ignore = result_opt.replace(op(cmd));
         });
     }
 
-    /// Applies a given operation to each command-result pair in `cmd_results` where the result is `None`.
+    /// Applies the provided operation to each command-result pair in `cmd_results` where the result is `None`.
     #[allow(clippy::pattern_type_mismatch)] // can't be fixed
-    fn map_results<F>(&mut self, op: F)
+    fn for_each_none_result<F>(&mut self, op: F)
     where
         F: FnMut(&mut (AfterSyncCmd<'_, Command>, Option<AfterSyncResult>)),
     {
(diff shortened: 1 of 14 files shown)


crates/xline/tests/it/kv_test.rs
```diff
@@ -12,6 +12,11 @@ use xline_test_utils::{
 #[tokio::test(flavor = "multi_thread")]
 #[abort_on_panic]
 async fn test_kv_put() -> Result<(), Box<dyn Error>> {
+    std::env::set_var("RUST_LOG", "curp=debug,xline=debug");
+    _ = tracing_subscriber::fmt()
+        .compact()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init();
```
lxl66566, line 19: these code is same as crates/curp-test-utils/src/lib.rs init_logger()

crates/xline/src/server/command.rs
```diff
@@ -337,6 +334,83 @@ impl CommandExecutor {
...
+
+impl<'a> ASResultStates<'a> {
+    /// Creates a new [`ASResultStates`].
+    fn new(cmds: Vec<AfterSyncCmd<'a, Command>>) -> Self {
+        Self {
+            results: iter::repeat_with(|| None::<AfterSyncResult>)
+                .take(cmds.len())
+                .collect(),
+            cmds,
+        }
```
lxl66566, line 356: I don't know if the `cmds` and `results` are one-to-one relationship. 
If the answer is "yes", why not use a `Vec<(AfterSyncCmd<'a, Command>, Option<AfterSyncResult>)>` instead?

crates/xline/tests/it/kv_test.rs
bsbds: It was used for temporary debugging, removed.

crates/xline/src/server/command.rs
bsbds: Yes, they are one-to-one relationship. I have write a new version of `ASResultStates`

crates/xline/src/server/command.rs
```diff
@@ -338,76 +338,71 @@ impl CommandExecutor {
...
 
-    /// Skip if the command execution has already errored
-    #[allow(clippy::pattern_type_mismatch)] // Can't be fixed
-    fn filter_ok(
-        (_cmd, result_opt): &(&AfterSyncCmd<'a, Command>, &mut Option<AfterSyncResult>),
-    ) -> bool {
-        result_opt.as_ref().is_none()
+    /// Applies a given operation to each command-result pair in `cmd_results` where the result is `None`.
+    #[allow(clippy::pattern_type_mismatch)] // can't be fixed
+    fn map_results<F>(&mut self, op: F)
```
lxl66566, line 382: does `result == None` represents *waiting for after sync result* status? 
I think the name `map_results` is a bit confusing because it just apply op to pairs which has no result. And the `for_each` is also differ from `map`.

bsbds: Renamed to `for_each_none_result`.

lxl66566 (approved):

PR #945, opened 2024-08-12
refactor: task manager

bsbds:
This PR reafctor the task manager, allowing directly abort cancel safe tasks.

### DIFF
--- a/crates/curp/src/server/curp_node.rs
+++ b/crates/curp/src/server/curp_node.rs
@@ -194,7 +194,6 @@ impl<C: Command, CE: CommandExecutor<C>, RC: RoleChange> CurpNode<C, CE, RC> {
         ce: Arc<CE>,
         curp: Arc<RawCurp<C, RC>>,
         rx: flume::Receiver<Propose<C>>,
-        shutdown_listener: Listener,
     ) {
         /// Max number of propose in a batch
         const MAX_BATCH_SIZE: usize = 1024;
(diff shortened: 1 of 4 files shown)


crates/curp/src/server/curp_node.rs
```diff
@@ -753,23 +749,9 @@ impl<C: Command, CE: CommandExecutor<C>, RC: RoleChange> CurpNode<C, CE, RC> {
...
-        loop {
-            tokio::select! {
-                _ = shutdown_listener.wait() => {
-                    break;
-                }
-                Ok(task) = as_rx.recv_async() => {
-                    Self::handle_as_task(&curp, &cmd_executor, task).await;
-                }
-            }
+        while let Ok(task) = as_rx.recv_async().await {
```
lxl66566, line 753: This change seems related to some previous changes?
Could we ensure that when cluster shutdown, all senders will be dropped?

bsbds: This task could be directly canceled without waiting for the senders to be dropped. Because the tasks in the channel are already stored in the curp logs and will be re-executed during next recovery.

lxl66566: But since you remove the `shutdown_listener`, this while-let loop will jump out only when all senders are dropped. ([flume::Receiver::recv_async](https://docs.rs/flume/latest/flume/struct.Receiver.html#method.recv_async), if my comprehension is not wrong)

bsbds: During the node shutdown, it will be canceled after it yield to the runtime.

PR #946, opened 2024-08-12
feat: dedup

bsbds:
Ref: #690


crates/curp/src/client/stream.rs
```diff
@@ -86,9 +86,12 @@ impl Streaming {
                             );
                             self.state.leader_notifier().listen().await;
                         }
-                        CurpError::ShuttingDown(()) => {
-                            debug!("shutting down stream client background task");
-                            break Err(err);
+                        CurpError::RpcTransport(()) => {
```
lxl66566, line 89: Another `CurpError::RpcTransport` below. maybe here should be `CurpError::ShuttingDown`?

crates/curp/src/server/metrics.rs
```diff
@@ -120,7 +120,7 @@ impl Metrics {
                 let sp_size = curp.spec_pool().lock().len();
                 observer.observe_u64(&sp_cnt, sp_size.numeric_cast(), &[]);
 
-                let client_ids = curp.lease_manager().read().expiry_queue.len();
+                let client_ids = curp.lease_manager().read().online_clients();
```
lxl66566, line 123: `client_ids` should be `clients_count`?

crates/curp/src/client/mod.rs
```diff
@@ -355,6 +389,14 @@ impl ClientBuilder {
         })
     }
 
+    /// Wait for client id
+    async fn wait_for_client_id(&self, state: Arc<state::State>) {
+        while state.client_id() == 0 {
+            tokio::time::sleep(*self.config.propose_timeout()).await;
+            debug!("waiting for client_id");
```
lxl66566, line 396: this log should be put before `sleep`

crates/curp/src/client/mod.rs
```diff
@@ -355,6 +389,14 @@ impl ClientBuilder {
         })
     }
 
+    /// Wait for client id
+    async fn wait_for_client_id(&self, state: Arc<state::State>) {
+        while state.client_id() == 0 {
```
lxl66566, line 394: I think it maybe dangerous to loop checking client_id without a break or error. If some reason that the client id not update, the program will hang on.

The only way to modify client_id is `check_gen_local_client_id`, which is a leader-only function. So what if a follower calls `wait_for_client_id`?

crates/curp/src/client/mod.rs
```diff
@@ -355,6 +389,14 @@ impl ClientBuilder {
...
+            tokio::time::sleep(*self.config.propose_timeout()).await;
```
lxl66566, line 395: waiting for a `propose_timeout()` maybe not a good idea? because `propose_timeout` could be set to a long duration, (that's a user-input configuration value).

.config/nextest.toml
```diff
@@ -3,3 +3,4 @@ retries = 0
 slow-timeout = { period = "10s", terminate-after = 3 }
 status-level = "all"
 final-status-level = "slow"
+fail-fast = false
```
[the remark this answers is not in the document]
bsbds: Removed.

crates/curp/src/client/stream.rs
bsbds: fixed

crates/curp/src/server/metrics.rs
bsbds: fixed

crates/curp/src/client/mod.rs
bsbds: fixed

crates/curp/src/client/mod.rs
bsbds: Now it will retry for a limited number of times.

bsbds: The place for other clients to modify client id is in `ClientApi::lease_keep_alive`.

lxl66566 (approved):

PR #949, opened 2024-08-14
refactor: record er and asr for recovered commands

bsbds:
In normal propose path, the leader maintains a gRPC stream to the client. However, if the leader crashes and a new leader is elected, the client won't be able to retrive the previous result.

This PR makes the leader record the er and asr results to the cmd board and they will be returned to the client on next retry.

### DIFF
--- a/crates/curp/src/server/raw_curp/mod.rs
+++ b/crates/curp/src/server/raw_curp/mod.rs
@@ -1861,6 +1861,7 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
     /// Apply new logs
     fn apply(&self, log: &mut Log<C>) {
         let mut entries = Vec::new();
+        let mut resp_txs_l = self.ctx.resp_txs.lock();
         for i in (log.last_as + 1)..=log.commit_index {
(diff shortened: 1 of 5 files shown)


crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -1861,14 +1861,15 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
...
+        let mut resp_txs_l = self.ctx.resp_txs.lock();
         for i in (log.last_as + 1)..=log.commit_index {
             let entry = log.get(i).unwrap_or_else(|| {
                 unreachable!(
                     "system corrupted, apply log[{i}] when we only have {} log entries",
                     log.last_log_index()
                 )
             });
-            let tx = self.ctx.resp_txs.lock().remove(&i);
+            let tx = resp_txs_l.remove(&i);
```
Phoenix500526, line 1872: Why do you enlarge the critical zone?

bsbds: Fixed.

Phoenix500526 (approved):

PR #950, opened 2024-08-14
refactor: cmd board gc

bsbds:
This PR refactored the GC of the cmd board. Commands will only be removed after their client lease has expired.

### DIFF
--- a/crates/curp/src/server/gc.rs
+++ b/crates/curp/src/server/gc.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use indexmap::IndexMap;
 use utils::task_manager::Listener;
 
 use crate::{cmd::Command, rpc::ProposeId, server::cmd_board::CmdBoardRef};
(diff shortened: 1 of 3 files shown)


crates/curp/src/server/gc.rs
```diff
@@ -23,25 +23,23 @@ pub(super) async fn gc_cmd_board<C: Command>(
...
+        let expired_asr_ids: Vec<_> = board
+            .asr_buffer
+            .keys()
+            .copied()
+            .filter(|ProposeId(client_id, _)| !lease_mamanger.read().check_alive(*client_id))
+            .collect();
+        for id in expired_asr_ids {
+            let _ignore = board.asr_buffer.swap_remove(&id);
         }
     }
```
lxl66566, line 44: how about some code like:
```rust
let remove_expired = |buf: &mut IndexMap<
    ProposeId,
    Result<<C as Command>::ER, <C as Command>::Error>,
>| {
    buf.keys()
        .filter(|ProposeId(client_id, _)| !lease_mamanger.read().check_alive(*client_id))
        .for_each(|id| {
            _ = board.asr_buffer.swap_remove(&id);
        });
};
remove_expired(&mut board.er_buffer);
remove_expired(&mut board.asr_buffer);
```
I think we don't need a `copied` here because we only uses `&ProposeId`.

lxl66566 (changes requested):

crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -2062,6 +2062,9 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
     }
 
     /// Acknowledge the propose id and GC it's cmd board result
```
Phoenix500526, line 2064: Be it "GC its cmd board result"?

bsbds: fixed

crates/curp/src/server/gc.rs
bsbds: THe `swap_remove` requires mutable borrow of the buf. We needs to collect the ids before any remove.

lxl66566: ref: https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.retain

PR #951, opened 2024-08-14
fix: after sync result of read only commands

bsbds:
Depends-On: #949 

This PR fixes read only commands execution. 

Currently read only commands won't go through the after sync phase. However, for etcd compatible API, the revision in after sync result must be returned to the client.

### DIFF
--- a/crates/curp-test-utils/src/test_cmd.rs
+++ b/crates/curp-test-utils/src/test_cmd.rs
@@ -287,28 +287,33 @@ impl CommandExecutor<TestCommand> for TestCE {
     fn after_sync(
         &self,
         cmds: Vec<AfterSyncCmd<'_, TestCommand>>,
-        highest_index: LogIndex,
+        highest_index: Option<LogIndex>,
     ) -> Vec<Result<AfterSyncOk<TestCommand>, <TestCommand as Command>::Error>> {
         let as_duration = cmds
             .iter()
             .fold(Duration::default(), |acc, c| acc + c.cmd().as_dur);
         std::thread::sleep(as_duration);
         let total = cmds.len();
-        for (i, cmd) in cmds.iter().enumerate() {
-            let index = highest_index - (total - i - 1) as u64;
-            self.after_sync_sender
-                .send((cmd.cmd().clone(), index))
-                .expect("failed to send after sync msg");
-        }
-        let mut wr_ops = vec![WriteOperation::new_put(
-            META_TABLE,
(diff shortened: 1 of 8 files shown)


crates/curp/src/server/cmd_worker/mod.rs
```diff
@@ -36,29 +36,37 @@ where
...
+    let EntryData::Command(ref cmd) = entry.entry_data else {
+        unreachable!("should not speculative execute {:?}", entry.entry_data);
+    };
+    if cmd.is_read_only() {
+        let result = ce
+            .after_sync(vec![AfterSyncCmd::new(cmd, true)], None)
+            .remove(0)?;
+        let (asr, er_opt) = result.into_parts();
+        let er = er_opt.unwrap_or_else(|| unreachable!("er should exist"));
+        Ok((er, Some(asr)))
```
lxl66566, line 59: little suggestion: `AfterSyncOk::into_parts` returns the tuple in form of `(asr, er)`, but `ErAsr` is a `(er, asr)`.
I think it maybe better to unify them into one form.

crates/curp-test-utils/src/test_cmd.rs
```diff
@@ -287,28 +287,33 @@ impl CommandExecutor<TestCommand> for TestCE {
...
-        for (i, cmd) in cmds.iter().enumerate() {
-            let index = highest_index - (total - i - 1) as u64;
-            self.after_sync_sender
-                .send((cmd.cmd().clone(), index))
-                .expect("failed to send after sync msg");
+        let mut wr_ops = Vec::new();
+
+        if let Some(index) = highest_index {
+            for (i, cmd) in cmds.iter().enumerate() {
+                let index = index - (total - i - 1) as u64;
```
Phoenix500526, line 301: What does this formula mean? Is the `hightest_index` a logical index or something like that?

crates/curp/src/server/cmd_worker/mod.rs
bsbds: They are different types. The first is `(Option<C::ASR>, C::ER)` and the second is `(C::ER, Option<C::ASR>)`

crates/curp-test-utils/src/test_cmd.rs
bsbds: The highest_index means the last index of all cmds, this formula is used to calculate the index of the current cmd.

Phoenix500526 (approved):
LGTM

PR #953, opened 2024-08-15
feat: implement read index

bsbds:
This PR implements the read index mechanism described in the Raft paper.

The implementation is differs slightly from the original algorithm. In this implementation, the client is responsible for broadcasting the `ReadIndex` request to the cluster. This modification reduces latency to 1RTT for non-conflicting read-only commands.

### DIFF
--- a/crates/curp/src/client/tests.rs
+++ b/crates/curp/src/client/tests.rs
@@ -4,7 +4,7 @@ use std::{
     time::Duration,
 };
 
-use curp_test_utils::test_cmd::TestCommand;
+use curp_test_utils::test_cmd::{TestCommand, TestCommandResult};
 use futures::{future::BoxFuture, Stream};
 #[cfg(not(madsim))]
 use tonic::transport::ClientTlsConfig;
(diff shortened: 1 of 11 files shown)


crates/curp/src/server/raw_curp/mod.rs
```diff
@@ -601,6 +601,20 @@ impl<C: Command, RC: RoleChange> RawCurp<C, RC> {
         self.ctx.id_barrier.wait_all(conflict_cmds)
     }
 
+    /// Wait all logs in previous term have been applied to state machine
+    pub(super) fn wait_no_op_applied(&self) -> Box<dyn Future<Output = ()> + Send + Unpin> {
+        // if the leader is at term 1, it won't commit a no-op log
+        if self.term() == 1 {
```
lxl66566, line 607: how to understand that *if the leader is at term 1, it won't commit a no-op log* ?

crates/curp/src/client/tests.rs
```diff
@@ -541,6 +542,68 @@ async fn test_retry_propose_return_retry_error() {
...
+#[traced_test]
+#[tokio::test]
+async fn test_read_index_success() {
+    let connects = init_mocked_connects(5, |id, conn| {
+        conn.expect_propose_stream()
+            .return_once(move |_req, _token, _timeout| {
+                assert_eq!(id, 0, "followers should not receive propose");
+                let resp = async_stream::stream! {
+                    yield Ok(build_propose_response(false));
+                    yield Ok(build_synced_response());
```
lxl66566, line 554: definition of this two `build_*_response` are not found?

crates/curp/src/server/raw_curp/mod.rs
bsbds: In Xline we specify the initial leader at startup, the no-op log is only append on vote success. On the other hand, if the leader is at term 1, there's no other log in previous term, there's no need to wait for a no-op log to be committed.

crates/curp/src/client/tests.rs
bsbds: Fixed.

lxl66566 (approved):

PR #954, opened 2024-08-15
fix: tests in v0.7 branch

bsbds:
Depends-On: #951 

This PR fixes various bugs in the new refactor, please refer to commit messages in each commit for more details.


liangyuanpeng:
Hi @bsbds  , Just wondering if the E2E kube tests are causing any trouble?  Would it help to upload the current xline data to github artifacts?  (I will do it in PR https://github.com/xline-kv/Xline/pull/745)

bsbds:
> Hi @bsbds , Just wondering if the E2E kube tests are causing any trouble? Would it help to upload the current xline data to github artifacts? (I will do it in PR #745) (just want to pushing #745 forward.)

The E2E failure is caused by a Xline bug that could potentially override the current revision. It has been fixed in #955.

crates/xline/src/storage/lease_store/mod.rs
```diff
@@ -52,6 +54,7 @@ pub(crate) struct LeaseStore {
     lease_collection: Arc<LeaseCollection>,
     /// Db to store lease
     db: Arc<DB>,
+    #[allow(unused)] // used in tests
```
Phoenix500526, line 57: Does this index only used in tests? Can we use `cfg` to instead?

crates/xline/src/storage/lease_store/mod.rs
```diff
@@ -98,18 +101,26 @@ impl LeaseStore {
...
+        let next_revision = revision_gen.get().overflow_add(1);
+        let updated = self.sync_request(request, next_revision, txn_db, index)?;
+        let rev = if updated {
             revision_gen.next()
+        } else {
+            revision_gen.get()
         };
-        self.sync_request(request, revision)
-            .map(|(rev, ops)| (SyncResponse::new(rev), ops))
+        // TODO: return only a `SyncResponse`
```
Phoenix500526, line 122: Do we still need this TODO?

crates/xline/src/storage/lease_store/mod.rs
bsbds: Fixed. I completely removed the index from the struct and construct it in tests.

crates/xline/src/storage/lease_store/mod.rs
bsbds: Yes, because the lease store operation does not return any `WriteOp`. This requires a refine to make the code looks more unified, this could be done in the future.

Phoenix500526 (approved):
LGTM

PR #960, opened 2024-08-22
chore: transfer utils's interval map impl into dependency

GG2002:

  Transfer `utils`'s interval map implementation into dependency on `rb-interval-map` which is based on `utils`'s implementation.
  No.


crates/utils/src/lib.rs
```diff
@@ -211,6 +209,8 @@ pub mod tokio_lock;
 pub mod tracing;
 
 use ::tracing::debug;
+/// Interval tree implementation
+pub use interval_map;
```
bsbds, line 213: We could directly use interval_map in xline crate.

GG2002: That would require changing the dependencies of two crates, `xline` and `xlineapi`. Is that okay? Now all we need to do is modify the `utils`.

bsbds (approved):

crates/utils/Cargo.toml
```diff
@@ -25,12 +25,11 @@ derive_builder = "0.20.0"
 event-listener = "5.3.1"
 futures = "0.3.30"
 getset = "0.1"
+interval_map = { package = "rb-interval-map" }
```
lxl66566, line 28: how about adding `version = "0.1.2"`? This line caused an error on rust-analyzer:
```
2024-08-26T00:21:28.930391Z ERROR FetchWorkspaceError
...
Caused by:
  dependency (interval_map) specified without providing a local path, Git repository, version, or workspace dependency to use
```

GG2002: adding `version = "0.1"` is ok, as this crate might require further modifications

PR #974, opened 2024-08-31
feat: replace hardcoded temp directory paths

### LINKED ISSUE #933: [feature]: replace hardcoding "/tmp/xxx" path
currently we use many `/tmp/xxx` temp directory on db testing, and needs manually delete the dir in the end of test.
It's better to use some crate like [tempfile](https://crates.io/crates/tempfile) to help us.

pros: 
- no need to delete folders manually. programmers may forget to add remove code after testing.
- no need to worry about folder name collision. each test will always use different folder.
- although xline only supports test on linux, it's not bad to add some cross-platform ability.

things to do:
- add crate to `[dev-dependencies]` and replace `/tmp/xxx` path
- for those using `/tmp/xxx` path not in test mod, ex. [crates/xline/src/server/command.rs](https://github.com/xline-kv/Xline/blob/master/crates/xline/src/server/command.rs#L383), use [std::env::temp_dir](https://doc.rust-lang.org/std/env/fn.temp_dir.html) instead. 
    - for this ex. case, we need to define a function to generate snapshot path from uuid. It's not a good idea to use hardcoded string in coding.

cswpy:

Fixes #933 
This PR replaces the hardcoded tmp path with calls to [tempfile](https://docs.rs/tempfile/latest/tempfile/index.html) crate

### DIFF
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -35,3 +35,4 @@ workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 
 [dev-dependencies]
 test-macros = { path = "../test-macros" }
+tempfile = "3"
--- a/crates/engine/src/rocksdb_engine/mod.rs
+++ b/crates/engine/src/rocksdb_engine/mod.rs
@@ -726,6 +726,7 @@ impl SnapshotApi for RocksSnapshot {
 #[cfg(test)]
 mod test {
     use std::env::temp_dir;
+    use tempfile::TempDir;
 
     use test_macros::abort_on_panic;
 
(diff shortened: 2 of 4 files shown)


cswpy:
I replaced all the hardcoded paths under `/engine`, just wanna make sure this looks right before proceeding to replace the other references.

bsbds:
Hi, please update the PR to resolve CI issues.

crates/engine/Cargo.toml
```diff
@@ -35,3 +35,4 @@ workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 
 [dev-dependencies]
 test-macros = { path = "../test-macros" }
+tempfile = "3"
```
bsbds, line 38: Please add a new line at the end.

crates/engine/src/rocksdb_engine/mod.rs
```diff
@@ -756,7 +757,7 @@ mod test {
         };
         let res = engine.apply_snapshot(fake_snapshot, &["not_exist"]).await;
         assert!(res.is_err());
-        fs::remove_dir_all(dir).unwrap();
+        dir.close().unwrap();
     }
 
     #[test]
```
bsbds, line 763: The `test_engine_size` test also needs to be updated for consistency.

PR #1005, opened 2024-10-09
feat: reimplement curp client fetch cluster

bsbds:
Depends-On: #981 


crates/curp/src/client/unary/mod.rs
```diff
@@ -64,6 +64,13 @@ pub(super) struct Unary<C: Command> {
     last_sent_seq: AtomicU64,
     /// marker
     phantom: PhantomData<C>,
+
+    #[allow(dead_code)]
```
rogercloud, line 68: It's not deadcode

PR #1006, opened 2024-10-09
feat: reimplement curp client keep alive

bsbds:
Depends-On: #1005


crates/curp/src/client/cluster_state.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Take an async function and map to the dedicated server, return None
+    /// if the server can not found in local state
+    pub(crate) fn map_server<R, F: Future<Output = Result<R, CurpError>>>(
+        &self,
+        id: ServerId,
+        f: impl FnOnce(Arc<dyn ConnectApi>) -> F,
+    ) -> Option<F> {
+        // If the leader id cannot be found in connects, it indicates that there is
+        // an inconsistency between the client's local leader state and the cluster
+        // state, then mock a `WrongClusterVersion` return to the outside.
```
rogercloud, line 61: The comment is not correct

crates/curp/src/client/cluster_state.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self {
+            leader,
+            term,
+            cluster_version,
+            connects,
+        }
+    }
+
+    /// Take an async function and map to the dedicated server, return None
+    /// if the server can not found in local state
```
rogercloud, line 53: The function doesn't need to be async, I think.

crates/curp/src/client/fetch.rs
```diff
@@ -0,0 +1,122 @@
...
+            .map(Response::into_inner)
+            .filter(move |resp| future::ready(resp.term == term))
+            .take(quorum)
+            .count()
+            .map(move |t| t >= quorum)
+            .await
+    }
+
+    /// Prefetch, send fetch cluster request to the cluster and get the
+    /// config with the greatest quorum.
```
rogercloud, line 81: Do you mean term number?

PR #1025, opened 2024-11-18
refactor: split config to multiple modules

### LINKED ISSUE #604: [Refactor]: config
Now all of our configurations are in a single file, `utils/src/config.rs`，It also contains a lot of `parse_xxx`, `xxx_format`, `default_xxx` methods, It's messy and difficult to manage. We need to organize these configurations into modules, each containing the configuration itself, its parse method, and default values.

In addition, the configuration can only be created through the new method, which is very inflexible, we need to refactor them use builder pattern,  and it should be easy to modify existing configurations

Standing-Man:
fixed #604

>  the configurations are only in single module that doesn't handy to manage.

>  i organized these configurations into multiple modules.


JianMinTang:
Hi @bsbds, Why do all self-tests pass locally, but errors occur when running on GitHub Actions? How can I perform self-checks locally to reduce the number of commits?

JianMinTang:
Hi @themanforfree and @bsbds, When I use `cargo audit` to check for dependency vulnerabilities, it suggests that the `rustls` dependency needs to be updated to a newer version. Would you recommend updating it to the latest version to address potential security issues?

standing-man:
Hi @bsbds, when I use `cargo audit` to check for dependency vulnerabilities, it suggests that the `rustls` dependency needs to be updated to a newer version. Would you recommend updating it to the latest version to address potential security issues?

Standing-Man:
WIP: refactor new methods to builder pattern! -> Done!

Standing-Man:
Hi @bsbds, PTAL.

